PR #3, opened 2015-03-31
Allow multiple thumbprints for server and possibility of custom certificate validator

fabiencomte:
Hello,

This version add support of multiples thumbprints for server and custom certificates validator. You can see in client sample for use.

### DIFF
--- a/source/Halibut.SampleClient/Program.cs
+++ b/source/Halibut.SampleClient/Program.cs
@@ -3,11 +3,18 @@ using System.Linq;
 using System.Security.Cryptography.X509Certificates;
 using System.Threading;
 using Halibut.SampleContracts;
+using System.Collections.Generic;
+using System.Net.Security;
 
 namespace Halibut.SampleClient
 {
     class Program
     {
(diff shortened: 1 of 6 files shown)


PaulStovell:
Instead of `ServiceEndPoint` taking multiple thumbprints, could we make it take an `IRemoteCertificateValidator` (an interface we create)? There could then be multiple implementations of that:

 - `SingleThumbprintValidator`
 - `MultipleThumbprintValidator`
 - `DelegateCertificateValidator`

So when people create a `ServiceEndpoint` then can specify their own implementation or use one of those.

fabiencomte:
Hello. I understand but i m not very used with interfaces so i m not really sure about were to put it...



Envoyé depuis Yahoo Mail pour Android

De:"Paul Stovell" <notifications@github.com>
Date:jeu. j avr. AM à 3:17
Objet:Re: [Halibut] Allow multiple thumbprints for server and possibility of custom certificate validator (#3)

Instead of ServiceEndPoint taking multiple thumbprints, could we make it take an IRemoteCertificateValidator (an interface we create)? There could then be multiple implementations of that:

SingleThumbprintValidatorMultipleThumbprintValidatorDelegateCertificateValidator 

So when people create a ServiceEndpoint then can specify their own implementation or use one of those.

PR #13, opened 2015-10-23
Allow both client & server to operate in dual mode to allow ipv4 or i…

zentron:
…pv6 connections

### DIFF
--- a/source/Halibut.SampleServer/Program.cs
+++ b/source/Halibut.SampleServer/Program.cs
@@ -13,7 +13,7 @@ namespace Halibut.SampleServer
             Console.Title = "Halibut Server";
             var certificate = new X509Certificate2("HalibutServer.pfx");
 
-            var endPoint = new IPEndPoint(IPAddress.Any, 8433);
+            var endPoint = new IPEndPoint(IPAddress.IPv6Any, 8433);
 
             var services = new DelegateServiceFactory();
             services.Register<ICalculatorService>(() => new CalculatorService());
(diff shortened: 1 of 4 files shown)


zentron:
What are thoughts on this approach? Should we enable dual mode like this, or provide a config setting to allow ipv6 when its needed?

zentron:
Relates to https://github.com/OctopusDeploy/Issues/issues/1890

source/Halibut.SampleServer/Program.cs
```diff
@@ -13,7 +13,7 @@ public static void Main(string[] args)
             Console.Title = "Halibut Server";
             var certificate = new X509Certificate2("HalibutServer.pfx");
 
-            var endPoint = new IPEndPoint(IPAddress.Any, 8433);
+            var endPoint = new IPEndPoint(IPAddress.IPv6Any, 8433);
```
PaulStovell: It's so weird that IPv6Any also listens on v4

zentron: It in-and-of itself doesnt.. It also requires
`listener.Server.DualMode = (endPoint.Address.AddressFamily == AddressFamily.InterNetworkV6);`
below

zentron:
I might add that this works for test clients, im currently in the process of testing against Octopus Deploy itself

source/Halibut.SampleServer/Program.cs
zentron: And apparently even accessing `DualMode` if the listener is only ipv4 throws an exception.

PR #14, opened 2015-11-08
Support multiple thumbprints for remote

fabiencomte:
It's compatible with single thumbprints but it also support a list of thumbprints for remote. It's useful for connecting different devices with Halibut.

### DIFF
--- a/source/Halibut/HalibutRuntime.cs
+++ b/source/Halibut/HalibutRuntime.cs
@@ -86,6 +86,11 @@ namespace Halibut
             return CreateClient<TService>(new ServiceEndPoint(endpointBaseUri, publicThumbprint));
         }
 
+        public TService CreateClient<TService>(string endpointBaseUri, List<string> publicThumbprint)
+        {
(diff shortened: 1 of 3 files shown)


mjrichardson:
Fabien, thanks for the PR. I apologize it's taken us a while to look at this.

A question: are you using something other than Halibut to implement the server?  I ask because although your changes allow a client to accept responses from multiple thumbprints, a Halibut server (either listening or polling) will still use a single certificate for responses.

fabiencomte:
Hello,
I use an application with Halibut server on multiple electronic boards (each have it s own certificate)
I run a desktop interface with Halibut client that can connect any board.
So it s useful in this case.


Envoyé depuis Yahoo Mail pour Android 
 
  Le lun. j déc. AM à 7:46, Michael Richardson<notifications@github.com> a écrit :   
Fabien, thanks for the PR. I apologize it's taken us a while to look at this.

A question: are you using something other than Halibut to implement the server? I ask because although your changes allow a client to accept responses from multiple thumbprints, a Halibut server (either listening or polling) will still use a single certificate for responses.

PR #15, opened 2015-11-17
Throw HalibutClientException with helpful message instead of an NRE

alfhenrik:
Throw HalibutClientException with helpful message instead of an unhandled exception if we fail to get a response from the endpoint.

This was causing the upgrade Tentacle task for write out a warning about the unhandled NRE being rethrown when upgrading polling Tentacles

image


zentron:
Looks good

PR #16, opened 2015-12-17
Increase timeouts

MJRichardson:
Increasing some default timeouts in an attempt to make connections more robust.

Connects to https://github.com/OctopusDeploy/Issues/issues/1977


tothegills:
lgtm

PR #20, opened 2016-03-22
Adding ability for DataStream to handle packages over 2GB

thebigmoosey:
Adding ability for DataStream to handle packages over 2GB (as MemoryStream maxes out at 2GB).

Fixes OctopusDeploy/Issues#2434

### DIFF
--- a/source/Halibut/DataStream.cs
+++ b/source/Halibut/DataStream.cs
@@ -32,7 +32,16 @@ namespace Halibut
 
         public IDataStreamReceiver Receiver()
         {
-            return receiver ?? new InMemoryDataStreamReceiver(writer);
+            if (receiver != null)
+                return receiver;
+
+            // Use a FileStream for packages over 2GB, or you risk running into OutOfMemory
+            // exceptions with MemoryStream.
+            var maxMemoryStreamLength = 2147483648;
+            if (Length >= maxMemoryStreamLength)
+                return new TemporaryFileDataStreamReceiver(writer);
+            else
+                return new InMemoryDataStreamReceiver(writer);
         }
 
         public bool Equals(DataStream other)
--- a/source/Halibut/DataStream.cs
+++ b/source/Halibut/DataStream.cs
@@ -1,6 +1,5 @@
 using System;
 using System.IO;
-using System.Runtime.Remoting.Messaging;
 using System.Text;
 using Halibut.Transport.Protocol;
 using Newtonsoft.Json;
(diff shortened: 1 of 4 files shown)


source/Halibut/DataStream.cs
```diff
@@ -32,7 +32,16 @@ public DataStream(long length, Action<Stream> writer)
 
         public IDataStreamReceiver Receiver()
         {
-            return receiver ?? new InMemoryDataStreamReceiver(writer);
+            if (receiver != null)
+                return receiver;
+
+            // Use a FileStream for packages over 2GB, or you risk running into OutOfMemory
+            // exceptions with MemoryStream.
+            var maxMemoryStreamLength = 2147483648;
```
zentron: can just use `int.MaxValue` here (2147483647)

source/Halibut/Transport/Protocol/TemporaryFileDataStreamReceiver.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            using (var file = new FileStream(filePath, FileMode.Create))
+            {
+                writer(file);
+            }
+        }
+
+        public void Read(Action<Stream> reader)
+        {
+            using (var tempFiles = new TempFileCollection())
```
zentron: I haven't used `TempFileCollection` before so this is an interesting approach. 
Elsewhere we use `Path.GetTempPath()` but if this works then im for it.
I noticed in the docs it says
> This method can be called only once for each file name extension, because it will return the same name if it is called with the same extension.

I assume this is per `TempFileCollection` and we wont get naming clashes if multiple files are transferred?

source/Halibut/DataStream.cs
thebigmoosey: Good call. I've switched this to Int32.MaxValue to avoid any potential issues on 64-bit systems

source/Halibut/Transport/Protocol/TemporaryFileDataStreamReceiver.cs
```diff
@@ -0,0 +1,36 @@
...
+        public void Read(Action<Stream> reader)
+        {
+            var path = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
+            using (var fileStream = new FileStream(path, FileMode.Create, FileAccess.ReadWrite, FileShare.None, 65536))
+            {
+                writer(fileStream);
+                fileStream.Seek(0, SeekOrigin.Begin);
+                reader(fileStream);
+            }
+            if (File.Exists(path))
```
zentron: maybe wrap this in a try statement with the delete in a finally. just to make sure if it fails  to transfer we still try and delete.

zentron:
Looks good :+1: so long as it passes all tests then merge away.

PR #21, opened 2016-06-01
Proxy support

MJRichardson:
To enable https://github.com/OctopusDeploy/Issues/issues/2490

Due to the changes to `ServiceEndPoint`, this will be a Server-Tentacle breaking-change.


source/Halibut/ServiceEndPoint.cs
```diff
@@ -8,20 +8,29 @@ public class ServiceEndPoint : IEquatable<ServiceEndPoint>
...
 
-        [JsonConstructor]
         public ServiceEndPoint(Uri baseUri, string remoteThumbprint)
         {
             baseUriString = baseUri.GetLeftPart(UriPartial.Authority).ToLowerInvariant();
             this.baseUri = new Uri(baseUriString);
             this.remoteThumbprint = remoteThumbprint;
         }
 
+        [JsonConstructor]
```
MJRichardson: I wonder if we can do this in a backwards-compatible way?

PR #23, opened 2016-07-08
Fixes #22, update version dependency for JSON.net

### LINKED ISSUE #22: NuGet Newtonsoft.Json Dependency
The `.nuspec` dependency for Newtonsoft.Json is `(6.0.0,)`.

However, the project itself seems to use v7.0.1 based on the packages.config. When building out a POC with the project from NuGet, I get this error, which would indicate that the nuspec probably needs to get updated to match the actual usage.

image

mattgwagner:


markryd:
Thanks for that @mattgwagner!

PR #27, opened 2017-01-09
Validated that the proxy address does not have a protocol prefix. 

droyad:
Validated that the proxy address does not have a protocol prefix. Also improved error message when a proxy error occurs. Related to https://github.com/OctopusDeploy/Issues/issues/3054


droyad:
[Validate that the proxy address does not have a protocol prefix](https://trello.com/c/vG2yWaK4/114-validate-that-the-proxy-address-does-not-have-a-protocol-prefix)

PR #29, opened 2017-02-20
Changed the wait for messages to be async.

droyad:
Fixes https://github.com/OctopusDeploy/Issues/issues/3190

By making this async, Octopus does not require a new thread for every Tentacle that is connected. This massively reduces the number of threads that are running.

### DIFF
--- a/source/Halibut/ServiceModel/IPendingRequestQueue.cs
+++ b/source/Halibut/ServiceModel/IPendingRequestQueue.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Threading.Tasks;
 using Halibut.Transport.Protocol;
 
 namespace Halibut.ServiceModel
@@ -7,6 +8,6 @@ namespace Halibut.ServiceModel
     {
         bool IsEmpty { get; }
         void ApplyResponse(ResponseMessage response);
-        RequestMessage Dequeue();
+        Task<RequestMessage> Dequeue();
     }
 }
--- a/source/Halibut/ServiceModel/IPendingRequestQueue.cs
+++ b/source/Halibut/ServiceModel/IPendingRequestQueue.cs
@@ -8,6 +8,7 @@ namespace Halibut.ServiceModel
     {
         bool IsEmpty { get; }
         void ApplyResponse(ResponseMessage response);
-        Task<RequestMessage> Dequeue();
+        RequestMessage Dequeue();
+        Task<RequestMessage> DequeueAsync();
     }
 }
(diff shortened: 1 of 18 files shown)


source/Halibut/ServiceModel/IPendingRequestQueue.cs
```diff
@@ -7,6 +8,6 @@ public interface IPendingRequestQueue
     {
         bool IsEmpty { get; }
         void ApplyResponse(ResponseMessage response);
-        RequestMessage Dequeue();
+        Task<RequestMessage> Dequeue();
```
michaelnoonan: Wonder if we should do this as a non-breaking additive change by adding `Task<RequestMessage> DequeueAsync();`?

droyad:
@michaelnoonan We actually ship .NET 4.5 version of tentacle. It is Calamari that is stuck on 4.0. That addition of `NET45` is a compiler symbol. Probably should have just used `NET40` like the rest of the code.

This changes does not change the public API. But I thought about making it additive, so we could opt in to this behaviour at runtime.

source/Halibut/ServiceModel/IPendingRequestQueue.cs
droyad: Ah, I see that is a public API, but OD doesn't use it. Sure can make it additive.

michaelnoonan:
You're absolutely right! 🤦‍♂️ I keep forgetting it's Calamari which is locked right now. And it's not Tentacle having the scaling problem - we're really just concerned with the Octopus side of things.

Ah, I jumped to conclusions about the public API, I thought because it was a `public interface`. :)

Otherwise I think it looks great.

droyad:
@michaelnoonan I've updated the branch so that non of the `public` classes/interfaces have breaking changes. Easiest to review whole change, not each commit.

source/Halibut/ServiceModel/PendingRequestQueue.cs
```diff
@@ -64,9 +67,27 @@ PendingRequest DequeueNext()
             if (first != null)
                 return first;
 
-            hasItems.Wait(HalibutLimits.PollingQueueWaitTimeout);
+            var cts = new CancellationTokenSource(HalibutLimits.PollingQueueWaitTimeout);
```
michaelnoonan: Should dispose this `CancellationTokenSource`?

source/Halibut/Util/AsyncEx/Source.txt
```diff
@@ -0,0 +1 @@
+﻿The code in this folder and child folders is modified from https://github.com/StephenCleary/AsyncEx
```
michaelnoonan: Cool! Nice to see the attribution.

PR #31, opened 2017-03-09
Convert to a VS2017 project

droyad:


global.json
```diff
@@ -0,0 +1,6 @@
+{
+  "projects": [ "source" ],
```
johnsimons: How did u find out about "projects" ? It is not documented https://docs.microsoft.com/en-us/dotnet/articles/core/tools/global-json

johnsimons: Apparently u don't need it, see https://github.com/dotnet/docs/pull/996#issuecomment-285505275

droyad: My bad, copy pasta :)

build.cake
```diff
@@ -2,243 +2,118 @@
 // TOOLS
 //////////////////////////////////////////////////////////////////////
 #tool "nuget:?package=GitVersion.CommandLine&version=4.0.0-beta0007"
-#addin "MagicChunks"
+#tool "nuget:?package=NUnit.ConsoleRunner"&version=3.6.1
```
johnsimons: @droyad given that we have upgraded to xunit, I don't think we need this anymore

PR #35, opened 2017-05-19
Made the wait for the other end to do work async

droyad:
After dispatching a command from the mailbox to a polling endpoint, the client (eg OctopusServer) will wait for the server (eg Tentacle) to ask for the next message, or send an end message. The server won't send that message however until has processed the request.

The problem is that if there are many polling endpoints, all waiting for the next message, they consume a thread, leading to thread starvation as the thread pool only increases slows (see #29). In otherwords the thread is sitting idle despite the data transfer having completed.

This is not a problem for listening endpoints.

### DIFF
--- a/source/Halibut.Tests/ProtocolFixture.cs
+++ b/source/Halibut.Tests/ProtocolFixture.cs
@@ -361,6 +361,8 @@ namespace Halibut.Tests
                 output.AppendLine("--> PROCEED");
             }
 
+            public Task SendProceedAsync() => Task.Run(() => SendProceed());
+
             public bool ExpectNextOrEnd()
             {
                 if (--numberOfReads == 0)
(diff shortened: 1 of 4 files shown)


droyad:
Yeh, it's purely an internal change, it will still send the data the same. I haven't tested it through our E2E suite yet.

distantcam:
Why not update `ExpectProceed` too? Then we can get rid of 'ReadLine' method.

@droyad Mind if I take this over?

droyad:
The reason I didn't change that or the other methods is that you would either have to make it async all the way up (including right through all the Octopus code), which we are not keen to do, or at some stage call `.Wait()` or `.Result` which defeats the whole point of making it async. 

This restriction does not apply to the incoming polling connections as that is fully contained inside Halibut and isolated.

You can take this @distantcam

PR #36, opened 2017-05-23
Spike to convert Halibut to async all the way

johnsimons:


source/Halibut/IDataStreamReceiver.cs
```diff
@@ -1,11 +1,12 @@
...
+using System.Threading.Tasks;
 
 namespace Halibut
 {
     public interface IDataStreamReceiver
     {
-        void SaveTo(string filePath);
-        void Read(Action<Stream> reader);
+        Task SaveTo(string filePath);
+        Task Read(Func<Stream, Task> reader);
```
danielmarbach: is the assumption here that nothing else needs to be accessed in the lambda but the stream? If not I'd say this probably requires an overload which accepts a state object if you want to avoid delegate allocations due to closure capturing.

source/Halibut/ProxyGen/DispatchProxyAsync.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace System.Reflection
+{
+    public abstract class DispatchProxyAsync
+    {
+        public static T Create<T, TProxy>() where TProxy : DispatchProxyAsync
+        {
+            return (T)CreateProxyInstance(typeof(TProxy), typeof(T));
+        }
+
+        public abstract object Invoke(MethodInfo method, object[] args);
```
danielmarbach: Why the sync invoke here?

source/Halibut/ServiceModel/HalibutProxy.cs
```diff
@@ -1,130 +1,65 @@
...
-                return new ReturnMessage(result, null, 0, null, methodCall);
-            }
-            catch (Exception ex)
-            {
-                return new ReturnMessage(ex, methodCall);
-            }
-        }
-
-        RequestMessage CreateRequest(IMethodMessage methodCall)
+        public override object Invoke(MethodInfo targetMethod, object[] args)
```
danielmarbach: Can be removed

source/Halibut/ServiceModel/HalibutProxy.cs
```diff
@@ -1,130 +1,65 @@
...
+using System.Threading.Tasks;
 using Halibut.Transport.Protocol;
 
 namespace Halibut.ServiceModel
 {
-#if HAS_REAL_PROXY
-    using System.Runtime.Remoting.Messaging;
-    using System.Runtime.Remoting.Proxies;
-    class HalibutProxy : RealProxy
+    public class HalibutProxy : DispatchProxyAsync
```
danielmarbach: Would it be possible to support cancellation?

source/Halibut/ServiceModel/IServiceInvoker.cs
```diff
@@ -1,10 +1,10 @@
-using System;
+using System.Threading.Tasks;
 using Halibut.Transport.Protocol;
 
 namespace Halibut.ServiceModel
 {
     public interface IServiceInvoker
     {
-        ResponseMessage Invoke(RequestMessage requestMessage);
+        Task<ResponseMessage> Invoke(RequestMessage requestMessage);
```
danielmarbach: Cancellation support?

source/Halibut/ServiceModel/PendingRequestQueue.cs
```diff
@@ -10,164 +9,96 @@ namespace Halibut.ServiceModel
...
-            {
-                inProgress.Remove(request.Id);
-            }
+            await pending.WaitUntilComplete().ConfigureAwait(false);
 
             return pending.Response;
         }
 
-        public bool IsEmpty
+        public Task<RequestMessage> DequeueAsync()
```
danielmarbach: CancellationSupport?

source/Halibut/ServiceModel/PendingRequestQueue.cs
```diff
@@ -10,164 +9,96 @@ namespace Halibut.ServiceModel
...
 
-            public RequestMessage Request
+            void ResponseArrived(ResponseMessage responseMessage)
             {
-                get { return request; }
+                SetResponse(responseMessage);
             }
 
-            public void WaitUntilComplete()
+            public RequestMessage Request => request;
```
danielmarbach: Doesn't have to be computed property

source/Halibut/ServiceModel/PendingRequestQueue.cs
```diff
@@ -10,164 +9,96 @@ namespace Halibut.ServiceModel
...
+
+            public async Task WaitUntilComplete()
             {
                 log.Write(EventType.MessageExchange, "Request {0} was queued", request);
 
-                var success = waiter.Wait(HalibutLimits.PollingRequestQueueTimeout);
+                var timeoutTask = Task.Delay(HalibutLimits.PollingRequestQueueTimeout);
+                var success = await Task.WhenAny(waiter.WaitAsync(), timeoutTask)
```
danielmarbach: Not ideal approach. This can leak. For example if the waiter return early the timeout task is still pending. So for each operation you always have two tasks scheduled and one completes only when the timeout is due. I think it would be simpler to ditch the async manual reset event and use TaskCompletionSource<T> with cancellation token source

source/Halibut/ServiceModel/PollingClientCollection.cs
```diff
@@ -19,15 +20,22 @@ public void Add(PollingClient pollingClient)
...
-                foreach (var worker in pollingClients)
-                {
-                    worker.Dispose();
-                }
+                pollingClientsCopy = pollingClients.ToArray();
+            }
 
+            foreach (var worker in pollingClientsCopy)
+            {
+                await worker.Stop().ConfigureAwait(false);
```
danielmarbach: Stop concurrent?

source/Halibut/ServiceModel/ServiceInvoker.cs
```diff
@@ -17,26 +18,34 @@ public ServiceInvoker(IServiceFactory factory)
...
+                var result = await InvokeAsync(method, lease.Service, args).ConfigureAwait(false);
                 return ResponseMessage.FromResult(requestMessage, result);
             }
         }
 
+        static async Task<object> InvokeAsync(MethodBase methodBase, object obj, params object[] parameters)
+        {
+            var task = (Task)methodBase.Invoke(obj, parameters);
+            await task.ConfigureAwait(false);
+            var resultProperty = task.GetType().GetProperty("Result");
```
danielmarbach: Is there a more efficient way to do this? Also this assumes it is always a Task<TResult>, is that safe?

source/Halibut/Transport/SecureListener.cs
```diff
@@ -108,34 +108,37 @@ public int Start()
...
+                {
+                }
+                catch (ObjectDisposedException)
+                {
+                }
+                catch (Exception ex)
+                {
+                    log.WriteException(EventType.Error, "Error accepting TCP client", ex);
+                    // Slow down the logs in case an exception is immediately encountered each AcceptTcpClientAsync
+                    await Task.Delay(1000).ConfigureAwait(false);
```
danielmarbach: would you want to cancel this when shutting down as well?

source/Halibut/Transport/SecureListener.cs
```diff
@@ -167,61 +171,73 @@ public int Start()
             var clientName = client.Client.RemoteEndPoint;
             var stream = client.GetStream();
 
-            using (var ssl = new SslStream(stream, true, AcceptAnySslCertificate))
+            using (cts.Token.Register(() =>
```
danielmarbach: This will allocate closure. Check how bad it is and eventually use the object state overload.

source/Halibut/Transport/SecureWebSocketClient.cs
```diff
@@ -37,7 +37,7 @@ public SecureWebSocketClient(ServiceEndPoint serviceEndpoint, X509Certificate2 c
 
         public ServiceEndPoint ServiceEndpoint => serviceEndpoint;
 
-        public async Task ExecuteTransaction(Func<MessageExchangeProtocol, Task> protocolHandler)
+        public async Task ExecuteTransaction(Func<MessageExchangeProtocol, Task> protocolHandler, CancellationToken token)
```
danielmarbach: you could use `CancellationToken token = default(CancellationToken)`

janpieterz:
Any idea on the future of this pull request / async Halibut concept? 
I'd be able to invest some time into it if needed.

johnsimons:
> Any idea on the future of this pull request / async Halibut concept?

Hi @janpieterz unfortunately I haven't really worked on this since this PR, as far as I know this may still be something we want to achieve, @droyad do you know what the future plans are for Halibut ?

droyad:
@janpieterz @johnsimons The team has no plans to move to async for Halibut. We have not observed any significant performance problem with 1,000's of connections so the benefit of going to async is unclear.

The downsides of a move are that we would need to modify pretty much all of the Octopus deployment pipeline, and some other APIs to make them async, which is a large task. This would inevitably introduce bugs. Another downside is that async is harder to debug, particularly from a process dump.

In short, there are lots of other things we want to work on that would deliver more benefits to the end user.

janpieterz:
@johnsimons  Thanks for jumping in!

@droyad So even if we'd invest the time to make it async it'd not be merged back is my guess? Our scenario is slightly different than Octopus itself, where we have high throughput that stalls on the sync Halibut call, blocking the calling threads. 

We haven't dug too deep into this, but executing an async mock that roughly took as long significantly increased throughput.

Would creating an async flow _and_ keeping the existing sync interface be an option? Not too familiar with the underlying mechanisms as contexts etc, but there should be ways to unwrap exceptions and it would allow a slow migration, wherever possible a switch to async could be made. The underlying (new) bugs within Halibut would still be an issue of course.

droyad:
I don't see a problem with having side by side methods one async and the other not (eg #35). I'm pretty sure we already have that in some places. What I don't want to do though is have sync code calling async methods as that introduces problems with ThreadPool availability AFAIK.

As for your performance problems, are you starting the Halibut requests on the `ThreadPool`? If so, try using `new Thread()` instead. We found that after a certain point the `ThreadPool` would only spin up new threads at a rate of 2 per second, limiting the speed at which we could open connections. There is also an option on `ThreadPool` which allows you to increase this threshold.

janpieterz:
> 
> As for your performance problems, are you starting the Halibut requests on the `ThreadPool`? If so, try using `new Thread()` instead. We found that after a certain point the `ThreadPool` would only spin up new threads at a rate of 2 per second, limiting the speed at which we could open connections. There is also an option on `ThreadPool` which allows you to increase this threshold.

This could hint at something, we're handling this in a thread from NServiceBus message processing. I'll investigate some more over the next weeks and get back.

danielmarbach:
Jan you can also send me details about how you trigger this to daniel dot marbach at particular dot netI might be able to help you track down issues if it is related to NSB

PR #44, opened 2017-11-09
Sleep between retries

frese:
In cases where the tentacle isn't quite ready to receive health-check, we need longer time between retries.


alfhenrik:
@frese I've merged your PR and I'm updating the necessary projects with this change and we will ship it as soon as possible.

Thank you very much for your contribution!

PR #45, opened 2017-11-18
Implemented easy way to override LogFactory

florisrobbemont:
This makes it easier to provide own logging

### DIFF
--- a/source/Halibut/Diagnostics/LogFactoryContext.cs
+++ b/source/Halibut/Diagnostics/LogFactoryContext.cs
@@ -1,0 +1,19 @@
+using System;
+
+namespace Halibut.Diagnostics
+{
+    public static class LogFactoryContext
+    {
+        static ILogFactory logFactory = new LogFactory();
+
+        public static ILogFactory CurrentLogFactory
+        {
+            get { return logFactory; }
+        }
+
+        public static void OverrideLogFactory(ILogFactory factory)
+        {
+            logFactory = factory;
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


MichaelJCompton:
Hi,  thanks for submitting the PR!

Please help me a little in understanding the use case.  Because it uses [LibLog](https://github.com/damianh/LibLog/wiki), the logging in Halibut should just pick up whatever logging is used in the calling application and use that.  E.G. If you remove all trace of Serilog from the [simple client example](https://github.com/OctopusDeploy/Halibut/blob/master/source/Halibut.SampleClient/Program.cs#L8-L18) including the project dependency and replace with a dependency on NLog, it should just pick up the logger and work.

With this PR are you trying to control the logging in a different way?

florisrobbemont:
Hi Michael,

The reason we want this is because we have our own abstraction of a Logging interface. This abstraction is used throughout multiple components.
By overriding the logging factory, I can easily swap in my own logging framework, which in turn uses NLog. However, some of our project use a custom version for logging, that’s why we have our own abstraction.

Does this answer your question?

Met vriendelijke groet,

Floris Robbemont

Op 20 nov. 2017 om 20:56 heeft Michael Compton <notifications@github.com<mailto:notifications@github.com>> het volgende geschreven:


Hi, thanks for submitting the PR!

Please help me a little in understanding the use case. Because it uses LibLog<https://github.com/damianh/LibLog/wiki>, the logging in Halibut should just pick up whatever logging is used in the calling application and use that. E.G. If you remove all trace of Serilog from the simple client example<https://github.com/OctopusDeploy/Halibut/blob/master/source/Halibut.SampleClient/Program.cs#L8-L18> including the project dependency and replace with a dependency on NLog, it should just pick up the logger and work.

With this PR are you trying to control the logging in a different way?

MichaelJCompton:
Thanks, that helps understanding why you have requested this change.

I'm still wondering though if it can be done with the facilities provided already, and I'd like to exhaust that option before making changes.  For example, is there a reason why you can't follow this model?

https://github.com/damianh/LibLog/wiki/Extending

Michael

MichaelJCompton:
Closing due to inactivity.  Please reopen if required.

PR #46, opened 2017-11-20
Use GitLink to build references to GitHub source into .pdb file

MichaelJCompton:
This adds the .pdb file to the nuget package and uses GitLink to embed refs to the repo in the .pdb.  That means, during debugging, VS auto loads the right source from GitHub and allows stepping into and inspection of halibut code when it's used as a NuGet library in another program.

Might also be nice if we served the .pdb on a symbol server.  The above works nicely for development debugging, but looking at crash dumps in VS would be nicer with Hailbut.pdb served off a symbol server.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,9 @@ ipch/
 *.vsp
 *.vspx
 
+# VS Code
+.vscode
+
 # Guidance Automation Toolkit
 *.gpState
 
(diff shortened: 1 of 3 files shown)


matt-richardson:
Awesome stuff, @MichaelJCompton. Been wanting something like this for ages 😄

PR #47, opened 2017-12-13
Added support for netstandard2.0

droyad:
**Breaking Changes**
- Support for netstandard1.5 has been dropped
- Minimum `Newtonsoft.Json` version is `10.0.3`
- For the netstandard build of the library, configuration is no longer read from `appsettings.json`, but via `ConfigurationManager` just like the netfx build

**Changes**
- Now supports `netstandard2.0`
- All builds support WebSockets
- `netstandard` build is not almost identical to the `netfx` build

**Notes**
WebSocket Polling connections cannot be initiated on .NET Core. However it can accept incoming WebSocket requests.


source/Halibut/Transport/SecureListener.cs
```diff
@@ -25,10 +25,9 @@ enum HANDLE_FLAGS : uint
 
     public class SecureListener : IDisposable
     {
-#if CAN_GET_SOCKET_HANDLE
         [DllImport("kernel32.dll", SetLastError = true)]
         static extern bool SetHandleInformation(IntPtr hObject, HANDLE_FLAGS dwMask, HANDLE_FLAGS dwFlags);
```
markryd: Is this going to explode on non-windows?

PR #48, opened 2017-12-13
Removed Async waits where possible

droyad:


source/Halibut/DataStream.cs
```diff
@@ -131,9 +131,9 @@ public void CopyAndReportProgress(Stream destination)
                 while (count > 0)
                 {
                     Swap(ref readBuffer, ref writeBuffer);
-                    var asyncResult = destination.WriteAsync(writeBuffer, 0, count);
+                    var asyncResult = destination.BeginWrite(writeBuffer, 0, count, null, null);
                     count = source.Read(readBuffer, 0, BufferSize);
-                    asyncResult.GetAwaiter().GetResult();
+                    destination.EndWrite(asyncResult);
```
droyad: This was how it was before the .NET Core conversion.

markryd: BeginWrite/EndWrite (and similar on other APIs) weren't in .net core 1. Good to have them back.

PR #49, opened 2017-12-20
Update file permissions after moving from users temp folder

tw17:
This fixes an issue where a script being run as another user is hit with an access denied exception.
When TemporaryFileStream saves a file it copies it from the Users local temp folder which only has read permissions for that User (in this case the user running the Tentacle service). when the file is copied to the Tentacle work directory the permissions from the work folder are not automatically inherited, so only the original user will have read access to the file, where as other files that were created directly in the Tentacle work folder (not copied) have read/execute permission for all users. This problem is only noticeable when the user running a script/accessing a file in the work directory is different to the user running the tentacle service.

Fixes OctopusDeploy/Issues#3881


source/Halibut/Halibut.csproj
```diff
@@ -33,6 +33,12 @@
     <Reference Include="Microsoft.CSharp" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="System.IO.FileSystem.AccessControl">
```
droyad: This should be on on line and up on line 28

source/Halibut/Transport/Protocol/TemporaryFileStream.cs
```diff
@@ -20,6 +21,13 @@ public void SaveTo(string filePath)
 
             AttemptToDelete(filePath);
             File.Move(path, filePath);
+
+            //Update the permission on the file to match inherited permission of the new location.
```
droyad: Instead of a comment, consider moving this code into a method named `SetFilePermissionsToInheritFromParent`.

source/Halibut/Transport/Protocol/TemporaryFileStream.cs
```diff
@@ -20,6 +21,13 @@ public void SaveTo(string filePath)
...
+            var fileInfo = new FileInfo(filePath);
+            var fileSecurity = fileInfo.GetAccessControl();
+            fileSecurity.SetAccessRuleProtection(false, false);
```
droyad: Isn't the 2nd `false` preventing access rules from being inherited, which does not match what the comment says.

source/Halibut/Transport/Protocol/TemporaryFileStream.cs
```diff
@@ -20,10 +20,22 @@ public void SaveTo(string filePath)
...
             GC.SuppressFinalize(this);
         }
 
+        void SetFilePermissionsToInheritFromParent(string filePath)
+        {
+            var fileInfo = new FileInfo(filePath);
+            var fileSecurity = fileInfo.GetAccessControl();
+
+            //When isProtected (first param) is false, SetAccessRuleProtection changes the permissions of the file to allow inherited permissions. 
+            //preserveInheritance (second param) is ignored when isProtected is false.
```
droyad: Riiiiiight. Got it

PR #50, opened 2018-01-16
Added LIBLOG_PUBLIC compilation constant

florisrobbemont:
In order to extend LibLog from external libraries, the LibLog class needs to be compiled with the LIBLOG_PUBLIC constant.

This pull requests overwrites https://github.com/OctopusDeploy/Halibut/pull/45

### DIFF
--- a/source/Halibut/Halibut.csproj
+++ b/source/Halibut/Halibut.csproj
@@ -35,7 +35,7 @@
   </ItemGroup>
 
   <PropertyGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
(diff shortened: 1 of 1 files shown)


droyad:
@florisrobbemont Thanks. I just want to understand the need for this instead of using `Halibut.Logging.LogProvider.SetCurrentLogProvider`.

For example [this LinqPad](http://share.linqpad.net/f43wvf.linq) script logs to the console ([Gist of the same code](https://gist.github.com/droyad/6d0cf8b88d3d55df6ae83a325290aa9d))

PR #53, opened 2018-08-21
Fixing client not sending end

distantcam:
The client code was never sending the `END` message, while the server was in a state where it was reading from the stream. When the client disconnects the stream would throw a disconnect exception.

The tentacle acts as the "server" so this error would appear on the tentacle, which is then reported as an error in the running task.

Fixes OctopusDeploy/Issues#4266

### DIFF
--- a/source/Halibut/Transport/SecureListener.cs
+++ b/source/Halibut/Transport/SecureListener.cs
@@ -113,7 +113,7 @@ namespace Halibut.Transport
                     {
                         log.WriteException(EventType.Error, "Error accepting TCP client", ex);
                         // Slow down the logs in case an exception is immediately encountered each AcceptTcpClientAsync
-                        await Task.Delay(1000);
+                        Thread.Sleep(1000);
                     }
                 }
             }


source/Halibut/Transport/Protocol/MessageExchangeStream.cs
```diff
@@ -321,8 +330,8 @@ void SetNormalTimeouts()
             if (!stream.CanTimeout)
                 return;
 
-            stream.WriteTimeout = (int) HalibutLimits.TcpClientSendTimeout.TotalMilliseconds;
-            stream.ReadTimeout = (int) HalibutLimits.TcpClientReceiveTimeout.TotalMilliseconds;
+            stream.WriteTimeout = (int)HalibutLimits.TcpClientSendTimeout.TotalMilliseconds;
+            stream.ReadTimeout = (int)HalibutLimits.TcpClientReceiveTimeout.TotalMilliseconds;
```
droyad: Spaces again

source/Halibut/Transport/SecureClient.cs
```diff
@@ -215,9 +211,9 @@ static TcpClient CreateTcpClient()
         {
             var client = new TcpClient(AddressFamily.InterNetworkV6)
             {
-                SendTimeout = (int) HalibutLimits.TcpClientSendTimeout.TotalMilliseconds,
-                ReceiveTimeout = (int) HalibutLimits.TcpClientReceiveTimeout.TotalMilliseconds,
-                Client = {DualMode = true}
+                SendTimeout = (int)HalibutLimits.TcpClientSendTimeout.TotalMilliseconds,
```
droyad: whitespace

source/Halibut/Transport/SecureListener.cs
```diff
@@ -86,7 +86,7 @@ public int Start()
 
             log = logFactory.ForEndpoint(new Uri("listen://" + listener.LocalEndpoint));
             log.Write(EventType.ListenerStarted, "Listener started");
-            Task.Run(async () => await Accept());
+            Task.Run(Accept);
```
droyad: Why was this changed? This will not return until the first await is hit from my understanding

source/Halibut/Transport/SecureListener.cs
```diff
@@ -113,7 +113,7 @@ async Task Accept()
                     {
                         log.WriteException(EventType.Error, "Error accepting TCP client", ex);
                         // Slow down the logs in case an exception is immediately encountered each AcceptTcpClientAsync
-                        Thread.Sleep(1000);
+                        await Task.Delay(1000);
```
droyad: Changing this would make it hard to see this loop in a debug dump

source/Halibut/Transport/ConnectionPool.cs
```diff
@@ -0,0 +1,90 @@
...
+using System.Runtime.Serialization;
+using Halibut.Diagnostics;
+using Halibut.Logging;
+
+namespace Halibut.Transport
+{
+    public class ConnectionPool<TKey, TPooledResource>
+        where TPooledResource : class, IPooledResource
+    {
+        readonly ConcurrentDictionary<TKey, ConcurrentDictionary<TPooledResource, byte>> pool = new ConcurrentDictionary<TKey, ConcurrentDictionary<TPooledResource, byte>>();
```
droyad: What's the purpose of this change? It seems unrelated.

source/Halibut.Tests/ProtocolFixture.cs
```diff
@@ -407,8 +412,8 @@ public void Send<T>(T message)
 
             public T Receive<T>()
             {
-                output.AppendLine("<-- " + typeof(T).Name);     
-                return (T) (nextReadQueue.Count > 0 ? nextReadQueue.Dequeue() : default(T));
```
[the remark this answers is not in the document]
distantcam: I suspect this is a resharper thing, as I don't have it so my formatter is removing the space.

source/Halibut/Transport/SecureListener.cs
distantcam: This is creating an unnecessary async state machine and object. Task run starts a new thread to handle the call anyway, and then returns immediately.

distantcam: I think you're thinking of the yield state machine.

source/Halibut/Transport/ConnectionPool.cs
distantcam: It's to fix this malarkey. https://github.com/OctopusDeploy/Halibut/blob/master/source/Halibut/Transport/SecureClientConnectionPool.cs#L51-L81

The initial problem is a concurrency issue where the same object could be returned twice, leading to it being added to the pool twice. This fixes the concurrency issue and ensures the item is only added once. That's why there's a new `ShouldNotAllowMultipleReturnsOfSameConnection` test.

This fix was needed because the pool was not cleaned up properly as part of the dispose, which meant one connection was not disconnecting when it should have.

distantcam:
I'm not sure why this only occurs sometimes. I could only get this error to appear with a high concurrency of clients to the server, so I guess when there's a lot of tasks being sent to the tentacle in a concurrent mode?

source/Halibut/Transport/SecureListener.cs
distantcam: I'm not super tied to this change. Potentially a thread sleep could cause a deadlock because it hogs a thread that could be returned to the async pool, but in practice I doubt that would happen anyway.

source/Halibut.Tests/ProtocolFixture.cs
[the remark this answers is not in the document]
distantcam: We should discuss using a standard formatting spec for our projects (editorconfig) but for now I don't want to hold this PR up. Should I change the spaces back or do you not care enough.

source/Halibut/Transport/ConnectionPool.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            while (true)
+            {
+                var connections = pool.GetOrAdd(endPoint, i => new ConcurrentDictionary<TPooledResource, byte>());
+                AtomicTake(connections, out var connection);
+
+                if (connection == null || !connection.HasExpired())
+                    return connection;
+
+                connection.Dispose();
```
droyad: This should probably have a try/catch around it so that a faulted connection doesn't fail this new connection attempt

source/Halibut/Transport/ConnectionPool.cs
```diff
@@ -0,0 +1,90 @@
...
+                if (AtomicTake(resources, out var dispose))
+                {
+                    dispose.Dispose();
+                }
+            }
+        }
+
+        public void Clear(TKey key, Diagnostics.ILog log = null)
+        {
+            if (!pool.TryRemove(key, out var connections))
```
droyad: What if another thread is on line 25 at this point. We would dispose all the connections and then the other thread would take an already disposed connection out of the pool. We could be halfway though the foreach on ln 56 when this occurs.

source/Halibut/Transport/SecureListener.cs
droyad: Yeh I think we should keep it the way it is, it's super useful when debugging

source/Halibut/Transport/SecureListener.cs
droyad: Ah yes, you are correct

droyad: I think we should change this back to `new Thread(Accept).Run()` now that .NET Core supports the Sync method.

source/Halibut/Transport/ConnectionPool.cs
distantcam: `ConnectionPool` is generic and only works with `IPooledResource`. I'm not sure what you mean by faulted connection or how to detect that from the interface.

droyad: `IPooledResource` is really a `SecureConnection`. The Dispose method of that could throw an exception (although SocketException is already covered). All I meant was that a failure in this dispose call shouldn't cause the call to this method to fail.

PR #54, opened 2018-08-21
Store log messages in a single place to limit total number in memory

distantcam:
Log messages are stored in the `InMemoryConnectionLog` and are limited to 100, but there's a `InMemoryConnectionLog` for each endpoint, which can lead to a memory leak when there's a lot of endpoints.

Fixes OctopusDeploy/Issues#4824

### DIFF
--- a/source/Halibut.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress.NETCore.approved.cs
+++ b/source/Halibut.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress.NETCore.approved.cs
@@ -138,6 +138,7 @@ namespace Halibut.Diagnostics
     public class HalibutLimits
     {
         public static TimeSpan ConnectionErrorRetryTimeout;
+        public static int LogStorageLimit;
         public static TimeSpan PollingQueueWaitTimeout;
         public static TimeSpan PollingRequestMaximumMessageProcessingTimeout;
         public static TimeSpan PollingRequestQueueTimeout;
@@ -167,7 +168,7 @@ namespace Halibut.Diagnostics
     }
     public class InMemoryConnectionLog : Halibut.Diagnostics.ILog
     {
-        public InMemoryConnectionLog(string endpoint) { }
+        public InMemoryConnectionLog(string endpoint, Halibut.Diagnostics.LogEventStorage logEventStorage) { }
         public IList<Halibut.Diagnostics.LogEvent> GetLogs() { }
         public void Write(Halibut.Diagnostics.EventType type, string message, Object[] args) { }
         public void WriteException(Halibut.Diagnostics.EventType type, string message, Exception ex, Object[] args) { }
@@ -182,6 +183,12 @@ namespace Halibut.Diagnostics
         public DateTimeOffset Time { get; }
         public Halibut.Diagnostics.EventType Type { get; }
         public string ToString() { }
+    }
+    public class LogEventStorage
+    {
+        public LogEventStorage() { }
+        public void AddLog(string endpoint, Halibut.Diagnostics.LogEvent logEvent) { }
+        public IList<Halibut.Diagnostics.LogEvent> GetLogs(string endpoint) { }
     }
     public class LogFactory : Halibut.Diagnostics.ILogFactory
     {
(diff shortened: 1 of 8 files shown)


droyad:
How much memory are we talking about here? This information is shown in the UI and it would be confusing if some endpoints had no logs even though they had been just deployed to.

source/Halibut/Diagnostics/LogEventStorage.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            if (events.TryGetValue(endpoint, out var logs))
+                return logs.ToArray();
+
+            return EmptyLogs;
+        }
+
+        public void AddLog(string endpoint, LogEvent logEvent)
+        {
+            lock (events)
```
droyad: Logs would be inserted often, but retrieved rarely. I would optimise this for the write scenario, which would also make it a lot simpler. Just keep a single queue of all messages `ConcurrentQueue` and filter it when logs are requests. No locks are needed then.

source/Halibut.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress.NETCore.approved.cs
```diff
@@ -167,7 +168,7 @@ public interface ILogFactory
     }
     public class InMemoryConnectionLog : Halibut.Diagnostics.ILog
     {
-        public InMemoryConnectionLog(string endpoint) { }
+        public InMemoryConnectionLog(string endpoint, Halibut.Diagnostics.LogEventStorage logEventStorage) { }
```
droyad: Breaking API change, we should bump the version number via commit message

source/Halibut/Diagnostics/HalibutLimits.cs
```diff
@@ -31,6 +34,7 @@ static HalibutLimits()
         public static TimeSpan TcpClientHeartbeatReceiveTimeout = TimeSpan.FromSeconds(60);
         public static TimeSpan TcpClientConnectTimeout = TimeSpan.FromSeconds(60);
         public static TimeSpan PollingQueueWaitTimeout = TimeSpan.FromSeconds(30);
+        public static int LogStorageLimit = 5000;
```
droyad: This number seems much too low. If I have 10,000 tentacles, every second one would have 1 message on average. Perhaps 100,000 is better.

source/Halibut/Diagnostics/LogEventStorage.cs
droyad: On second thoughts, that does allow a single endpoint to monopolise the messages (as it also does with this new code). Previously a single endpoint could only have 100 messages. perhaps also add that logic...

source/Halibut.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress.NETCore.approved.cs
distantcam: Ahhh, I didn't think this would be a breaking change, but you're right. TBH if we're gonna break the API I'd change the logging api so we don't have to hold onto the format objects of the logs, which is what's causing this problem in the first place.

source/Halibut/Diagnostics/HalibutLimits.cs
distantcam: 100,000 is close to where the original user had the issue (300,000 in their case). If customers have 10,000 tentacles (which is a rare case) then they can bump the `Halibut.LogStorageLimit` config value.

In any case the `GetLogs` call that means Halibut has to hold onto logs isn't even used in OD or Tentacle. If we could obsolete that call the better solution is just not store the logs.

distantcam:
> How much memory are we talking about here? This information is shown in the UI and it would be confusing if some endpoints had no logs even though they had been just deployed to.

The specific API call `GetLogs` isn't even used by OD or Tentacle. They get logs through the `SendToTrace` call which is a fire-and-forget style, so nothing is retained. If we're willing to break the public API then we should simplify `GetLogs` to return a formatted string instead of the full log object.

source/Halibut/Diagnostics/HalibutLimits.cs
droyad: Unfortunately any Halibut settings need to be re-applied after a Server Upgrade :(

droyad:
Let's consider whether we can do without this API

distantcam:
Closing this in favour of #55

PR #64, opened 2019-02-22
Added Microsoft.DotNet.Analyzers.Compatibility to get better compile …

johnsimons:
…time multi platform errors


source/Halibut/Transport/Protocol/TemporaryFileStream.cs
```diff
@@ -33,12 +42,14 @@ void SetFilePermissionsToInheritFromParent(string filePath)
...
+#pragma warning disable PC001 // API not supported on all platforms
                 var fileSecurity = fileInfo.GetAccessControl();
 
                 //When isProtected (first param) is false, SetAccessRuleProtection changes the permissions of the file to allow inherited permissions. 
                 //preserveInheritance (second param) is ignored when isProtected is false.
                 fileSecurity.SetAccessRuleProtection(false, false);
                 fileInfo.SetAccessControl(fileSecurity);
+#pragma warning restore PC001 // API not supported on all platforms
             }
             catch (PlatformNotSupportedException)
```
zentron: You can probably take this out now. This was a work around for the same problem

johnsimons: fixed

PR #66, opened 2019-03-19
Under .net core linux + mac dispose of HalibutRuntime hung

johnsimons:
Making dispose a little bit better by allowing cancellation of timeouts

Because I Threads with Tasks, we lose naming of threads, not sure this is a big deal or not?


johnsimons:
Replaced with #67

zentron:
Looks good.

PR #68, opened 2019-04-01
Ignore access exceptions when attempting to correct ACL

jburger:
These exceptions can cause deployments to fail when Server uses this code to write out artifacts to disk.

Fixes: https://github.com/OctopusDeploy/Issues/issues/5433

### DIFF
--- a/source/Halibut/Transport/Protocol/TemporaryFileStream.cs
+++ b/source/Halibut/Transport/Protocol/TemporaryFileStream.cs
@@ -53,7 +53,7 @@ namespace Halibut.Transport.Protocol
             }
             catch (UnauthorizedAccessException ex)
             {
-                log.Write(EventType.Security, $"Ignoring an unauthorized access issue: {ex.Message}. {nameof(TemporaryFileStream)} assumes that filesystem permissions allow full control to the executing user for {filePath}. Update those permissions to remove this log.");
+                log?.Write(EventType.Security, $"Ignoring an unauthorized access issue: {ex.Message}. {nameof(TemporaryFileStream)} assumes that filesystem permissions allow full control to the executing user for {filePath}. Update those permissions to remove this log.");
             }
         }
 
--- a/source/Halibut/Transport/Protocol/TemporaryFileStream.cs
+++ b/source/Halibut/Transport/Protocol/TemporaryFileStream.cs
@@ -51,9 +51,9 @@ namespace Halibut.Transport.Protocol
                 fileInfo.SetAccessControl(fileSecurity);
 #pragma warning restore PC001 // API not supported on all platforms
             }
-            catch (UnauthorizedAccessException)
+            catch (UnauthorizedAccessException ex)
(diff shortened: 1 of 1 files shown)


source/Halibut/Transport/Protocol/TemporaryFileStream.cs
```diff
@@ -39,15 +39,22 @@ public void SaveTo(string filePath)
...
-            fileInfo.SetAccessControl(fileSecurity);
+                //When isProtected (first param) is false, SetAccessRuleProtection changes the permissions of the file to allow inherited permissions. 
+                //preserveInheritance (second param) is ignored when isProtected is false.
+                fileSecurity.SetAccessRuleProtection(false, false);
+                fileInfo.SetAccessControl(fileSecurity);
 #pragma warning restore PC001 // API not supported on all platforms
+            }
+            catch (UnauthorizedAccessException)
+            {
+                // ignored, user has likely disallowed full control over their artifact directory
```
thebigmoosey: Feels like we should be logging this comment here, even if it's just a verbose log, so customers have some hint that it's falling into this code path. I guess the flip side of this, is there a good reason not to log it?

jburger: :thinking: Good point, as a diagnostic patch we originally did a bunch of logging (to `OctopusServer.txt`) to ensure we had the right fix. I ripped it out after a chat with some other team members as we felt that that 'cut' was probably too noisy. 

That said, I think you've identified a good middle ground, in the event something unexpected occurs, it would be nice to be able to capture this intent in the logs somewhere. 

Thanks! :bowing_man:

source/Halibut/Transport/Protocol/TemporaryFileStream.cs
```diff
@@ -39,15 +39,22 @@ public void SaveTo(string filePath)
...
-            fileInfo.SetAccessControl(fileSecurity);
+                //When isProtected (first param) is false, SetAccessRuleProtection changes the permissions of the file to allow inherited permissions. 
+                //preserveInheritance (second param) is ignored when isProtected is false.
+                fileSecurity.SetAccessRuleProtection(false, false);
+                fileInfo.SetAccessControl(fileSecurity);
 #pragma warning restore PC001 // API not supported on all platforms
+            }
+            catch (UnauthorizedAccessException ex)
+            {
+                log.Write(EventType.Security, $"Ignoring an unauthorized access issue: {ex.Message}. {nameof(TemporaryFileStream)} assumes that filesystem permissions allow full control to the executing user for {filePath}. Update those permissions to remove this log.");
```
thebigmoosey: Ooo, one thing .. can `log` be null? Should it need a null check `log?.Write...` ? Worth checking just in case

jburger: Good pick @thebigmoosey hadn't considered that.

Issue #70, opened 2019-04-11
External management of Trust


vincentparrett:
I'm investigating adding support for the MS Generic hosting environment, and have gotten quite far. One issue I have found though, is that the abstraction hides access to the runtime, which makes adding and removing trusted thumbprints difficult. 

What are your thoughts on adding a ITrustProvider (and a default implementation to use when not provided), and a constructor on HalibutRuntime to take one in.

Something like this :

````c#
public interface ITrustProvider 
{
  void Add(string clientThumbprint);
  void Remove(string clientThumbprint);
  bool IsTrusted(string clientThumbprint);
}

public class HalibutRuntime : IHalibutRuntime
{
        public HalibutRuntime(IServiceFactory serviceFactory, X509Certificate2 serverCertificate, ITrustProvider trustProvider)

}

````
This would allow us to inject the ITrustProvider elsewhere in our application to manage the trust without needing access to the Halibut runtime (which in my case would be contained in an IHostedService instance (which would then take the ITrustProvider in it's constructor).

vincentparrett:
PR submitted, was a really simple change.

vincentparrett:
Closing, implemented in PR #81

PR #71, opened 2019-04-11
Added ITrustProvider and default implementation

vincentparrett:
The PR makes it possible to externalise the management of client thumprint trust, which makes it easier to manage trust without exposing the runtime instance in places where it's not really needed.  

BTW, I updated the public surface tests with the changes, not sure if was something I should do or should leave for you, but seemed like bad form to push failing tests so I went ahead and did it anyway. The interface changes should still be backwards compatible.


vincentparrett:
This implements the suggestion in #70

vincentparrett:
PR #80  messes up this PR - I moved the locking around the trustedThumbprints inside the trustProvider - however you just added more code inside the lock. I need some guidance here as it's no longer a simple change.

vincentparrett:
One option is to change the ITrustProvider interface so the methods take an action to run in the lock

````c#
public interface ITrustProvider
{
    void Add(string clientThumbprint, Action onChanged);
...
}
````
Then the implementation needs to call the action
````C#
public void Add(string clientThumbprint, Action onChange = null)
{
    lock (trustedThumbprints)
    {
       trustedThumbprints.Add(clientThumbprint);
       onChange?.Invoke();
    }
}
`````
The trouble there is that is relying on the implementer on ITrustProvider to do the right thing. 
Thoughts?

**Comment by tothegills** [comment]
I wonder if we can do it without a callback?

I don't think `DisconnectFromAllListeners` needs to be inside the lock. `RemoveTrust` is straightforward:
```
        public void RemoveTrust(string clientThumbprint)
        {
            trustProvider.Remove(clientThumbprint);
            DisconnectFromAllListeners(clientThumbprint);
        }
```

`TrustOnly` is a bit more complicated, `DisconnectFromAllListeners` wants to know which thumbprints have had trust revoked which is currently difficult to do via `ITrustProvider`.

vincentparrett:
What about adding a ToArray method to the trustProvider

````c#
       public string[] ToArray()
        {
            lock (trustedThumbprints)
            {
                return trustedThumbprints.ToArray();
            }
        }
````
Then TrustOnly can be something like this 

`````c#
        public void TrustOnly(IReadOnlyList<string> thumbprints)
        {
            var thumbprintsRevoked = trustProvider.ToArray().Except(thumbprints).ToArray();
            trustProvider.TrustOnly(thumbprints);
            DisconnectFromAllListeners(thumbprintsRevoked);
        }
`````

tothegills:
I think that approach is reasonable. Do you want to update the PR and I will merge?

vincentparrett:
Yep will do.

tothegills:
Merged in #81

Issue #73, opened 2019-04-16
Notify when unauthorized client connects.


vincentparrett:
Currently there doesn't seem to be any way of notifying the HalibutRuntime consumer when an unauthorized users attempts to connect (other than a debug log message). 

I think it might be useful for some applications to make the users or admins aware of these attempts. Looking at the code, it doesn't look too difficult to add a callback from the SecureListener and SecureWebSocketListener's Authorize method. 

PR on the way.

vincentparrett:
Implemented.

PR #74, opened 2019-04-16
Added HalibutRuntime.UnAuthorizedClientConnect action

vincentparrett:
Implements #73 - notify runtime user when unauthorized client connects.


vincentparrett:
Any feedback on this PR or my other one? About to make a big commitment to switching from WCF to Halibut... but this particular feature is something I need.. really don't want to maintain a fork.

source/Halibut/Transport/SecureWebSocketListener.cs
```diff
@@ -203,6 +205,7 @@ async Task<bool> Authorize(HttpListenerContext context, EndPoint clientName)
             if (!isAuthorized)
             {
                 log.Write(EventType.ClientDenied, "A client at {0} connected, and attempted a message exchange, but it presented a client certificate with the thumbprint '{1}' which is not in the list of thumbprints that we trust", clientName, thumbprint);
+                unauthorizedClientConnect(clientName.ToString(), thumbprint);
```
droyad: Instead of returning something here, what if you just re-checked `verifyClientThumbprint`?

source/Halibut/Transport/SecureWebSocketListener.cs
```diff
@@ -17,31 +17,32 @@ public class SecureWebSocketListener : IDisposable
...
         readonly ILogFactory logFactory;
         readonly Func<string> getFriendlyHtmlPageContent;
         readonly Func<Dictionary<string, string>> getFriendlyHtmlPageHeaders;
         readonly CancellationTokenSource cts = new CancellationTokenSource();
         ILog log;
         HttpListener listener;
 
-        public SecureWebSocketListener(string endPoint, X509Certificate2 serverCertificate, Action<MessageExchangeProtocol> protocolHandler, Predicate<string> verifyClientThumbprint, ILogFactory logFactory, Func<string> getFriendlyHtmlPageContent)
-            : this(endPoint, serverCertificate, h => Task.Run(() => protocolHandler(h)), verifyClientThumbprint, logFactory, getFriendlyHtmlPageContent, () => new Dictionary<string, string>())
+        public SecureWebSocketListener(string endPoint, X509Certificate2 serverCertificate, Action<MessageExchangeProtocol> protocolHandler, Predicate<string> verifyClientThumbprint, Action<string, string> unauthorizedClientConnect, ILogFactory logFactory, Func<string> getFriendlyHtmlPageContent)
```
droyad: Can we add a new constructor instead and the rest just pass an empty callback? Then we don't need to bump a major and the other consumers of this lib won't have to change their code.

source/Halibut/Transport/SecureWebSocketListener.cs
vincentparrett: What if I changed the callback to a 
````Func<string, string, bool>````
That would allow us to avoid calling ````verifyClientThumbprint```` again to just do what we already probably just did in the callback (trust the thumbprint).

File: source/Halibut/Transport/SecureWebSocketListener.cs (hunk 2 above)
Comment (vincentparrett) [reply source/Halibut/Transport/SecureWebSocketListener.cs]: Done.

File: source/Halibut/Transport/SecureWebSocketListener.cs (hunk 1 above)
Comment (vincentparrett) [reply source/Halibut/Transport/SecureWebSocketListener.cs]: Changed to func.

Comment (matt-richardson) [reply source/Halibut/Transport/SecureWebSocketListener.cs]: This was actually part of an internal conversation - @droyad had thought he was replying to me.

What if we returned an enum instead? That was is super clear what the ramification of the return value is - I find bools in method definitions are unclear, especially when doing code reviews etc. 

Maybe something like `BlockConnection`/`TrustThisThumbprintAndAllowConnection?

**Comment by matt-richardson** [comment]
Sorry for the delay in looking at this. We just started discussing internally - looks like we’ve got a good use case for this ourselves.

File: source/Halibut/Transport/SecureListener.cs [hunk 3]
```diff
@@ -65,12 +66,27 @@ public SecureListener(IPEndPoint endPoint, X509Certificate2 serverCertificate, F
             this.serverCertificate = serverCertificate;
             this.protocolHandler = protocolHandler;
```
matt-richardson: It’d be better for the constructor in line 63 to call your new constructor - will reduce duplication

vincentparrett: Done

source/Halibut/Transport/SecureWebSocketListener.cs
vincentparrett: Seems a bit wordy to me, and there is already a predicate param verifyClientThumbprint as a precedent. 

Returning TrustThisThumbprintAndAllowConnection suggests that the thumbprint would be added to the trusted list as a result, but that's not the case, the listener doesn't have the ability to do that. Perhaps that could be done by passing in the list instead of a predicate, but a better option would be to pass an ITrustProvider (seem my other PR).

vincentparrett: Thought about this a bit more and I agree returning an enum makes sense.. and it was possible to do the trust thing by handling that in the runtime.

source/Halibut/Transport/SecureWebSocketListener.cs
```diff
@@ -37,11 +38,16 @@ public SecureWebSocketListener(string endPoint, X509Certificate2 serverCertifica
         }
 
         public SecureWebSocketListener(string endPoint, X509Certificate2 serverCertificate, Func<MessageExchangeProtocol, Task> protocolHandler, Predicate<string> verifyClientThumbprint, ILogFactory logFactory, Func<string> getFriendlyHtmlPageContent)
-            : this(endPoint, serverCertificate, h => Task.Run(() => protocolHandler(h)), verifyClientThumbprint, logFactory, getFriendlyHtmlPageContent, () => new Dictionary<string, string>())
+            : this(endPoint, serverCertificate, h => Task.Run(() => protocolHandler(h)), verifyClientThumbprint,  logFactory, getFriendlyHtmlPageContent, () => new Dictionary<string, string>())
```
matt-richardson:
```suggestion
            : this(endPoint, serverCertificate, h => Task.Run(() => protocolHandler(h)), verifyClientThumbprint, logFactory, getFriendlyHtmlPageContent, () => new Dictionary<string, string>())
```

source/Halibut/Transport/SecureWebSocketListener.cs
```diff
@@ -37,11 +38,16 @@ public SecureWebSocketListener(string endPoint, X509Certificate2 serverCertifica
...
         {
         }
 
         public SecureWebSocketListener(string endPoint, X509Certificate2 serverCertificate, Func<MessageExchangeProtocol, Task> protocolHandler, Predicate<string> verifyClientThumbprint, ILogFactory logFactory, Func<string> getFriendlyHtmlPageContent, Func<Dictionary<string, string>> getFriendlyHtmlPageHeaders)
+            : this(endPoint, serverCertificate, protocolHandler, verifyClientThumbprint, logFactory, getFriendlyHtmlPageContent, getFriendlyHtmlPageHeaders, (x, y) => false)
```
matt-richardson:
```suggestion
            : this(endPoint, serverCertificate, protocolHandler, verifyClientThumbprint, logFactory, getFriendlyHtmlPageContent, getFriendlyHtmlPageHeaders, (clientName, thumbprint) => false)
```

source/Halibut/HalibutRuntime.cs
```diff
@@ -216,6 +221,20 @@ public void Dispose()
             }
         }
 
+        protected bool NotifyUnAuthorizedClientConnect(string clientName, string thumbPrint)
```
matt-richardson: This returning a `bool` doesn't sit right with me. The name of the method doesn't really help me understand what a return value of `true` or `false` actually means.
Can we use another enum? `BlockConnection`/`AllowConnection`?

source/Halibut/HalibutRuntime.cs
```diff
@@ -206,6 +210,7 @@ public void SetFriendlyHtmlPageHeaders(IEnumerable<KeyValuePair<string, string>>
             friendlyHtmlPageHeaders = headers?.ToDictionary(x => x.Key, x => x.Value) ?? new Dictionary<string, string>();
         }
 
+
```
matt-richardson:
```suggestion
```

source/Halibut/HalibutRuntime.cs
```diff
@@ -13,6 +13,7 @@
 
 namespace Halibut
 {
+
```
matt-richardson:
```suggestion
```

source/Halibut/HalibutRuntime.cs
```diff
@@ -43,6 +44,9 @@ public ILogFactory Logs
             get { return logs; }
         }
 
+        public Func<string, string, HandleUnAuthorizedClientMode> UnAuthorizedClientConnect { get; set; }
+
+
```
matt-richardson:
```suggestion
```

source/Halibut/IHalibutRuntime.cs
```diff
@@ -5,6 +5,12 @@
 
 namespace Halibut
 {
+    public enum HandleUnAuthorizedClientMode
```
matt-richardson: Consider `UnAuthorizedClientConnectAction`?

Also, the capitalization on the `A` is throwing me. I wonder if `Unauthorized` is better? A quick google says its a valid word, so no need to capitalize the `A`.

vincentparrett: Yep, agreed.

source/Halibut/HalibutRuntime.cs
vincentparrett: I changed it to use the HandleUnauthorizedClientMode enum instead, and changed the name to OnUnauthorizedClientConnect as it's really firing an event if assigned.

matt-richardson:
Thanks for the PR, @vincentparrett! Sorry it took us so long to jump on it.

I've merged it in, with a few tweaks to the names, and it's been published as `4.3.0`.

vincentparrett:
Thanks. Any chance you can look my other PR?

PR #75, opened 2019-04-29
Dispose of stream to allow cleanup of open connections

distantcam:
This should help with the open connections left over from a closed server.

### DIFF
--- a/source/Halibut/Transport/SecureListener.cs
+++ b/source/Halibut/Transport/SecureListener.cs
@@ -100,7 +100,7 @@ namespace Halibut.Transport
             };
             backgroundThread.Start();
 
-            return ((IPEndPoint) listener.LocalEndpoint).Port;
+            return ((IPEndPoint)listener.LocalEndpoint).Port;
         }
 
         void Accept()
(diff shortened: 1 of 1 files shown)


source/Halibut/Transport/SecureListener.cs
```diff
@@ -200,7 +200,17 @@ async Task ExecuteRequest(TcpClient client)
...
                     if (Authorize(ssl, clientName))
                     {
+                        // The ExchangeMessage call can hang on reading the stream which keeps a thread alive,
+                        // so we dispose the stream which will cause the thread to abort with an exceptions.
+                        var weakSSL = new WeakReference(ssl);
+                        cts.Token.Register(() =>
+                        {
+                            if (weakSSL.IsAlive)
+                                ((IDisposable)weakSSL.Target).Dispose();
+                        }, true);
```
johnsimons: Does this need to be `true`?

source/Halibut/Transport/SecureListener.cs
```diff
@@ -200,7 +200,17 @@ async Task ExecuteRequest(TcpClient client)
...
+
                         // Delegate the open stream to the protocol handler - we no longer own the stream lifetime
+                        //await Task.WhenAny(cts.Token.WhenCanceled(), ExchangeMessages(ssl));
```
johnsimons: Not needed?

distantcam: Whoops. Leftover from other things I was trying.

source/Halibut/Transport/SecureListener.cs
distantcam: Nope. Will change.

PR #76, opened 2019-04-30
Add method to dispose all connections to an endpoint

tothegills:
This method will be used when a machine is deleted from Octopus to immediately close existing connections to that machine.

Do we need to worry about connections that are not in the pool?

### DIFF
--- a/source/Halibut/Transport/ConnectionManager.cs
+++ b/source/Halibut/Transport/ConnectionManager.cs
@@ -1,4 +1,6 @@
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using Halibut.Diagnostics;
 
 namespace Halibut.Transport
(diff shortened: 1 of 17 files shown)


distantcam:
There's 2 types of connections that won't be in the pool.

- Currently active connections that are being used in `HalibutRuntime.SendOutgoingHttpsRequest`. You could use a [CountdownEvent](https://docs.microsoft.com/en-us/dotnet/api/system.threading.countdownevent?view=netframework-4.5) to wait until all active connections are done.

- Connections used for polling clients. This would need a way to remove polling clients from the collection in `HalibutRuntime` based on the endpoint.

tothegills:
It looks to me like `pollingClients` are the clients that the server (Tentacle) polls. I'm struggling to see how a polling connection can be closed from a client. Tempted to leave it as polling servers only ever open one connection so there isn't as large an impact on open ports.

source/Halibut/Transport/ConnectionManager.cs
```diff
@@ -0,0 +1,125 @@
...
+                if (activeConnections.TryGetValue(serviceEndpoint, out var connections))
+                {
+                    connections.Remove(connection);
+                }
+            }
+        }
+
+        public void ClearPooledConnections(ServiceEndPoint serviceEndPoint, ILog log)
+        {
+            pool.Clear(serviceEndPoint, log);
```
johnsimons: Does this one need to be locked?

source/Halibut/Transport/SecureConnection.cs
```diff
@@ -36,6 +37,7 @@ public bool HasExpired()
 
         public void Dispose()
         {
+            OnDisposed?.Invoke(this, EventArgs.Empty);
```
johnsimons: So should `OnDisposed?.Invoke` be called at the end of this method, given is past tense?

PR #77, opened 2019-05-12
Ensure DeflateStream is wrapped in BufferedStream

johnsimons:
This PR also removed our BufferedStream copy, no longer required.

According to http://faithlife.codes/blog/2012/06/always-wrap-gzipstream-with-bufferedstream/, it seems `DeflateStream` compresses individual bytes as they get written, hence we need to wrap it in a `BufferedStream`.
Our code implementation was doing this wrapping the wrong way.

This also fix an issue in netcore, see https://octopusdeploy.slack.com/archives/CD1VC0KRQ/p1557451375076300


source/Halibut/Transport/Protocol/MessageExchangeStream.cs
```diff
@@ -294,15 +294,69 @@ static DataStream FindStreamById(StreamCapture capture, Guid id)
...
+            using (var zip = new DeflateStream(stream, CompressionMode.Compress, true))
+            using (var buffer = new BufferedStream(new NonDisposableStreamWrapper(zip), 4096))
+            using (var bson = new BsonDataWriter(buffer) { CloseOutput = false })
             {
                 serializer.Serialize(bson, new MessageEnvelope { Message = messages });
                 bson.Flush();
             }
         }
 
+        class NonDisposableStreamWrapper : Stream
```
droyad: Why is this needed?

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
```diff
@@ -235,9 +235,9 @@ void ExpectServerIdentity()
 
         T ReadBsonMessage<T>()
         {
-            using (var buffer = new BufferedStream(stream, 8192, true))
-            using (var zip = new DeflateStream(buffer, CompressionMode.Decompress, true))
-            using (var bson = new BsonDataReader(zip) { CloseInput = false })
+            using (var zip = new DeflateStream(stream, CompressionMode.Decompress, true))
+            using (var buffer = new BufferedStream(new NonDisposableStreamWrapper(zip), 8192))
```
droyad: The blog post only mentions `Compress` wouldn't `Decompress` use the reverse logic?

I would be good to get some metrics around this change to confirm it works as intended.

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
johnsimons: We want to leave the wrapped stream opened.

johnsimons: At least that is what the previous code was doing

johnsimons: But maybe given we explicitly saying to leave the inner stream open, we don't need it!

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
johnsimons: The read part of the code is interesting because it makes me wonder if we even need to buffer at all, to me ` return (T)serializer.Deserialize<MessageEnvelope>(bson).Message;` needs the whole payload anyway.

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
johnsimons: Ok I have removed NonDisposableStreamWrapper

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
johnsimons: @droyad any thoughts here?

distantcam: Yeah this feels like a copy-n-paste error from the write part. We don't need buffering on read.

droyad: After dealing with this on the Server Nancy side, I think it's best to run a test on this. Perhaps use the console apps to test out the transfer speed of a large file from the US to AU with the various combinations. 

This might actually be the cause of the slowness some customers are reporting when transferring packages.

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
droyad: The original code passes `true` to the `BufferedStream` ctor while the new code doesn't.

droyad: It looks like Halibut has it's own `BufferedStream` implementation

droyad: Ah, you removed it, GH hid that change away from me :)

distantcam: The `DeflateStream` can keep the stream open, and that's passed into the `BufferedStream` so we don't need a custom implementation for that anymore.

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
```diff
@@ -294,9 +294,9 @@ static DataStream FindStreamById(StreamCapture capture, Guid id)
 
         void WriteBsonMessage<T>(T messages)
         {
-            using (var buffer = new BufferedStream(stream, 4096, true))
-            using (var zip = new DeflateStream(buffer, CompressionMode.Compress, true))
-            using (var bson = new BsonDataWriter(zip) { CloseOutput = false })
+            using (var zip = new DeflateStream(stream, CompressionMode.Compress, true))
+            using (var buffer = new BufferedStream(zip, 4096))
```
distantcam: Could you add a comment saying why we need a buffer here, and linking to that blog post please.

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
droyad: 👍 Makes sense,only the first stream in the stack needs a keep open

distantcam:
Why was the project file changed?

johnsimons:
@droyad run a test before and after transferring 850MB file from AU to US in Azure.
Before these modifications, it took a minute
With these modifications took 30 secs.

The new code seems to be a  lot faster.
Anything else i should look for?

johnsimons:
> Why was the project file changed?

@distantcam because the wildcard causes a recursive copy, so if you run the tests multiple times on your machine and then have a look at bin\debug, you will see end up with bin\debug\bin\debug\bin\debug....

droyad:
This seems like worth calling out in an Octopus Release note "Improved package transfer speed from the server to Tentacles"

PR #80, opened 2019-05-20
Disconnect polling servers when trust is revoked

tothegills:
Prior to this PR polling servers would maintain their connection to the client after the client had revoked trust. This PR will close any TCP clients associated with a thumbprint when trust is revoked.

As an added bonus, active listening connections are now disconnected when the halibut runtime is disposed.


source/Halibut.Tests/TcpClientManagerFixture.cs
```diff
@@ -0,0 +1,46 @@
...
+        [Test]
+        public void ShouldRemoveDisconnectedClients()
+        {
+            const string thumbprint = "123";
+            var manager = new TcpClientManager();
+
+            manager.AddActiveClient(thumbprint, new TcpClient());
+            manager.AddActiveClient(thumbprint, new TcpClient());
+
+            manager.GetActiveClients(thumbprint).Should().HaveCount(1);
```
matt-richardson: Not clear why this should only have 1?

source/Halibut/HalibutRuntime.cs
```diff
@@ -170,16 +170,42 @@ public void Trust(string clientThumbprint)
...
+
+        void DisconnectUntrustedClients(IReadOnlyCollection<string> thumbprints)
+        {
+            foreach (var thumbprint in thumbprints)
+            {
+                DisconnectUntrustedClient(thumbprint);
+            }
+        }
+
+        void DisconnectUntrustedClient(string thumbprint)
```
matt-richardson: Does this need to have `Untrusted` in the name? Does it matter if they're trusted or not?

source/Halibut/HalibutRuntime.cs
```diff
@@ -170,16 +170,42 @@ public void Trust(string clientThumbprint)
...
+        {
+            foreach (var listener in listeners)
```
matt-richardson: Consider folding the if into the foreach

source/Halibut/Transport/ConnectionManager.cs
```diff
@@ -119,6 +127,18 @@ void OnConnectionDisposed(IConnection connection)
...
 
+        void SafelyDisposeConnection(IConnection connection, ILog log)
+        {
+            try
+            {
+                connection?.Dispose();
+            }
+            catch (Exception ex)
+            {
+                log?.WriteException(EventType.Error, "Exception disposing connection from pool", ex);
```
matt-richardson: Any thoughts what might cause this? I'm thinking along the lines of trying to have actionable errors logged if possible
Or is this a "just in case"?

source/Halibut/Transport/SecureListener.cs
```diff
@@ -277,17 +294,21 @@ void SendFriendlyHtmlPage(Stream stream)
...
-                log.Write(EventType.ClientDenied, "A client at {0} connected, and attempted a message exchange, but did not present a client certificate", clientName);
-                return false;
+                return null;
             }
 
             var thumbprint = new X509Certificate2(stream.RemoteCertificate.Export(X509ContentType.Cert)).Thumbprint;
+            return thumbprint;
+        }
+
+        bool Authorize(string thumbprint, EndPoint clientName)
```
matt-richardson: Consider putting a guard on the thumbprint != null, now that the null check has been moved out of here.

source/Halibut/Transport/ConnectionManager.cs
tothegills: This is just in case, I've never seen one in the wild

distantcam: We don't do this anywhere else, and if `log` is null there will be other problems.

source/Halibut/Transport/TcpClientManager.cs
```diff
@@ -0,0 +1,61 @@
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Sockets;
+
+namespace Halibut.Transport
+{
+    public class TcpClientManager
```
distantcam: This class doesn't need to be public, and therefore doesn't need to be part of the public API.

source/Halibut/Transport/TcpClientManager.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            lock (activeClients)
+            {
+                if (activeClients.TryGetValue(thumbprint, out var value))
+                {
+                    return value.ToArray();
+                }
+            }
+
+            return Enumerable.Empty<TcpClient>().ToArray();
```
distantcam: `return new TcpClient[0];`

Even better, create a static field with the empty array, and return that.

source/Halibut/Transport/TcpClientManager.cs
```diff
@@ -0,0 +1,61 @@
...
+                }
+                else
+                {
+                    tcpClients = new HashSet<TcpClient> {client};
+                    activeClients.Add(thumbprint, tcpClients);
+                }
+            }
+        }
+
+        static void RemoveStaleClients(HashSet<TcpClient> tcpClients)
```
distantcam: This method can be inlined.

source/Halibut/HalibutRuntime.cs
```diff
@@ -170,16 +170,40 @@ public void Trust(string clientThumbprint)
...
+        {
+            foreach (var thumbprint in thumbprints)
+            {
+                DisconnectFromAllListeners(thumbprint);
+            }
+        }
+
+        void DisconnectFromAllListeners(string thumbprint)
+        {
+            var secureListeners = listeners.Where(l => l is SecureListener).Cast<SecureListener>();
```
distantcam: This `Where.Cast` is equivalent to `.OfType<SecureListener>()`.

PR #81, opened 2019-05-21
Clean up of #71

tothegills:
Refer to #71


vincentparrett:
Why does DefaultTrustProvider need to be public? I made it internal as it only intended for use when an ITrustProvider is not supplied.

vincentparrett:
Ah my apologies, I actually made it public last week so I could use it with DI in the hosting extensions project... then promptly forgot that and made it internal today. It can stay internal, the hosting extensions project can provide it's own implementation of the interface.

PR #84, opened 2019-05-27
Properly dispose SecureClient to not throw exception when shutting down

tw17:
When shutting down a polling client, it was rethrowing an IOException exception, we now check to see if we are disposing before rethrowing the exception.


tw17:
taking another approach

PR #85, opened 2019-05-27
Check if we are disposing/shutting down before throwing exception

tw17:
Fixes an issue where an IOException gets thrown when shutting down polling tentacles


johnsimons:
@tw17 @droyad I think this fix can be problematic because when we set the `IsDisposed` flag in connection manager (https://github.com/OctopusDeploy/Halibut/pull/85/files#diff-60f680517315c19d5268ef4c73d60782R99) we are inside the Dispose method, so when we check it in https://github.com/OctopusDeploy/Halibut/pull/85/files#diff-e46247574e5858d794a199a58deff411R63 GC could have already kicked in and that object is gone, thoughts?

droyad:
GC would never (outside of a `WeakReference`) remove an object that is referenced by another object that is still alive. The GC also doesn't call dispose https://stackoverflow.com/a/1691850 apparently.

PR #86, opened 2019-06-05
Added a phased random retry to if the polling target is not available…

droyad:
… or refuses the connection. Also cleared up the log messages.

To Test:
- Uncomment just lines 32 and 33 of SampleClient.Program (Comment out the thumbprint Trust and the websocket listening part)
- Comment in just line 39 (//Begin Polling Setup) in the Sample Server

### DIFF
--- a/source/Halibut/Diagnostics/InMemoryConnectionLog.cs
+++ b/source/Halibut/Diagnostics/InMemoryConnectionLog.cs
@@ -33,17 +33,27 @@ namespace Halibut.Diagnostics
 
         void WriteInternal(LogEvent logEvent)
         {
-            var logLevel = logEvent.Type == EventType.Diagnostic ||
-                           logEvent.Type == EventType.SecurityNegotiation ||
-                           logEvent.Type == EventType.MessageExchange
-                ? LogLevel.Trace 
-                : LogLevel.Info;
+            var logLevel = GetLogLevel(logEvent);
             SendToTrace(logEvent, logLevel);
 
             events.Enqueue(logEvent);
 
-            LogEvent ignore;
-            while (events.Count > 100 && events.TryDequeue(out ignore)) { }
+            while (events.Count > 100 && events.TryDequeue(out _)) { }
+        }
+
+        static LogLevel GetLogLevel(LogEvent logEvent)
+        {
+            switch (logEvent.Type)
+            {
+                case EventType.Error:
+                    return LogLevel.Error;
+                case EventType.Diagnostic:
+                case EventType.SecurityNegotiation:
+                case EventType.MessageExchange:
+                    return LogLevel.Trace;
+                default:
+                    return LogLevel.Info;
+            }
         }
 
         void SendToTrace(LogEvent logEvent, LogLevel level)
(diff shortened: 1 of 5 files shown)


source/Halibut/Diagnostics/InMemoryConnectionLog.cs
```diff
@@ -33,17 +33,27 @@ public IList<LogEvent> GetLogs()
...
-            LogEvent ignore;
-            while (events.Count > 100 && events.TryDequeue(out ignore)) { }
+            while (events.Count > 100 && events.TryDequeue(out _)) { }
+        }
+
+        static LogLevel GetLogLevel(LogEvent logEvent)
+        {
+            switch (logEvent.Type)
+            {
+                case EventType.Error:
```
droyad: Errors were being treated as Trace

source/Halibut/Transport/PollingClient.cs
```diff
@@ -40,14 +41,23 @@ public void Start()
...
                 try
                 {
+                    retry.Try();
                     secureClient.ExecuteTransaction(protocol =>
                     {
                         protocol.ExchangeAsSubscriber(subscription, handleIncomingRequest);
                     });
+                    retry.Success();
+                }
+                catch (HalibutClientException hce)
```
droyad: I was wrong, the code does not fall through to this block regularly. However it does so with a `HalibutClientException`. The next catch is for those really out of ordinary cases.

source/Halibut/Transport/SecureClient.cs
```diff
@@ -79,6 +79,7 @@ public void ExecuteTransaction(Action<MessageExchangeProtocol> protocolHandler)
                 {
                     log.Write(EventType.Error, $"The remote host at {(ServiceEndpoint == null ? "(Null EndPoint)" : ServiceEndpoint.BaseUri.ToString())} refused the connection, this may mean that the expected listening service is not running.");
                     lastError = cex;
+                    retryAllowed = false;
```
droyad: If the connection is refused we don't want to retry immediately

source/Halibut/Transport/SecureClient.cs
```diff
@@ -137,21 +138,20 @@ void HandleError(Exception lastError, bool retryAllowed)
...
             lastError = lastError.UnpackFromContainers();
 
+            var innermost = lastError;
+            while (innermost.InnerException != null)
+                innermost = innermost.InnerException;
+            
+            if (innermost is SocketException se && !retryAllowed)
+                if (se.SocketErrorCode == SocketError.ConnectionAborted || se.SocketErrorCode == SocketError.ConnectionReset)
+                    throw new HalibutClientException($"The server {ServiceEndpoint.BaseUri} aborted the connection before it was fully established. This usually means that the server rejected the certificate that we provided. We provided a certificate with a thumbprint of '{clientCertificate.Thumbprint}'.");
+
```
droyad: This cleans up the error message that gets returned without the extra guff that was previously added

source/Halibut/Transport/SecureClient.cs
```diff
@@ -137,21 +138,20 @@ void HandleError(Exception lastError, bool retryAllowed)
 
             lastError = lastError.UnpackFromContainers();
 
+            var innermost = lastError;
```
johnsimons: Replace with `lastError.GetBaseException()`?

droyad: Oh! nice TIL

PR #87, opened 2019-06-06
Disposed of WebSocket connections when the client's certificate is not recognised.

droyad:
This was keeping connections open by the server and causing the client to run out of memory.

Added a timeout to the receive on the client so that it kills those open connections (for connections by previous version of Halibut).
Correctly closed the connections at both ends. I'm guessing it changed between versions of the lib and now requires CloseAsync and CloseOutputAsync calls.
Cleaned up the message output
Fixes https://github.com/OctopusDeploy/Issues/issues/4227

### DIFF
--- a/source/Halibut/Transport/Protocol/WebSocketStream.cs
+++ b/source/Halibut/Transport/Protocol/WebSocketStream.cs
@@ -4,6 +4,7 @@ using System.Net.WebSockets;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using Halibut.Diagnostics;
 
 namespace Halibut.Transport.Protocol
 {
@@ -49,13 +50,24 @@ namespace Halibut.Transport.Protocol
             var buffer = new ArraySegment<byte>(new byte[10000]);
             while (true)
             {
-                var result = await context.ReceiveAsync(buffer, cancel.Token);
-                if (result.MessageType != WebSocketMessageType.Text)
-                    throw new Exception($"Encountered an unexpected message type {result.MessageType}");
-                sb.Append(Encoding.UTF8.GetString(buffer.Array, 0, result.Count));
+                using(var cts = new CancellationTokenSource(HalibutLimits.TcpClientReceiveTimeout))
+                using(var combined = CancellationTokenSource.CreateLinkedTokenSource(cts.Token, cancel.Token))
+                {
+                    var result = await context.ReceiveAsync(buffer, combined.Token);
+                    if (result.MessageType == WebSocketMessageType.Close)
+                    {
+                        using(var sendCancel = new CancellationTokenSource(TimeSpan.FromSeconds(1)))
+                            await context.CloseOutputAsync(WebSocketCloseStatus.NormalClosure, "Close received", sendCancel.Token);
+                        return null;
+                    }
 
-                if (result.EndOfMessage)
-                    return sb.ToString();
+                    if (result.MessageType != WebSocketMessageType.Text)
+                        throw new Exception($"Encountered an unexpected message type {result.MessageType}");
+                    sb.Append(Encoding.UTF8.GetString(buffer.Array, 0, result.Count));
+
+                    if (result.EndOfMessage)
+                        return sb.ToString();
+                }
             }
         }
 
@@ -102,9 +114,9 @@ namespace Halibut.Transport.Protocol
             if (context.State != WebSocketState.Open)
                 return;
 
-            var sendCancel = new CancellationTokenSource(TimeSpan.FromSeconds(1));
-            context.SendAsync(new ArraySegment<byte>(new byte[0]), WebSocketMessageType.Close, true, sendCancel.Token)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+            using(var sendCancel = new CancellationTokenSource(TimeSpan.FromSeconds(1)))
+                context.CloseOutputAsync(WebSocketCloseStatus.NormalClosure, "Closing", sendCancel.Token)
+                    .ConfigureAwait(false).GetAwaiter().GetResult();
         }
 
         protected override void Dispose(bool disposing)
--- a/source/Halibut/Transport/SecureClient.cs
+++ b/source/Halibut/Transport/SecureClient.cs
@@ -10,6 +10,7 @@ using System.Threading;
 using Halibut.Diagnostics;
 using Halibut.Transport.Protocol;
 using Halibut.Transport.Proxy;
+using Halibut.Util;
 
 namespace Halibut.Transport
 {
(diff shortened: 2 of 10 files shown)


source/Halibut/Transport/Protocol/WebSocketStream.cs
```diff
@@ -49,13 +50,24 @@ public async Task<string> ReadTextMessage()
...
             while (true)
             {
-                var result = await context.ReceiveAsync(buffer, cancel.Token);
-                if (result.MessageType != WebSocketMessageType.Text)
-                    throw new Exception($"Encountered an unexpected message type {result.MessageType}");
-                sb.Append(Encoding.UTF8.GetString(buffer.Array, 0, result.Count));
-
-                if (result.EndOfMessage)
-                    return sb.ToString();
+                using(var cts = new CancellationTokenSource(HalibutLimits.TcpClientReceiveTimeout))
```
droyad: Drops connections that have not sent data after 10 minutes... don't know if that is too long

source/Halibut/Transport/Protocol/WebSocketStream.cs
```diff
@@ -49,13 +50,24 @@ public async Task<string> ReadTextMessage()
...
+                using(var combined = CancellationTokenSource.CreateLinkedTokenSource(cts.Token, cancel.Token))
+                {
+                    var result = await context.ReceiveAsync(buffer, combined.Token);
+                    if (result.MessageType == WebSocketMessageType.Close)
```
droyad: Other end closed

source/Halibut/Transport/SecureClient.cs
```diff
@@ -80,7 +80,6 @@ public void ExecuteTransaction(Action<MessageExchangeProtocol> protocolHandler)
                     log.Write(EventType.Error, $"The remote host at {(ServiceEndpoint == null ? "(Null EndPoint)" : ServiceEndpoint.BaseUri.ToString())} refused the connection, this may mean that the expected listening service is not running.");
                     lastError = cex;
                     retryAllowed = false;
-                    Thread.Sleep(retryInterval);
```
droyad: Should have removed this from the other change. No need to sleep here, the outer loop will do that

source/Halibut/Transport/SecureWebSocketClient.cs
```diff
@@ -77,7 +77,6 @@ public void ExecuteTransaction(Action<MessageExchangeProtocol> protocolHandler)
                 }
                 catch (AuthenticationException aex)
                 {
-                    log.WriteException(EventType.Error, $"Authentication failed while setting up connection to {(serviceEndpoint == null ? "(Null EndPoint)" : serviceEndpoint.BaseUri.ToString())}", aex);
```
droyad: Message will be printed by the parent

source/Halibut/Transport/WebSocketConnectionFactory.cs
```diff
@@ -62,15 +62,25 @@ ClientWebSocket CreateConnectedClient(ServiceEndPoint serviceEndpoint)
                         .ConfigureAwait(false).GetAwaiter().GetResult();
                 ServerCertificateInterceptor.Validate(connectionId, serviceEndpoint);
             }
+            catch
```
droyad: Fix the connection leak

droyad:
I figured out that `SecureClient` is shared between polling and listening. The retries for polling are a bit different between the two (polling retries for ever, listening doesn't). We've been changing this class and breaking it for one or the other. I'll put up another PR to split that into two.

droyad:
I ended up rolling back the recent changes to `SecureClient` as a new class `SecureListeningClient` as `SecureClient` was used by both polling and listening and they were in conflict when it comes to the retry logic. This will give us a break

droyad:
[Halibut memory leak](https://trello.com/c/AXw70vGI/1012-halibut-memory-leak)

source/Halibut/Transport/Protocol/WebSocketStream.cs
```diff
@@ -102,9 +114,9 @@ void SendCloseMessage()
             if (context.State != WebSocketState.Open)
                 return;
 
-            var sendCancel = new CancellationTokenSource(TimeSpan.FromSeconds(1));
-            context.SendAsync(new ArraySegment<byte>(new byte[0]), WebSocketMessageType.Close, true, sendCancel.Token)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+            using(var sendCancel = new CancellationTokenSource(TimeSpan.FromSeconds(1)))
+                context.CloseOutputAsync(WebSocketCloseStatus.NormalClosure, "Closing", sendCancel.Token)
+                    .ConfigureAwait(false).GetAwaiter().GetResult();
```
zentron: Not that it probably matters much in what we are doing, but shouldn't the close initiator wait for the close response from the other side?

source/Halibut/Transport/Protocol/WebSocketStream.cs
zentron: Nice. have been playing around with websockets myself recently and had to double think when it came to the proper close handshake

source/Halibut/Transport/SecureListeningClient.cs
```diff
@@ -0,0 +1,157 @@
...
+                catch (AuthenticationException aex)
+                {
+                    log.WriteException(EventType.Error, $"Authentication failed while setting up connection to {(ServiceEndpoint == null ? "(Null EndPoint)" : ServiceEndpoint.BaseUri.ToString())}", aex);
+                    lastError = aex;
+                    retryAllowed = false;
+                    break;
+                }
+                catch (SocketException cex) when (cex.SocketErrorCode == SocketError.ConnectionRefused)
+                {
+                    log.Write(EventType.Error, $"The remote host at {(ServiceEndpoint == null ? "(Null EndPoint)" : ServiceEndpoint.BaseUri.ToString())} refused the connection, this may mean that the expected listening service is not running.");
```
zentron: nothing major. just a tidy thing, the repeated `(ServiceEndpoint == null ? "(Null EndPoint)" : ServiceEndpoint.BaseUri.ToString()` could be put into a local method.

source/Halibut/Transport/Protocol/WebSocketStream.cs
zentron: On further reading it looks like `CloseOutputAsync` will sort that out for you

source/Halibut/Transport/WebSocketConnectionFactory.cs
```diff
@@ -62,15 +62,25 @@ ClientWebSocket CreateConnectedClient(ServiceEndPoint serviceEndpoint)
...
+            {
+                if (client.State == WebSocketState.Open)
+                    using (var sendCancel = new CancellationTokenSource(TimeSpan.FromSeconds(1)))
+                        client.CloseAsync(WebSocketCloseStatus.ProtocolError, "Certificate thumbprint not recognised", sendCancel.Token)
```
zentron: Should we be using `CloseOutputAsync` here as well?

source/Halibut/Transport/Protocol/WebSocketStream.cs
droyad: It falls into a a heap if I try a `CloseAsync` here, possibly because the connection has already been killed by the other end

source/Halibut/Transport/WebSocketConnectionFactory.cs
droyad: I'm not sure to be honest. It works like this and both sides are happy. This https://stackoverflow.com/a/26744644/10784 suggests CloseAsync is a nicer way of doing things.

> There are two ways to close a WebSocket connection. The graceful way is CloseAsync which when initiated sends a message to the connected party, and waits for acknowledgement. The keyword here is sends. Remember in our previous point we discussed that you can only have a single send or receive at any given point in time? So if you are sending data, and at the same time try to CloseAsync this leads to an exception because CloseAsync will also try to send a message. The other option is to use CloseOutputAsync this is more of a “fire-and-forget” approach.

PR #88, opened 2019-06-21
Avoiding locking around creation of connections

TomPeters:
Fixes: https://github.com/OctopusDeploy/Issues/issues/5641

Previously, [this lock](https://github.com/OctopusDeploy/Halibut/compare/FixLockingInConnectionManager?expand=1#diff-60f680517315c19d5268ef4c73d60782L18):
```
        lock (activeConnections)	            
            {	            
                var connection = pool.Take(serviceEndpoint) ?? CreateConnection(connectionFactory, serviceEndpoint, log);
                AddConnectionToActiveConnections(serviceEndpoint, connection);	

                return connection;	
            }
```

was wrapping the `CreateConnection` call. This `CreateConnection` call actually establishes a connection to the target, so can take some time.

This means that any other code that is trying to acquire a connection must wait for this connection to be established before they can proceed. This turns a process that should happen concurrently (creation of multiple connections) into a serial process.

This is made even worse by the fact that if the connection could not be established, this may need to wait for a timeout period to elapse. If there are many consumers trying to acquire connections at the same time, this can easily add up and cause significant delays.

The fix is the smallest possible change, which involves continuing to lock the surrounding code, but ensuring that `CreateConnection` is not within the lock.

## How to review this PR
Without worrying too much about style, verify that this change looks safe to make and won't introduce any new problems. 

It would be nice to ship this change ASAP to fix some real customer problems. We can always come back and redesign the code later if we think it is worthwhile. So please leave any comments, but also approve this PR if you think it is shippable.

### DIFF
--- a/source/Halibut/Transport/ConnectionManager.cs
+++ b/source/Halibut/Transport/ConnectionManager.cs
@@ -15,15 +15,17 @@ namespace Halibut.Transport
 
         public IConnection AcquireConnection(IConnectionFactory connectionFactory, ServiceEndPoint serviceEndpoint, ILog log)
         {
-            lock (activeConnections)
-            {
-                var connection = pool.Take(serviceEndpoint) ?? CreateConnection(connectionFactory, serviceEndpoint, log);
-                AddConnectionToActiveConnections(serviceEndpoint, connection);
-
-                return connection;
-            }
+            var connection = GetExistingConnectionFromPool(serviceEndpoint) ?? CreateConnection(connectionFactory, serviceEndpoint, log);
+            AddConnectionToActiveConnections(serviceEndpoint, connection);
+            return connection;
         }
 
+        IConnection GetExistingConnectionFromPool(ServiceEndPoint serviceEndpoint)
+        {
+            lock (activeConnections)
+                return pool.Take(serviceEndpoint);
+        }
+        
         IConnection CreateConnection(IConnectionFactory connectionFactory, ServiceEndPoint serviceEndpoint, ILog log)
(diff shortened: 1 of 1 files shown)


source/Halibut/Transport/ConnectionManager.cs
```diff
@@ -14,21 +14,36 @@ public class ConnectionManager : IDisposable
...
+                var openableConnection = existingConnectionFromPool != null 
+                    ? Tuple.Create<IConnection, Action>(existingConnectionFromPool, () => { }) // existing connections from the pool are already open
+                    : CreateNewConnection(connectionFactory, serviceEndpoint, log);
+                AddConnectionToActiveConnections(serviceEndpoint, openableConnection.Item1);
+                return openableConnection;
             }
         }
 
-        IConnection CreateConnection(IConnectionFactory connectionFactory, ServiceEndPoint serviceEndpoint, ILog log)
+        Tuple<IConnection, Action> CreateNewConnection(IConnectionFactory connectionFactory, ServiceEndPoint serviceEndpoint, ILog log)
```
droyad: Could you return `(IConnection connection, Action create)` instead?

TomPeters: I opted not to do that right now because this requires a new dependency on `System.ValueTuple`. I'll ship it as is and come back to this

droyad: Ah, no worries, I thought it already had that, but that must be Nevermore

Issue #92, opened 2019-08-12
Listening Server leaks TCP connections


droyad:
It looks like #80 (`4.3.4`) introduced a problem where TCP connections are not being disposed, or at least are still in memory. Customers report that the TCP connection count on listening tentacles increase.

When I tested it, the issue is not present in `4.3.3`, but is in `4.3.4` and `4.3.15`.

# Steps to Reproduce
1. In the Halibut project set `SampleClient.Program.Main` to:
```
Log.Logger = new LoggerConfiguration()
    .WriteTo.ColoredConsole()
    .CreateLogger();

Console.Title = "Halibut Client";
var certificate = new X509Certificate2("HalibutClient.pfx");

var hostName = args.FirstOrDefault() ?? "localhost";
var port = args.Skip(1).FirstOrDefault() ?? "8433";
while (true)
{
    using (var runtime = new HalibutRuntime(certificate))
    {
        var calculator = runtime.CreateClient<ICalculatorService>("https://" + hostName + ":" + port + "/", "EF3A7A69AFE0D13130370B44A228F5CD15C069BC");
    
        try
        {
            var result = calculator.Add(12, 18);
            Console.WriteLine("12 + 18 = " + result);
            Thread.Sleep(100);
        }
        catch (Exception ex)
        {
            Console.Write(ex.Message);
        }
    }
}
```

2. `SampleServer.Program.Main` to:
```
Log.Logger = new LoggerConfiguration()
    .WriteTo.ColoredConsole()
    .MinimumLevel.Warning()
    .CreateLogger();

Console.Title = "Halibut Server";
var certificate = new X509Certificate2("HalibutServer.pfx");

var endPoint = new IPEndPoint(IPAddress.IPv6Any, 8433);

var services = new DelegateServiceFactory();
services.Register<ICalculatorService>(() => new CalculatorService());

using (var server = new HalibutRuntime(services, certificate))
{
    server.Listen(endPoint);
    server.Trust("2074529C99D93D5955FEECA859AEAC6092741205");

    Console.WriteLine("Server listening on port 8433. Type 'exit' to quit, or 'cls' to clear...");
    while (true)
    {
        var line = Console.ReadLine();
        if (string.Equals("cls", line, StringComparison.OrdinalIgnoreCase))
        {
            Console.Clear();
            continue;
        }

        if (string.Equals("q", line, StringComparison.OrdinalIgnoreCase))
            return;

[...]

PR #93, opened 2019-08-12
Always close TcpClient at the end of a conversation

matt-richardson:
Previously, a finished conversation would leave the TCPClient around, but GC would clean it up. Now we were tracking Clients so that we could disconnect them, this was keeping a reference to them, causing leaked connections

Fixes https://github.com/OctopusDeploy/Halibut/issues/92

Note: @michaelnoonan - this removes a comment of yours ("we no longer own the stream lifetime") that we think was wrong. Can you :eyes: this closely.

### DIFF
--- a/source/Halibut.Tests/MemoryFixture.cs
+++ b/source/Halibut.Tests/MemoryFixture.cs
@@ -22,7 +22,6 @@ namespace Halibut.Tests
     {
         public long Add(long a, long b)
         {
-            //Console.WriteLine($"Got a request to add {a} + {b}");
             return a + b;
         }
     }
@@ -46,11 +45,16 @@ namespace Halibut.Tests
 
             using (var server = RunServer(Certificates.Octopus, out var port))
             {
+                var expectedTcpClientCount = 1; //server listen = 1 tcpclient
                 //valid requests
                 for (var i = 0; i < NumberOfClients; i++)
                     RunListeningClient(Certificates.TentacleListening, port, Certificates.OctopusPublicThumbprint);
                 for (var i = 0; i < NumberOfClients; i++)
+                {
+                    expectedTcpClientCount++; // each time the server polls, it keeps a tcpclient (as we dont have support to say StopPolling)
                     RunPollingClient(server, Certificates.TentaclePolling, Certificates.TentaclePollingPublicThumbprint);
+                }
+
 #if SUPPORTS_WEB_SOCKET_CLIENT
                 for (var i = 0; i < NumberOfClients; i++)
                     RunWebSocketPollingClient(server, Certificates.TentaclePolling, Certificates.TentaclePollingPublicThumbprint, Certificates.OctopusPublicThumbprint);
@@ -63,19 +67,11 @@ namespace Halibut.Tests
                     GC.WaitForPendingFinalizers();
                 }
 
-                //server listen doesn't keep a port
-                //server polling doesn't keep a port
-                //server websocket polling _keeps_ a port
-                //client listening _keeps_ a port
-                //client polling doesn't keep a port
-                //client websocket polling doesn't keep a port
-
-                const int expectedTcpClientCount = 1 /* server listening */ + NumberOfClients /* number of times the server has setup a poll  */;
                 dotMemory.Check(memory =>
                 {
-                    var objectsCount = memory.GetObjects(x => x.Type.Is<TcpClient>()).ObjectsCount;
-                    Console.WriteLine($"Found {objectsCount} instances of TcpClient still in memory.");
-                    Assert.That(objectsCount, Is.LessThanOrEqualTo(expectedTcpClientCount));
+                    var tcpClientCount = memory.GetObjects(x => x.Type.Is<TcpClient>()).ObjectsCount;
+                    Console.WriteLine($"Found {tcpClientCount} instances of TcpClient still in memory.");
+                    Assert.That(tcpClientCount, Is.EqualTo(expectedTcpClientCount), "Unexpected number of TcpClient objects in memory");
                 });
             }
         }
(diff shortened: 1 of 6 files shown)


source/Halibut/Transport/SecureListener.cs
```diff
@@ -262,12 +252,10 @@ async Task ExecuteRequest(TcpClient client)
...
-                    if (!keepStreamOpen)
-                    {
-                        // Closing an already closed stream or client is safe, better not to leak
-                        stream.Close();
-                        client.Close();
-                    }
+                    // Closing an already closed stream or client is safe, better not to leak
+                    stream.Close();
+                    client.Close();
+                    tcpClientManager.RemoveDisconnectedClient(client);
```
droyad: The `close` statements could throw an exception, so perhaps this should go first.

tothegills: Could `tcpClientManager.RemoveDisconnectedClient(client);` throw?

source/Halibut.Tests/MemoryFixture.cs
```diff
@@ -0,0 +1,181 @@
...
+
+            using (RunServer(Certificates.Octopus, out var port))
+            {
+                for (var i = 0; i < Clients; i++)
+                    RunListeningClient(Certificates.TentacleListening, port, Certificates.OctopusPublicThumbprint);
+                for (var i = 0; i < Clients; i++)
+                    RunPollingClient(Certificates.TentaclePolling, Certificates.TentaclePollingPublicThumbprint);
+#if SUPPORTS_WEB_SOCKET_CLIENT
+                for (var i = 0; i < Clients; i++)
+                    RunWebSocketPollingClient(Certificates.TentaclePolling, Certificates.TentaclePollingPublicThumbprint, Certificates.OctopusPublicThumbprint);
```
matt-richardson: Sometimes this doesn't work. I dont know why.

source/Halibut.Tests/MemoryFixture.cs
```diff
@@ -0,0 +1,181 @@
...
+#endif
+
+                //https://dotnettools-support.jetbrains.com/hc/en-us/community/posts/360000088690-How-reproduce-DotMemory-s-Force-GC-button-s-behaviour-on-code-with-c-?page=1#community_comment_360000072750
+                for (var i = 0; i < 4; i++)
+                {
+                    GC.Collect(2, GCCollectionMode.Forced, true);
+                    GC.WaitForPendingFinalizers();
+                }
+
+                //server listen doesn't keep a port
```
matt-richardson: I have no idea why some of these _sometimes_ keep ports open, and sometimes dont.
Need an assist here.

source/Halibut/Transport/SecureListener.cs
```diff
@@ -230,14 +226,7 @@ async Task ExecuteRequest(TcpClient client)
                         });
 
                         tcpClientManager.AddActiveClient(thumbprint, client);
-                        // Delegate the open stream to the protocol handler - we no longer own the stream lifetime
                         await ExchangeMessages(ssl);
-
-                        if (verifyClientThumbprint(thumbprint))
-                        {
-                            // Mark the stream as delegated once everything has succeeded
-                            keepStreamOpen = true;
```
matt-richardson: This was the bug... When a conversation ended correctly, we were leaving the stream/client open and never closing it.

source/Halibut/Transport/SecureListener.cs
```diff
@@ -230,14 +226,7 @@ async Task ExecuteRequest(TcpClient client)
                         });
 
                         tcpClientManager.AddActiveClient(thumbprint, client);
-                        // Delegate the open stream to the protocol handler - we no longer own the stream lifetime
```
matt-richardson: Removing this comment concerns me a lot. From what RobW and I could tell, once the `ExchangeMessages` call returns, we _do_ own the stream, and should be responsible for cleaning it up.

source/Halibut/Transport/SecureListener.cs
matt-richardson: @tothegills - dont think so...

source/Halibut/Transport/TcpClientManager.cs
```diff
@@ -53,5 +53,24 @@ public IReadOnlyCollection<TcpClient> GetActiveClients(string thumbprint)
 
             return NoClients;
         }
+
+        public void RemoveDisconnectedClient(TcpClient client)
+        {
+            lock (activeClients)
+            {
+                foreach(var activeClient in activeClients)
```
michaelnoonan: From a clarity point of view, I'm finding the naming `activeClient` harder to understand. Reading it through a few times, it appears to be a `KeyValuePair<Thumbprint, TcpClient>`. I wonder if a simple name change could improve this? `kvp`?

source/Halibut/Transport/SecureListener.cs
michaelnoonan: I looked at the `stream.Close()` methods a long time ago, and they appeared safe. However, after all the leak problems we've had, I would feel more confident if there was some super defensive code in here. Either @shane's suggestion, or `try/catch` each statement.

source/Halibut.Tests/MemoryFixture.cs
droyad: How long does it keep it open? There is some small amount of time Windows may keep it unavailable to prevent port hijack attacks.

matt-richardson: bad choice of words. I mean "an instance of `TcpClient` stays in memory"

source/Halibut.Tests/MemoryFixture.cs
matt-richardson: Rob helped my solve this - nothing to see here

Issue #101, opened 2019-09-10
Unhandled exception logged


matt-richardson:
When a client is unreachable, it logs a full exception stacktrace, making it appear this is a completely unexpected situation. We're already specifically catching this and throwing a custom exception - we're just not handling it at the spot where we log.

```
https://example.com:10933/      128  Unexpected exception executing transaction.
Halibut.HalibutClientException: The client was unable to establish the initial connection within 00:01:00
   at Halibut.Transport.TcpClientExtensions.ConnectWithTimeout(TcpClient client, String host, Int32 port, TimeSpan timeout)
   at Halibut.Transport.TcpClientExtensions.ConnectWithTimeout(TcpClient client, Uri remoteUri, TimeSpan timeout)
   at Halibut.Transport.TcpConnectionFactory.CreateConnectedTcpClient(ServiceEndPoint endPoint, ILog log)
   at Halibut.Transport.TcpConnectionFactory.EstablishNewConnection(ServiceEndPoint serviceEndpoint, ILog log)
   at Halibut.Transport.ConnectionManager.<>c__DisplayClass8_0.<CreateNewConnection>b__0()
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.LazyInitValue()
   at System.Lazy`1.get_Value()
   at Halibut.Transport.ConnectionManager.<>c__DisplayClass8_0.<CreateNewConnection>b__1()
   at Halibut.Transport.ConnectionManager.AcquireConnection(IConnectionFactory connectionFactory, ServiceEndPoint serviceEndpoint, ILog log)
   at Halibut.Transport.SecureListeningClient.ExecuteTransaction(Action`1 protocolHandler)
```

We should not log the stack trace here, nor should we call it an "unexpected exception".

PR #102, opened 2019-09-10
Handle "unable to connect" exception nicer

matt-richardson:
Dont log the stack trace on an exception that can happen in a relatively normal situation.

Fixes https://github.com/OctopusDeploy/Halibut/issues/101

### Before
```
https://example.com:10933/      128  Unexpected exception executing transaction.
Halibut.HalibutClientException: The client was unable to establish the initial connection within 00:01:00
   at Halibut.Transport.TcpClientExtensions.ConnectWithTimeout(TcpClient client, String host, Int32 port, TimeSpan timeout)
   at Halibut.Transport.TcpClientExtensions.ConnectWithTimeout(TcpClient client, Uri remoteUri, TimeSpan timeout)
   at Halibut.Transport.TcpConnectionFactory.CreateConnectedTcpClient(ServiceEndPoint endPoint, ILog log)
   at Halibut.Transport.TcpConnectionFactory.EstablishNewConnection(ServiceEndPoint serviceEndpoint, ILog log)
   at Halibut.Transport.ConnectionManager.<>c__DisplayClass8_0.<CreateNewConnection>b__0()
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.LazyInitValue()
   at System.Lazy`1.get_Value()
   at Halibut.Transport.ConnectionManager.<>c__DisplayClass8_0.<CreateNewConnection>b__1()
   at Halibut.Transport.ConnectionManager.AcquireConnection(IConnectionFactory connectionFactory, ServiceEndPoint serviceEndpoint, ILog log)
   at Halibut.Transport.SecureListeningClient.ExecuteTransaction(Action`1 protocolHandler)
```

### After

```
https://example.com:10933/      128  Opening a new connection
https://example.com:10933/      128  The client was unable to establish the initial connection within 00:01:00. Retrying in 1.0 seconds.
https://example.com:10933/      128  Retry attempt 1
https://example.com:10933/      128  Opening a new connection
https://example.com:10933/      128  The client was unable to establish the initial connection within 00:01:00. Retrying in 1.0 seconds.
https://example.com:10933/      128  Retry attempt 2
https://example.com:10933/      128  Opening a new connection
```

### DIFF
--- a/source/Halibut/Transport/SecureListeningClient.cs
+++ b/source/Halibut/Transport/SecureListeningClient.cs
@@ -81,6 +81,12 @@ namespace Halibut.Transport
                     lastError = cex;
                     Thread.Sleep(retryInterval);
                 }
+                catch (HalibutClientException hce)
+                {
+                    lastError = hce;
+                    log.Write(EventType.Error, $"{hce.Message?.TrimEnd('.')}. Retrying in {retryInterval.TotalSeconds:n1} seconds.");
+                    Thread.Sleep(retryInterval);
+                }
                 catch (SocketException sex)
                 {
                     log.WriteException(EventType.Error, $"Socket communication error with connection to  {ServiceEndpoint.Format()}", sex);
--- a/source/Halibut.Tests/FailureModesFixture.cs
+++ b/source/Halibut.Tests/FailureModesFixture.cs
@@ -104,7 +104,7 @@ namespace Halibut.Tests
                 };
                 var echo = octopus.CreateClient<IEchoService>(endpoint);
                 var ex = Assert.Throws<HalibutClientException>(() => echo.Crash());
-                ex.Message.Should().Contain("when sending a request to 'https://google.com:88/', before the request");
+                ex.Message.Should().Be("An error occurred when sending a request to 'https://google.com:88/', before the request could begin: The client was unable to establish the initial connection within 00:00:02.");
             }
         }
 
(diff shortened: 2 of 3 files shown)


source/Halibut/Transport/SecureListeningClient.cs
```diff
@@ -81,6 +81,12 @@ public void ExecuteTransaction(Action<MessageExchangeProtocol> protocolHandler)
                     lastError = cex;
                     Thread.Sleep(retryInterval);
                 }
+                catch (HalibutClientException hce)
+                {
+                    lastError = hce;
+                    log.Write(EventType.Error, $"{hce.Message?.TrimEnd('.')}. Retrying in {retryInterval.TotalSeconds:n1} seconds.");
+                    Thread.Sleep(retryInterval);
```
michaelnoonan: Would there still be any value writing the raw exception as a verbose message? Since this will retry and may eventually give up, perhaps that is when the actual error details will be logged.

matt-richardson: I was actually thinking around this early this morning (or was it late last night?)
I was thinking we should include an error code in each spot we throw / log an exception, so we can trace them better.

I've added a link to this to our [halibut tracking card](https://trello.com/c/KLR8nZwY/1458-halibut-horrors) so we can take it into account if/when we do some work there.

PR #105, opened 2019-10-10
Enable development on *nix environments

jburger:
This PR is intended to provide appropriate framework targeting and compiler directives to allow developers on non-windows environments to work against the majority of the codebase.


build.sh
```diff
@@ -1,101 +1,101 @@
...
-fi
-
-# Start Cake
-if $SHOW_VERSION; then
-    exec mono "$CAKE_EXE" -version
-else
-    exec mono "$CAKE_EXE" $SCRIPT -verbosity=$VERBOSITY -configuration=$CONFIGURATION -target=$TARGET $DRYRUN "${SCRIPT_ARGUMENTS[@]}"
-fi
\ No newline at end of file
+#!/usr/bin/env bash
```
jburger: This file had windows line endings, despite being `for Linux and OS X`.

There is still work to be done for cake to work though, it assumes `mono` at the moment.

source/Halibut.CertificateGenerator/CertificateGenerator.cs
```diff
@@ -78,7 +78,7 @@ static X509Certificate2 AttemptToGenerate(string fullName)
                    };
         }
 
-#if !__MonoCS__
+#if NETFRAMEWORK
```
jburger: Couldn't see an xplat alternative to `CryptoKeySecurity` - there may be more work todo here.

source/Halibut.OctopusSample/Halibut.OctopusSample.csproj
```diff
@@ -45,4 +53,8 @@
     </Compile>
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)'=='netcoreapp2.2'">
```
jburger: Couldn't see a 'neat' way to skip compilation, so I stubbed out the executable instead.

source/Halibut.OctopusSample/Program.cs
```diff
@@ -1,19 +1,26 @@
...
+#if NETFRAMEWORK
         [STAThread]
         static void Main()
         {
             Application.EnableVisualStyles();
             Application.SetCompatibleTextRenderingDefault(false);
             Application.Run(Execute.MainForm = new OctopusForm());
         }
+#endif
+#if NETCOREAPP2_2
```
jburger: > so I stubbed out the executable instead.

Here 'tis

johnsimons:
@jburger given `net45` is only used in samples, I think we can safely upgrade those, or maybe even just have samples in `netcoreapp`?

source/Halibut.Tests/Halibut.Tests.csproj
```diff
@@ -2,13 +2,17 @@
...
     <VersionPrefix>0.0.0</VersionPrefix>
-    <TargetFrameworks>net462;netcoreapp2.2</TargetFrameworks>
     <AssemblyName>Halibut.Tests</AssemblyName>
     <PackageId>Halibut.Tests</PackageId>
     <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
     <GenerateAssemblyTitleAttribute>false</GenerateAssemblyTitleAttribute>
   </PropertyGroup>
-
+  <PropertyGroup Condition="!$([MSBuild]::IsOSUnixLike())">
+	  <TargetFrameworks>netcoreapp2.2;net462</TargetFrameworks>
```
johnsimons: We should bump this to `net472`

johnsimons:
Given how small this project is, do we even need cake at all?

johnsimons:
The `CertificateGenerator` seems to be just for convenience, it doesn't seem to be used in cake either.
I think we can just remove it? maybe replace with multi-plat instructions in readme?

jburger:
Great points there @johnsimons

>  maybe even just have samples in netcoreapp

I like this idea, I'll give it a go. If others need a working net4.5 example perhaps that can be provided via the readme at a later time.

> Given how small this project is, do we even need cake at all?

I suspect not! I'll see what its doing and reason about minimizing that dependency

> The CertificateGenerator seems to be just for convenience

Yeah I personally prefer using `openssl` though I haven't tried in the context of Halibut, I'll give that a try and see what comes of it.

source/create_cert.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
```
johnsimons: I tend to use `#!/bin/bash`, interesting explanation https://unix.stackexchange.com/a/206366

build.cake
```diff
@@ -1,9 +1,11 @@
 //////////////////////////////////////////////////////////////////////
 // TOOLS
 //////////////////////////////////////////////////////////////////////
-#tool "nuget:?package=GitVersion.CommandLine&version=4.0.0"
+#tool "nuget:?package=GitVersion.CommandLine&version=5.0.1"
 #tool "nuget:?package=gitlink&version=3.1.0"
+#if !NETCOREAPP
```
johnsimons: Does this work in cake?

build.cake
```diff
@@ -97,7 +109,17 @@ Task("Test")
         },
         ToolPath = "./tools/JetBrains.dotMemoryUnit.3.0.20171219.105559/lib/tools/dotMemoryUnit.exe"
     });
-});
+}
+void RunTestsWithoutProfiling() {
+    DotNetCoreTest("./source/Halibut.Tests/Halibut.Tests.csproj", new DotNetCoreTestSettings
+    {
+        ArgumentCustomization = args => {
+            args.Append("--configuration=" + configuration);
```
johnsimons: I thought `configuration` is an argument of `DotNetCoreTestSettings` ?

build.cake
jburger: It does :)

build.cake
```diff
@@ -134,6 +153,14 @@ Task("Pack")
     DeleteFiles(artifactsDir + "*symbols*");
 });
 
+void EnableGitLink() {
+    var pdbs = GetFiles($"./source/Halibut/bin/{configuration}/**/Halibut.pdb");
+    foreach(var pdb in pdbs)
+    {
+        GitLink3(pdb);
```
johnsimons: There is a new setting in netcore, I think is `<DebugType>embedded</DebugType>` or something else! Maybe we can use that for netcore builds?

source/Halibut.CertificateGenerator/CertificateGenerator.cs
```diff
@@ -78,7 +78,7 @@ static X509Certificate2 AttemptToGenerate(string fullName)
                    };
         }
 
-#if !__MonoCS__
+#if !__MonoCS__ && !NETCOREAPP
```
johnsimons: I wonder if we need `__MonoCS__` ?

source/Halibut.CertificateGenerator/Halibut.CertificateGenerator.csproj
```diff
@@ -2,24 +2,29 @@
...
     <VersionPrefix>0.0.0</VersionPrefix>
-    <TargetFramework>net45</TargetFramework>
     <AssemblyName>Halibut.CertificateGenerator</AssemblyName>
     <OutputType>Exe</OutputType>
     <PackageId>Halibut.CertificateGenerator</PackageId>
     <GenerateAssemblyTitleAttribute>false</GenerateAssemblyTitleAttribute>
   </PropertyGroup>
 
+  <PropertyGroup Condition="!$([MSBuild]::IsOSUnixLike())">
+	  <TargetFrameworks>net472</TargetFrameworks>
```
johnsimons: Do we actually need this target at all, netcoreapp2.2 works on Windows?

source/Halibut.OctopusSample/Halibut.OctopusSample.csproj
```diff
@@ -2,22 +2,23 @@
...
     <OutputType>Exe</OutputType>
     <PackageId>Halibut.OctopusSample</PackageId>
     <GenerateAssemblyTitleAttribute>false</GenerateAssemblyTitleAttribute>
   </PropertyGroup>
 
-  <ItemGroup>
-    <None Include="App.config" />
-  </ItemGroup>
+  <PropertyGroup Condition="!$([MSBuild]::IsOSUnixLike())">
+    <TargetFramework>net472</TargetFramework>
```
johnsimons: Do we actually need this target at all, netcoreapp2.2 works on Windows?
And I may be mistaken, but won't the net472 pick the halibut.dll from netstandard2 anyway?

build.cake
jburger: Interesting, will check it out

build.cake
jburger: Ah good pickup thanks

source/create_cert.sh
jburger: Ah this was Rider being 'helpful' ... I agree

source/Halibut.Tests/Halibut.Tests.csproj
```diff
@@ -2,13 +2,17 @@
...
     <VersionPrefix>0.0.0</VersionPrefix>
-    <TargetFrameworks>net462;netcoreapp2.2</TargetFrameworks>
     <AssemblyName>Halibut.Tests</AssemblyName>
     <PackageId>Halibut.Tests</PackageId>
     <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
     <GenerateAssemblyTitleAttribute>false</GenerateAssemblyTitleAttribute>
   </PropertyGroup>
-
+  <PropertyGroup Condition="!$([MSBuild]::IsOSUnixLike())">
+	  <TargetFrameworks>netcoreapp2.2;net472</TargetFrameworks>
```
TomPeters: Did we need to upgrade this to net472? I'm actually surprised it was on net462 to start with. I think its better if the test project has the same targets as the project under test, so that you are actually testing the same thing :)

source/Halibut.OctopusSample/Halibut.OctopusSample.csproj
jburger: Alas, this one is a windows form application, which requires the full fat version of .NET.

> but won't the net472 pick the halibut.dll from netstandard2 anyway?
Seems to run on windows OK.

jburger: Arguably this sample isn't of huge value, it could be removed, but I don't think it is hurting anyone to leave it in.

source/Halibut.CertificateGenerator/CertificateGenerator.cs
jburger: I'm not sure of the original intent, seemed safer to leave it alone

source/Halibut.CertificateGenerator/Halibut.CertificateGenerator.csproj
jburger: This code won't work under netcoreapp because of a dependency on `CryptoKeySecurity` and I can't justify the time to find alternatives. So the `netcoreapp` build will output 

```
throw new NotSupportedException("Please refer to the README for alternatives that will run on this platform");
```

jburger: In addition, most of the samples have self signed certificates embedded in them anyway

source/Halibut.Tests/Halibut.Tests.csproj
jburger: Good point - I'll revert it to net45 to line it up with the assembly under test.

PR #107, opened 2019-11-28
Roll back behaviour of SecureListener on Windows.

andrew-moore-octo:
# Background
A customer upgrades from Tentacle version < 4.0.2 to > 4.0.2 and discovered that their performance monitoring on Tentacles was showing an abnormally high number of IO operations/second, even when the Tentacle was idle - usually a difference of 3 or 4 orders of magnitude.

The underlying issue is that `SecureListener` is checking `TcpListener.Pending` within a `SpinWait.WaitUntil`, which is causing a high number of IO operations per second (800 to 1200 per second, depending on the machine) even when idle.

# Results
This PR rolls back the behaviour of `SecureListener` on Windows to how it used to function. Non-Windows behaviour is unchanged, as there are issues with `TcpListener.Stop` on Linux (see https://github.com/dotnet/corefx/issues/26034).

Relates to https://github.com/OctopusDeploy/Issues/issues/6035

# How to review this PR
Some things for consideration:

- The code for Windows and Linux is really similar, but differs in some annoyingly-hard-torefactor ways. I wanted to avoid having completely separate logic for different environments so I've just split out the bits that vary. Is the way that I've done it easy enough to follow? I am worried that it's a little confusing.
- I've added a test to make sure that `SecureListener` isn't hitting the performance counters too hard but it's a tricky one to test. Microsoft recommend waiting 1 second between performance counter readings, so this test is always going to be slow (although it spends most of its time in `Thread.Sleep`. There's also some variability in the numbers so I've added a pretty wide margin of error. Given that performance counters are process-wide there's potential for this to be flaky if tests run in parallel. It also only runs on .NET Framework because performance counters aren't available on .NET Core.


source/Halibut.Tests/SecureListenerFixture.cs
```diff
@@ -0,0 +1,71 @@
...
+                    .Average();
+            }
+
+            var idleAverageWithErrorMargin = idleAverage * 250f;
+            
+            TestContext.Out.WriteLine($"idle average:      {idleAverage} ops/second");
+            TestContext.Out.WriteLine($"listening average: {listeningAverage} ops/second");
+            TestContext.Out.WriteLine($"expectation:     < {idleAverageWithErrorMargin} ops/second");
+
+            listeningAverage.Should().BeLessThan(idleAverageWithErrorMargin);
```
droyad: ❤

source/Halibut/Transport/SecureListener.cs
```diff
@@ -118,49 +115,74 @@ public void Disconnect(string thumbprint)
...
-            var numberOfFailedAttemptsInRow = 0;
-
-            while (!cts.IsCancellationRequested)
-            {
-                try
+            // See: https://github.com/OctopusDeploy/Issues/issues/6035
+            // See: https://github.com/dotnet/corefx/issues/26034
+            var tokenCancellationListener = IsWindows()
+                ? cts.Token.Register(listener.Stop)
+                : (IDisposable) new NoopDisposable();
```
droyad: You can just use `null` here, `using(null)` totally works. I'd also move the declaration closer to the using statement, or even within it `using(IsWindows() ? cts.Token.Register(listener.Stop) : null)`

source/Halibut.Tests/SecureListenerFixture.cs
```diff
@@ -0,0 +1,71 @@
...
+using FluentAssertions;
+using Halibut.Diagnostics;
+using Halibut.Transport;
+using NUnit.Framework;
+
+namespace Halibut.Tests
+{
+    public class SecureListenerFixture
+    {
+#if NETFRAMEWORK
```
droyad: Why only `NETFRAMEWORK`? Pretty soon Halibut will be `netcoreapp` only.

source/Halibut/Transport/SecureListener.cs
```diff
@@ -118,49 +115,74 @@ public void Disconnect(string thumbprint)
...
+            using (tokenCancellationListener)
+            {
+                var numberOfFailedAttemptsInRow = 0;
+                while (!cts.IsCancellationRequested)
+                {
+                    try
                     {
-                        continue;
-                    }
+                        if (shouldAbort())
```
droyad: I think if you write it similar to below, it clarifies what this part of the code does
```
void WaitForPendingConnectionOrCancellation() => SpinWait.SpinUntil(() => cts.IsCancellationRequested || listener.Pending())

// snip

if (!IsWindows())
   WaitForPendingConnectionOrCancellation();

if (cts.IsCancellationRequested)
    return;
```

```

File: source/Halibut/Transport/SecureListener.cs [hunk 5]
```diff
@@ -118,49 +115,74 @@ public void Disconnect(string thumbprint)
...
-            while (!cts.IsCancellationRequested)
-            {
-                try
+            // See: https://github.com/OctopusDeploy/Issues/issues/6035
+            // See: https://github.com/dotnet/corefx/issues/26034
+            var tokenCancellationListener = IsWindows()
+                ? cts.Token.Register(listener.Stop)
+                : (IDisposable) new NoopDisposable();
+            
+            var shouldAbort = IsWindows()
```
droyad: `shouldAbort` doesn't tell me much about what this code does. Also the use of the ternary to return funcs is a bit confusing. I think this could be better written:

```
bool WaitForAnIncommingRequestIfNotWindows()
{
    if (IsWindows())
        return true;
    
    SpinWait.SpinUntil(() => cts.IsCancellationRequested || listener.Pending());
    return !cts.IsCancellationRequested;
}
```

source/Halibut/Transport/SecureListener.cs
andrew-moore-octo: Nice! I'd never tried `using(null)` before. That is good to know!

source/Halibut.Tests/SecureListenerFixture.cs
andrew-moore-octo: This is a weird one. Performance counters work in .net core 2.2 apps, but in .net core 2.2 unit test projects they always return 0. I'll see if I can get it working.

PR #108, opened 2019-12-12
Improve some logging messages

matt-richardson:
Some logging messages weren't so clear. This improves them so we can correlate logs better

image

_tagging @droyad as an fyi_


matt-richardson:
> it's already at the start of the log message

Oh, is that's what the prefix is at the start of the line. I wasn't sure.

PR #109, opened 2019-12-13
Ensure we dont leave unobserved exceptions

matt-richardson:
Unobserved exceptions were happening when:
a) a connection was being attempted to a host that wasn't there
b) the timeout was being hit,
c) after the timeout was hit, the connection attempt fails with an exception

We now ignore any exceptions raised after the timeout is hit

How to reproduce:

1. turn off a Tentacle
1. run a health check on the same Tentacle (that will not respond)
1. put a breakpoint in the code somewhere (say in `FilterEnvironmentsByProjectAndChannelRule`)
1. trigger an action that will hit the breakpoint (ie, viewing the Environments page)
-> the ide will break at the break point
1. wait at the break point for a while (say 3min) (ie, as if you were trying to debug whats going on)
1. hit F5 to run the code again
-> the TaskScheduler.UnobservedTaskException in OctopusProgram will be hit with:
```
System.NullReferenceException: Object reference not set to an instance of an object.
   at System.Net.Sockets.TcpClient.EndConnect(IAsyncResult asyncResult)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncCoreLogic(IAsyncResult iar, Func`2 endFunction, Action`1 endAction, Task`1 promise, Boolean requiresSynchronization)
```

### DIFF
--- a/source/Halibut.Tests/Util/AsyncEx/TaskExtensionsFixture.cs
+++ b/source/Halibut.Tests/Util/AsyncEx/TaskExtensionsFixture.cs
@@ -35,7 +35,9 @@ namespace Halibut.Tests.Util.AsyncEx
             });
             Func<Task> act = async () => await task.TimeoutAfter(TimeSpan.FromMilliseconds(100), CancellationToken.None);
             act.ShouldThrow<TimeoutException>();
-            triggered.Should().Be(false, "the task should have been aborted");
+            triggered.Should().Be(false, "we should have stopped waiting on the task when timeout happened");
+            await Task.Delay(200);
+            triggered.Should().Be(true, "task should have continued executing in the background");
         }
         
         [Test]
@@ -57,7 +59,9 @@ namespace Halibut.Tests.Util.AsyncEx
             });
             Func<Task> act = async () => await task.TimeoutAfter(TimeSpan.FromMilliseconds(150), cancellationTokenSource.Token);
             act.ShouldThrow<TaskCanceledException>();
-            triggered.Should().Be(false, "the task should have been aborted");
+            triggered.Should().Be(false, "we should have stopped waiting on the task when cancellation happened");
+            await Task.Delay(200);
+            triggered.Should().Be(true, "task should have continued executing in the background (not entirely ideal, but this task is designed to handle non-cancelable tasks)");
         }
         
         [Test]
--- a/source/Halibut.Tests/Util/AsyncEx/TaskExtensionsFixture.cs
+++ b/source/Halibut.Tests/Util/AsyncEx/TaskExtensionsFixture.cs
@@ -1,0 +1,117 @@
+using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Halibut.Util.AsyncEx;
+using NUnit.Framework;
+
+namespace Halibut.Tests.Util.AsyncEx
+{
+    [TestFixture]
+    public class TaskExtensionsFixture
+    {
+        [Test]
+        public async Task When_TaskCompletesWithinTimeout_TaskCompletesSuccessfully()
+        {
+            var triggered = false;
+            var task = Task.Run(async () =>
+            {
+                await Task.Delay(TimeSpan.FromMilliseconds(100));
+                triggered = true;
+            });
+            await task.TimeoutAfter(TimeSpan.FromMilliseconds(300), CancellationToken.None);
+            triggered.Should().Be(true, "the task should have triggered");
+        }
+        
+        [Test]
+        public async Task When_TaskDoesNotCompleteWithinTimeout_ThrowsTimeoutException()
+        {
+            var triggered = false;
+            var task = Task.Run(async () =>
+            {
+                await Task.Delay(TimeSpan.FromMilliseconds(200));
+                triggered = true;
+            });
+            Func<Task> act = async () => await task.TimeoutAfter(TimeSpan.FromMilliseconds(100), CancellationToken.None);
+            act.ShouldThrow<TimeoutException>();
+            triggered.Should().Be(false, "the task should have been aborted");
+        }
+        
+        [Test]
+        public async Task When_TaskGetsCancelled_ThrowsTaskCanceledException()
+        {
+            var triggered = false;
+            var cancellationTokenSource = new CancellationTokenSource();
+
+            Task.Run(async () =>
+            {
+                await Task.Delay(TimeSpan.FromMilliseconds(100));
+                cancellationTokenSource.Cancel();
+            });
+            
+            var task = Task.Run(async () =>
+            {
+                await Task.Delay(TimeSpan.FromMilliseconds(200));
+                triggered = true;
+            });
+            Func<Task> act = async () => await task.TimeoutAfter(TimeSpan.FromMilliseconds(150), cancellationTokenSource.Token);
+            act.ShouldThrow<TaskCanceledException>();
+            triggered.Should().Be(false, "the task should have been aborted");
+        }
+        
+        [Test]
+        public async Task When_TaskThrowsExceptionAfterTimeout_ExceptionsAreObserved()
+        {
+            await VerifyNoUnobservedExceptions<TimeoutException>(Task.Run(async () =>
+            {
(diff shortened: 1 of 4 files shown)


source/Halibut/Util/AsyncEx/TaskExtensions.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// <param name="timeout">The amount of time to wait until we give up</param>
+        /// <param name="cancellationToken">Supports task cancellation</param>
+        /// <returns>The task if successful, otherwise a TimeoutException or OperationCanceledException</returns>
+        /// <exception cref="TimeoutException">If the timeout gets reached before the task completes</exception>
+        /// <exception cref="OperationCanceledException">The task was cancelled via the cancellation token</exception>
+        public static async Task TimeoutAfter(this Task task, TimeSpan timeout, CancellationToken cancellationToken)
+        {
+            var timeOutTask = Task.Delay(timeout);
+            var cancellationTask = cancellationToken.AsTask();
+            var timeoutCancellation = new CancellationTokenSource();
```
michaelnoonan: This should be disposed: https://docs.microsoft.com/en-us/dotnet/api/system.threading.cancellationtokensource.dispose?view=netframework-4.8

source/Halibut/Util/AsyncEx/TaskExtensions.cs
```diff
@@ -0,0 +1,73 @@
...
+    {
+        /// <summary>
+        /// Waits for a) the task to complete, b) the timeout to pass or c) the cancellationToken to be triggered.
+        /// Wraps it all up so that we dont get UnobservedTaskExceptions 
+        /// </summary>
+        /// <param name="task">The task to wait on</param>
+        /// <param name="timeout">The amount of time to wait until we give up</param>
+        /// <param name="cancellationToken">Supports task cancellation</param>
+        /// <returns>The task if successful, otherwise a TimeoutException or OperationCanceledException</returns>
+        /// <exception cref="TimeoutException">If the timeout gets reached before the task completes</exception>
```
michaelnoonan:
```suggestion
        /// <exception cref="TimeoutException">If the timeout gets reached before the task completes or is cancelled</exception>
```

source/Halibut/Util/AsyncEx/TaskExtensions.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// <exception cref="OperationCanceledException">The task was cancelled via the cancellation token</exception>
+        public static async Task TimeoutAfter(this Task task, TimeSpan timeout, CancellationToken cancellationToken)
```
michaelnoonan: A common pattern I've seen is to offer an optional `CancellationToken` argument, which defaults to a `CancellationToken.None` (never cancel). It depends on the use-cases you want to promote.
```suggestion
        public static async Task TimeoutAfter(this Task task, TimeSpan timeout, CancellationToken cancellationToken = null)
```

source/Halibut/Util/AsyncEx/TaskExtensions.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            var timeOutTask = Task.Delay(timeout);
```
michaelnoonan: This can be a neat trick. I have seen it done before, but IIRC we had to change our implementation in a production system. We changed it to `using (var timeoutCancellationTokenSource = new CancellationTokenSource(timeout))` https://docs.microsoft.com/en-us/dotnet/api/system.threading.cancellationtokensource.-ctor?view=netframework-4.8#System_Threading_CancellationTokenSource__ctor_System_TimeSpan_

Instead of both a `Task.Delay(timeout)` and a `CancellationTokenSource`, both which consume unmanaged resources, you only have the `CancellationTokenSource` which you dispose with the `using` statement.

source/Halibut/Util/AsyncEx/TaskExtensions.cs
matt-richardson: I mis-interpreted your version, so I've taken it and expanded further 👍

source/Halibut/Util/AsyncEx/TaskExtensions.cs
matt-richardson: I'm tempted to leave it for now, and let future requirements drive it.

source/Halibut/Transport/TcpClientExtensions.cs
```diff
@@ -1,46 +1,58 @@
...
         public static void ConnectWithTimeout(this TcpClient client, Uri remoteUri, TimeSpan timeout)
         {
-            client.ConnectWithTimeout(remoteUri.Host, remoteUri.Port, timeout);
+            ConnectWithTimeout(client, remoteUri.Host, remoteUri.Port, timeout);
         }
 
         public static void ConnectWithTimeout(this TcpClient client, string host, int port, TimeSpan timeout)
         {
-            var connectResult = false;
+            Connect(client, host, port, timeout, CancellationToken.None).GetAwaiter().GetResult();
```
matt-richardson: This changes in a subsequent to actually pass through the cancellation token.

source/Halibut/Util/AsyncEx/TaskExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// <returns>The task if successful, otherwise a TimeoutException or OperationCanceledException</returns>
+        /// <exception cref="TimeoutException">If the timeout gets reached before the task completes or before the task is cancelled</exception>
+        /// <exception cref="OperationCanceledException">The task was cancelled via the cancellation token</exception>
+        public static async Task TimeoutAfter(this Task task, TimeSpan timeout, CancellationToken cancellationToken)
+        {
+            var timeoutTask = Task.Delay(timeout, cancellationToken);
+            var wrappedTask = AwaitAndSwallowExceptionsWhenTimedOut(task, timeoutTask);
+            await Task.WhenAny(wrappedTask, timeoutTask);
+            
+            if (cancellationToken.IsCancellationRequested)
```
TomPeters: replace with `cancellationToken.ThrowIfCancellationRequested()`

source/Halibut/Util/AsyncEx/TaskExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+                await task;
+            }
+            catch (Exception)
+            {
+                if (!timeoutTask.IsCompleted)
+                    throw;
+            }
+        }
+        
+        static Task AsTask(this CancellationToken cancellationToken)
```
TomPeters: This is now unused - let's delete it

source/Halibut.Tests/Util/AsyncEx/TaskExtensionsFixture.cs
```diff
@@ -0,0 +1,117 @@
...
+        {
+            var triggered = false;
+            var task = Task.Run(async () =>
+            {
+                await Task.Delay(TimeSpan.FromMilliseconds(200));
+                triggered = true;
+            });
+            Func<Task> act = async () => await task.TimeoutAfter(TimeSpan.FromMilliseconds(100), CancellationToken.None);
+            act.ShouldThrow<TimeoutException>();
+            triggered.Should().Be(false, "the task should have been aborted");
```
TomPeters: There's a race condition here. The underlying task does not "get aborted" and this test presumably only passes because the assertion is executed before the underlying task sets `triggered` to `true`.

source/Halibut.Tests/Util/AsyncEx/TaskExtensionsFixture.cs
```diff
@@ -0,0 +1,117 @@
...
+            });
+            
+            var task = Task.Run(async () =>
+            {
+                await Task.Delay(TimeSpan.FromMilliseconds(200));
+                triggered = true;
+            });
+            Func<Task> act = async () => await task.TimeoutAfter(TimeSpan.FromMilliseconds(150), cancellationTokenSource.Token);
+            act.ShouldThrow<TaskCanceledException>();
+            triggered.Should().Be(false, "the task should have been aborted");
```
TomPeters: Same race condition as above.

PR #119, opened 2020-07-16
Add delay before all retry attempts...

uglybugger:
... to avoid attempting to DDoS an Octopus server or offend intermediate firewalls

### DIFF
--- a/source/Halibut/Transport/SecureClient.cs
+++ b/source/Halibut/Transport/SecureClient.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Diagnostics;
 using System.IO;
-using System.Net.Security;
 using System.Net.Sockets;
 using System.Security.Authentication;
 using System.Security.Cryptography.X509Certificates;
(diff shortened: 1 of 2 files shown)


source/Halibut/Transport/SecureClient.cs
```diff
@@ -47,7 +45,11 @@ public void ExecuteTransaction(Action<MessageExchangeProtocol> protocolHandler,
             var watch = Stopwatch.StartNew();
             for (var i = 0; i < ServiceEndpoint.RetryCountLimit && retryAllowed && watch.Elapsed < ServiceEndpoint.ConnectionErrorRetryTimeout; i++)
             {
-                if (i > 0) log.Write(EventType.Error, "Retry attempt {0}", i);
+                if (i > 0)
+                {
+                    Thread.Sleep(retryInterval);
+                    log.Write(EventType.OpeningNewConnection, $"Retrying connection to {ServiceEndpoint.Format()} - attempt #{i}.");
```
emilol, line 51: nitpick: string interpolation here prevents structured logging.

uglybugger: Agreed. We have no structured logging here, though - it's all string interpolation :( We need us a Serilog, we do.

uglybugger:
https://github.com/OctopusDeploy/Issues/issues/6482

PR #122, opened 2020-07-28
Convert retry policy to exponential backoff

uglybugger:
This PR converts the retry policy from phased to a simple bounded exponential back-off.

The intention is to more-rapidly reduce load on Octopus Server instances with large numbers of polling tentacles when network instability or other connection issues cause tentacles to make multiple successive retry attempts.


source/Halibut.sln.DotSettings
```diff
@@ -257,4 +257,5 @@ II.2.12 &lt;HandlesEvent /&gt;&#xD;
...
...ofile name="Halibut"&gt;&lt;AspOptimizeRegisterDirectives&gt;True&lt;/AspOptimizeRegisterDirectives&gt;&lt;HtmlReformatCode&gt;True&lt;/HtmlReformatCode&gt;&lt;CSArrangeThisQualifier&gt;True&lt;/CSArrangeThisQualifier&gt;&lt;CSRemoveCodeRedundancies&gt;True&lt;/CSRemoveCodeRedundancies&gt;&lt;CSUseAutoProperty&gt;True&lt;/CSUseAutoProperty&gt;&lt;CSMakeFieldReadonly&gt;True&lt;/CSMakeFieldReadonly&gt;&lt;CSUseVar&gt;&lt;BehavourStyle&gt;CAN_CHANGE_TO_IMPLICIT&lt;/BehavourStyle&gt;&lt;LocalVariableStyle&gt;ALWAYS_IMPLICIT&lt;/LocalVariableStyle&gt;&lt;ForeachVariableStyle&gt;ALWAYS_IMPLICIT&lt;/ForeachVariableStyle&gt;&lt;/CSUseVar&gt;&lt;CSUpdateFileHeader&gt;True&lt;/CSUpdateFileHeader&gt;&lt;CSOptimizeUsings&gt;&lt;OptimizeUsings&gt;True&lt;/OptimizeUsings&gt;&lt;EmbraceInRegion&gt;False&lt;/EmbraceInRegion&gt;&lt;RegionName&gt;&lt;/RegionName&gt;&lt;/CSOptimizeUsings&gt;&lt;CSShortenReferences&gt;True&lt;/CSShortenReferences&gt;&lt;CSReformatCode&gt;True&lt;/CSReformatCode&gt;&lt;CSharpFormatDocComments&gt;True&lt;/CSharpFormatDocComments&gt;&lt;CSReorderTypeMembers&gt;True&lt;/CSReorderTypeMembers&gt;&lt;XMLReformatCode&gt;True&lt;/XMLReformatCode&gt;&lt;/Profile&gt;</s:String>
```
uglybugger, line 260: Adds "backoff" to the team dictionary.

source/Halibut/Util/RetryPolicy.cs
```diff
@@ -0,0 +1,63 @@
+using System;
+using System.Diagnostics;
+
+namespace Halibut.Util
+{
+    class RetryPolicy
```
emilol, line 6: nitpick: this could be called "ExponentialBackoffWithJitterRetryPolicy" to give folks a hint what's going on here.

source/Halibut/Transport/PollingClient.cs
```diff
@@ -46,37 +46,43 @@ public void Start()
...
                 {
-                    retry.Try();
-                    secureClient.ExecuteTransaction(protocol =>
+                    try
+                    {
+                        retry.Try();
+                        secureClient.ExecuteTransaction(protocol => { protocol.ExchangeAsSubscriber(subscription, handleIncomingRequest); }, cancellationToken);
+                        retry.Success();
+                    }
+                    finally
```
MarkSiedle, line 68: Reading through the previous `while` loop, this appears to address the _"don't DDoS ourselves when we're successful"_ scenario, while also providing a consistent way to get the next sleep period when there _is_ an exception 👍

When last we spoke, we were confused as to whether there was something higher up the chain providing backoff/sleep, but it looks like this has just been constantly hammering things?

source/Halibut.sln.DotSettings
emilol: "backoff r# I just want to type 'backoff'!"

source/Halibut/Util/RetryPolicy.cs
```diff
@@ -0,0 +1,63 @@
...
+            var elapsed = stopwatch.Elapsed;
+            var delay = CalculateDelay(elapsed);
+            return delay;
+        }
+
+        internal TimeSpan CalculateDelay(TimeSpan elapsed)
+        {
+            var multiple = 1d + BackoffMultiplier;
+            var delay = TimeSpan.FromMilliseconds(elapsed.TotalMilliseconds * multiple);
+            delay = delay > MinimumDelay
```
MarkSiedle, line 53: This is a good way of making sure your value is within a range 👍 (I would have most likely done some nested-ternary nastiness to jam it all on one line, but I really like this approach) :pipe-thinking-wizard:

source/Halibut.Tests/RetryPolicyFixture.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System;
+using Halibut.Util;
+using NUnit.Framework;
+
+namespace Halibut.Tests
+{
+    public class RetryPolicyFixture
```
MarkSiedle, line 7: 💯

source/Halibut/Util/RetryPolicy.cs
uglybugger: It could, but as far as the consumer is concerned it shouldn't know what the retry policy is. If we had an interface as the abstraction then I'd be more inclined to make the concrete class's name a bit more implementation-specific, but at the moment the consuming code just news up a `RetryPolicy` so it shouldn't really care what the implementation is.

If only there were a way we could get something else to provide dependencies to classes...

source/Halibut/Transport/PollingClient.cs
uglybugger: I still don't know :( I'm having a diabolical time with nested retry loops trying to get test instances to fail in the desired manner. I'm hoping that even if this change doesn't address the entirety of the issue, it will get most of the worst of it.

source/Halibut/Util/RetryPolicy.cs
uglybugger: Thank you :)

For what it's worth, under the covers the compiler will optimise it to look pretty much what the nested ternary would have done anyway. I'm happy to optimise for humans to read and let the compiler optimise it for machines to read :)

PR #123, opened 2020-11-05
Support for developing halibut on non-windows OSes

TomPeters:
This unblocks us from being able to develop halibut on non-windows OSes.

We still need to target net45 for now because Tentacle is still built against net45 to support older windows OSes.


andrew-moore-octo (approved):
LGTM 👍

PR #124, opened 2020-11-18
Fix polling disconnects

matt-richardson:
When polling clients (tentacles) disconnect, the server can end up getting an error that implies a comms error:

```
The step failed: Activity Run a Script on EC2AMAZ-25701AL-Polling failed with error 'The archive entry was compressed using an unsupported compression method. Server exception: System.IO.InvalidDataException: The archive entry was compressed using an unsupported compression method.
  at System.IO.Compression.Inflater.Inflate(FlushCode flushCode)
  at System.IO.Compression.Inflater.ReadInflateOutput(Byte* bufPtr, Int32 length, FlushCode flushCode, Int32& bytesRead)
  at System.IO.Compression.Inflater.ReadOutput(Byte* bufPtr, Int32 length, Int32& bytesRead)
  at System.IO.Compression.Inflater.InflateVerified(Byte* bufPtr, Int32 length)
  at System.IO.Compression.DeflateStream.ReadCore(Span`1 buffer)
  at System.IO.Compression.DeflateStream.Read(Byte[] array, Int32 offset, Int32 count)
  at System.IO.BinaryReader.InternalRead(Int32 numBytes)
  at System.IO.BinaryReader.ReadInt32()
  at Newtonsoft.Json.Bson.BsonDataReader.ReadNormal()
  at Newtonsoft.Json.Bson.BsonDataReader.Read()
  at Newtonsoft.Json.JsonReader.ReadForType(JsonContract contract, Boolean hasConverter)
  at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.Deserialize(JsonReader reader, Type objectType, Boolean checkAdditionalContent)
  at Newtonsoft.Json.JsonSerializer.DeserializeInternal(JsonReader reader, Type objectType)
  at Newtonsoft.Json.JsonSerializer.Deserialize[T](JsonReader reader)
  at Halibut.Transport.Protocol.MessageExchangeStream.ReadBsonMessage[T]()
  at Halibut.Transport.Protocol.MessageExchangeStream.Receive[T]()
  at Halibut.Transport.Protocol.MessageExchangeProtocol.ProcessReceiverInternalAsync(IPendingRequestQueue pendingRequests, RequestMessage nextRequest)'.
```
This has been tracked to:
* a clean shutdown of the server
* which sends an `END` control message
* which in common use cases, the `MessageExchangeProtocol` isn't expecting, as it's actually expecting a expected deflated bson message

This is a WIP PR that throws a nicer error when this happens.

In Octopus server, due to timing, this can manifest as:
* a failed task with a message `Connection ended by remote`
* a task that has an apparent pause in the middle (if the tentacle comes back on line within the timeout (usually 2 minutes))
* a task that fails with a polling timeout (if the tentacle doesn't come back online)

Still outstanding:
 - [ ] tests for this scenario
 - [ ] improve the `BufferedReadStream` so that it doesn't buffer the whole message, just the first 10? 100? characters.
 - [ ] perf and memory tests
 - [ ] disable the logging of bytes & json by default, but allow it to be enabled for diagnostics

### DIFF
--- a/source/Halibut/Transport/Protocol/MessageExchangeStream.cs
+++ b/source/Halibut/Transport/Protocol/MessageExchangeStream.cs
@@ -15,6 +15,13 @@ namespace Halibut.Transport.Protocol
 {
     public class MessageExchangeStream : IMessageExchangeStream
     {
+        const string MxClient = "MX-CLIENT";
+        const string Next = "NEXT";
+        const string Proceed = "PROCEED";
+        const string End = "END";
+        const string MxSubscriber = "MX-SUBSCRIBER";
+        const string MxServer = "MX-SERVER";
+        static readonly string[] ControlMessages = {MxClient, MxSubscriber, MxServer, Next, Proceed, End};
         readonly Stream stream;
         readonly ILog log;
         readonly StreamWriter streamWriter;
@@ -37,9 +44,9 @@ namespace Halibut.Transport.Protocol
 
         public void IdentifyAsClient()
         {
+            log.Write(EventType.Diagnostic, $"Sent: {MxClient} {currentVersion}");
             log.Write(EventType.Diagnostic, "Identifying as a client");
-            streamWriter.Write("MX-CLIENT ");
-            streamWriter.Write(currentVersion);
+            streamWriter.Write($"{MxClient} {currentVersion}");
             streamWriter.WriteLine();
             streamWriter.WriteLine();
             streamWriter.Flush();
@@ -48,8 +55,9 @@ namespace Halibut.Transport.Protocol
 
         public void SendNext()
         {
+            log.Write(EventType.Diagnostic, $"Sent: {Next}");
             SetShortTimeouts();
-            streamWriter.Write("NEXT");
+            streamWriter.Write(Next);
             streamWriter.WriteLine();
             streamWriter.Flush();
             SetNormalTimeouts();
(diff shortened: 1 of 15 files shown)


source/Halibut/Transport/Protocol/MessageExchangeStream.cs
```diff
@@ -37,9 +44,9 @@ public MessageExchangeStream(Stream stream, ILog log)
 
         public void IdentifyAsClient()
         {
+            log.Write(EventType.Diagnostic, $"Sent: {MxClient} {currentVersion}");
```
matt-richardson, line 47: This logging should be extracted to a method (that logs (if enabled) and writes to the stream)

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
```diff
@@ -246,14 +260,51 @@ void ExpectServerIdentity()
...
-                return (T)messageEnvelope.Message;
+                try
+                {
+                    using (var zip = new DeflateStream(buffer, CompressionMode.Decompress, true))
+                    {
+                        using (var bson = new BsonDataReader(zip) {CloseInput = false})
+                        {
+                            var messageEnvelope = serializer.Deserialize<MessageEnvelope>(bson);
+
+                            log.Write(EventType.Diagnostic, $"ReadBsonMessage: {BitConverter.ToString(buffer.GetBytes())}");
```
matt-richardson, line 273: We want to wrap these log messages in an "Is env var set" type thing. This diagnostics would have been so useful for this. We also want to update the readme for this.
Need to liaise with @OctopusDeploy/school-security to make sure this isn't A Bad Thing :tm:

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
```diff
@@ -246,14 +260,51 @@ void ExpectServerIdentity()
...
+                                throw new Exception("messageEnvelope is null");
+                            return (T) messageEnvelope.Message;
+                        }
+                    }
+                }
+                catch (Exception ex)
+                {
+                    var plaintext = SafelyGetPlainText(buffer);
+                    if (plaintext.Equals(End))
+                        throw new HalibutClientException("Connection ended by remote.");
```
matt-richardson, line 285: I'm on the fence about re-using these exceptions. On one hand, we (appear to) get nice transient error handling automatically in Octopus. On the other hand, it might be nicer to be specific.
Either way, we want to:
a) suppress the stack trace
b) include an error code of some description to make it easy to track down where the error occured

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
```diff
@@ -355,4 +411,107 @@ void SetShortTimeouts()
             stream.ReadTimeout = (int)HalibutLimits.TcpClientHeartbeatReceiveTimeout.TotalMilliseconds;
         }
     }
+
+    class BufferedReadStream : Stream
```
matt-richardson, line 415: We should extract to a different file

andrew-moore-octo: I've pulled these out into their own files, renamed them from "BufferXStream" to "CaptureXStream" to better reflect what they do, and added some tests.

matt-richardson:
Closing as very stale

PR #125, opened 2020-11-23
Pull out some constants.

andrew-moore-octo:
This PR extracts some important strings out to constants in `MessageExchangeStream`.

### DIFF
--- a/source/Halibut/Transport/Protocol/MessageExchangeStream.cs
+++ b/source/Halibut/Transport/Protocol/MessageExchangeStream.cs
@@ -110,7 +110,7 @@ namespace Halibut.Transport.Protocol
                 case End:
                     return false;
                 default:
-                    throw new ProtocolException($"Expected {Next} or {End}, got:  " + line);
+                    throw new ProtocolException($"Expected {Next} or {End}, got: " + line);
             }
         }
 
--- a/source/Halibut/Transport/Protocol/MessageExchangeStream.cs
+++ b/source/Halibut/Transport/Protocol/MessageExchangeStream.cs
@@ -15,6 +15,13 @@ namespace Halibut.Transport.Protocol
 {
     public class MessageExchangeStream : IMessageExchangeStream
     {
+        const string Next = "NEXT";
+        const string Proceed = "PROCEED";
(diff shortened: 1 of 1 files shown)


source/Halibut/Transport/Protocol/MessageExchangeStream.cs
```diff
@@ -143,10 +149,7 @@ async Task<string> ReadLineAsync()
 
         public void IdentifyAsSubscriber(string subscriptionId)
         {
-            streamWriter.Write("MX-SUBSCRIBER ");
-            streamWriter.Write(currentVersion);
-            streamWriter.Write(" ");
-            streamWriter.Write(subscriptionId);
+            streamWriter.Write($"{MxSubscriber} {currentVersion} {subscriptionId}");
```
matt-richardson, line 152: I wonder if we should make this a `WriteLine` instead? (and drop the `streamWriter.WriteLine();` on the next line)

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
```diff
@@ -156,8 +159,7 @@ public void IdentifyAsSubscriber(string subscriptionId)
 
         public void IdentifyAsServer()
         {
-            streamWriter.Write("MX-SERVER ");
-            streamWriter.Write(currentVersion.ToString());
+            streamWriter.Write($"{MxServer} {currentVersion}");
```
matt-richardson, line 162: I wonder if we should make this a WriteLine instead?

matt-richardson (approved):
Nice one. A couple of minor suggestions.

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
andrew-moore-octo: Agreed. The next PR targets this area so I'll do it as part of that PR.

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
andrew-moore-octo: Will do this in the next PR.

PR #126, opened 2020-11-23
Refactor control/identity messages.

andrew-moore-octo:
This PR serves two purposes:

- While working in Halibut recently, we realised that the `NEXT`, `END` and `PROCEED`, `MX-CLIENT`, `MX-SUBSCRIBER` and `MX-SERVER` messages didn't have any name to collectively describe them. This made talking about these things hard. This PR introduces some terminology:
  - `MX-CLIENT`, `MX-SERVER` and `MX-SUBSCRIBER` are referred to as _identity_ messages
  - `NEXT`, `END` and `PROCEED` are referred to as _control_ messages
  - Not part of this PR, but we've taken to referring to the rest as _payload_ messages.
- Armed with this new terminology, this PR refactors some of the duplicate code out of `MessageExchangeStream` and hopefully improves the readability.

# How to review this PR
- The code should be functionally the same
- General code quality
- Tests should be green

### DIFF
--- a/source/Halibut/Transport/Protocol/MessageExchangeStream.cs
+++ b/source/Halibut/Transport/Protocol/MessageExchangeStream.cs
@@ -45,42 +45,50 @@ namespace Halibut.Transport.Protocol
         public void IdentifyAsClient()
         {
             log.Write(EventType.Diagnostic, "Identifying as a client");
-            streamWriter.Write($"{MxClient} {currentVersion}");
-            streamWriter.WriteLine();
+            SendIdentityMessage($"{MxClient} {currentVersion}");
+            ExpectServerIdentity();
+        }
+
+        void SendControlMessage(string message)
+        {
+            streamWriter.WriteLine(message);
+            streamWriter.Flush();
+        }
+
+        async Task SendControlMessageAsync(string message)
+        {
+            await streamWriter.WriteLineAsync(message);
+            await streamWriter.FlushAsync();
+        }
+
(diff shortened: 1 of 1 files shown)


source/Halibut/Transport/Protocol/MessageExchangeStream.cs
```diff
@@ -45,42 +45,50 @@ public MessageExchangeStream(Stream stream, ILog log)
...
+            streamWriter.Flush();
+        }
+
+        async Task SendControlMessageAsync(string message)
+        {
+            await streamWriter.WriteLineAsync(message);
+            await streamWriter.FlushAsync();
+        }
+
+        void SendIdentityMessage(string identityLine)
```
Waldo000000, line 64: Ok, so I'm inferring from this that it's important (for some reason) for "identity" messages (but not "control" messages) to be followed by an extra newline. Have I got that right?

Waldo000000 (approved):

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
andrew-moore-octo: Yep! I don't know why, and I haven't looked into it, but it's been that way for 6 years, so I've left it as-is. One thought that came to mind is that it might be for some conceptual similarity to HTTP, or left as a possible extension point (.e.g everything before a blank line is connection/identity metadata).

PR #129, opened 2020-12-04
Async rework

richev-o:
PR raised to get branch build

### DIFF
--- a/source/Halibut/Transport/SecureListener.cs
+++ b/source/Halibut/Transport/SecureListener.cs
@@ -143,9 +143,7 @@ namespace Halibut.Transport
                         }
 
                         var client = listener.AcceptTcpClient();
-#pragma warning disable CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
-                        HandleClient(client);
-#pragma warning restore CS4014
+                        Task.Run(async () => await HandleClient(client).ConfigureAwait(false)).ConfigureAwait(false);
                         numberOfFailedAttemptsInRow = 0;
                     }
                     catch (SocketException e) when (e.SocketErrorCode == SocketError.Interrupted)
@@ -153,6 +151,7 @@ namespace Halibut.Transport
                     }
                     catch (ObjectDisposedException)
                     {
+                        // Happens on shutdown
                     }
                     catch (Exception ex)
                     {
@@ -190,7 +189,7 @@ namespace Halibut.Transport
                 client.ReceiveTimeout = (int)HalibutLimits.TcpClientReceiveTimeout.TotalMilliseconds;
 
                 log.Write(EventType.ListenerAcceptedClient, "Accepted TCP client: {0}", client.Client.RemoteEndPoint);
-                await ExecuteRequest(client);
+                await ExecuteRequest(client).ConfigureAwait(false);
             }
             catch (ObjectDisposedException)
             {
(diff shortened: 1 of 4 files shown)


source/Halibut/Transport/SecureWebSocketListener.cs
```diff
@@ -86,12 +84,12 @@ async Task Accept()
                 {
                     try
                     {
-                        var context = await listener.GetContextAsync();
+                        var context = await listener.GetContextAsync().ConfigureAwait(false);
 
                         if (context.Request.IsWebSocketRequest)
                         {
 #pragma warning disable CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
```
matt-richardson, line 91: Is this still relevant? I saw it disappear in `SecureListener.cs `, so slight question about the difference

source/Halibut/Transport/SecureWebSocketListener.cs
```diff
@@ -152,10 +150,10 @@ async Task ExecuteRequest(HttpListenerContext listenerContext)
                     return;
                 }
 
-                if (await Authorize(listenerContext, clientName))
+                if (await Authorize(listenerContext, clientName).ConfigureAwait(false))
```
matt-richardson, line 153: Consider pulling this into a variable. Looks a tad confusing now.

matt-richardson (approved):
Looks reasonable. A couple of comments. Can you add a detailed description to the PR for future us?

source/Halibut/Transport/SecureListener.cs
```diff
@@ -143,16 +143,15 @@ void WaitForPendingConnectionOrCancellation()
                         }
 
                         var client = listener.AcceptTcpClient();
-#pragma warning disable CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
-                        HandleClient(client);
-#pragma warning restore CS4014
+                        Task.Run(async () => await HandleClient(client).ConfigureAwait(false)).ConfigureAwait(false);
```
johnsimons, line 146: I don't think we need `ConfigureAwait(false)` here, given we are not awaiting

johnsimons: Which brings me to untrackable tasks. I think we need to track this dangling tasks

richev-o: Can you clarify what you mean please @johnsimons ?

source/Halibut/Transport/SecureWebSocketListener.cs
richev-o: It is, because we're in an `async Task` method here, whereas in the equivalent location in `SecureListener.cs` the containing method is just `void`.

source/Halibut/Transport/SecureListener.cs
johnsimons: See https://github.com/OctopusDeploy/Halibut/pull/36/files#diff-26fa9bedf8a3cc85fcb9b46910dfe94d21c3a269b5f53a26cdacac8c928f7656R61, there I am tracking the returned task, so I can dispose of it correctly at shutdown time

source/Halibut/Transport/SecureWebSocketListener.cs
droyad: We could make `SecureListener` async... but I don't want to poke the bear right now.

source/Halibut/Transport/SecureListener.cs
droyad: This article says we don't need to: https://devblogs.microsoft.com/pfxteam/do-i-need-to-dispose-of-tasks/#:~:text=medium%2Dlength%20answer%3A-,%E2%80%9CNo.,to%20meet%20your%20performance%20goals.

And stack overflow answer: https://stackoverflow.com/a/3734298/10784

Taub works(ed) on the .NET parallel tasks team apparently

droyad (approved):

PR #133, opened 2021-08-11
Limiting methods that can be invoked

andrew-at-octopus:
Found that methods that could be invoked was not limited to those on the interface, but included other methods such as the object.Equals method.  Used an extension method which limits this to only intended methods.


andrew-at-octopus:
[Remote Code Execution via Deserialisation in Halibut Protocol](https://trello.com/c/EZa6ulAd/507-remote-code-execution-via-deserialisation-in-halibut-protocol)

MissedTheMark (approved):
LGTM 👍

PR #134, opened 2021-08-12
Registering Protocol Request and Response message types

andrew-at-octopus:
Tentacle doesn't work, as the RequestMessage was not deserializable.
This should allow those messages to deserialize.


MissedTheMark (approved):
LGTM 👍

PR #135, opened 2021-08-13
Enhance Halibut tests, fix client and protocol

andrew-at-octopus:
When using a polling Client, the message is parsed before being passed to the client.  As such, all types on created clients need to be available for deserialization.  Previously only registered service types were being included.  This worked because the tests used the same IServiceFactory for both client and server, so both knew the types and the tests passed.  There was also no test passing a non-standard object as a property or response, so the type binder was not being verified properly.

Also I found that mixing a new message source (eg Server) which was not including the $type property in the JSON, could not be deserialized by the old message receiver, which, missing the type (eg "$type":"Octopus.Shared.Contracts.ScriptTicket, Octopus.Shared"), would deserialize it as JObject.  So for now all messages sent need to include the $type property like they used to, while the new reciever can specify the type to the deserializer if the $type is present without incident.

### DIFF
--- a/source/Halibut.Tests/TestServices/SupportedServices.cs
+++ b/source/Halibut.Tests/TestServices/SupportedServices.cs
@@ -95,6 +95,7 @@ namespace Halibut.Tests.TestServices
 
         public MapLocation GetLocation(MapLocation loc)
         {
+            // Swap the latitude and longitude for the round trip verification... never know where you will end up! 
             return new MapLocation { Latitude = loc.Longitude, Longitude = loc.Latitude };
         }
     }
--- a/source/Halibut.Tests/TestServices/SupportedServices.cs
+++ b/source/Halibut.Tests/TestServices/SupportedServices.cs
@@ -92,5 +92,10 @@ namespace Halibut.Tests.TestServices
         {
             return "Hello tuple";
         }
+
+        public MapLocation GetLocation(MapLocation loc)
+        {
+            return new MapLocation { Latitude = loc.Longitude, Longitude = loc.Latitude };
+        }
     }
 }
--- a/source/Halibut.Tests/UsageFixture.cs
+++ b/source/Halibut.Tests/UsageFixture.cs
@@ -5,10 +5,10 @@ using System.IO;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
-using System.Net.Security;
 using System.Security.Cryptography.X509Certificates;
 using System.Threading.Tasks;
 using FluentAssertions;
+using FluentAssertions.Common;
 using Halibut.ServiceModel;
 using Halibut.Tests.TestServices;
 using NUnit.Framework;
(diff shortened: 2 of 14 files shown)


source/Halibut.Tests/UsageFixture.cs
```diff
@@ -86,23 +86,25 @@ public void OctopusCanSendMessagesToPollingTentacle()
...
 
                     tentaclePolling.Poll(new Uri("poll://SQ-TENTAPOLL"), new ServiceEndPoint(new Uri($"wss://localhost:{octopusPort}/Halibut"), Certificates.SslThumbprint));
 
-                    var echo = octopus.CreateClient<IEchoService>("poll://SQ-TENTAPOLL", Certificates.TentaclePollingPublicThumbprint);
-                    for (var i = 0; i < 2000; i++)
+                    var echo = octopus.CreateClient<ISupportedServices>("poll://SQ-TENTAPOLL", Certificates.TentaclePollingPublicThumbprint);
+                    for (var i = 1; i < 100; i++)
                     {
-                        echo.SayHello("Deploy package A" + i).Should().Be("Deploy package A" + i + "...");
+                        var i1 = i;
```
MissedTheMark, line 106: Do we need i1? Can't we just use i?

source/Halibut.Tests/TestServices/SupportedServices.cs
```diff
@@ -92,5 +92,10 @@ public string Ambiguous(string a, Tuple<string, string> b)
         {
             return "Hello tuple";
         }
+
+        public MapLocation GetLocation(MapLocation loc)
+        {
+            return new MapLocation { Latitude = loc.Longitude, Longitude = loc.Latitude };
```
MissedTheMark, line 98: Latitude = loc.Longitude?

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
```diff
@@ -315,7 +315,10 @@ void WriteBsonMessage<T>(T messages)
             using (var zip = new DeflateStream(stream, CompressionMode.Compress, true))
             using (var bson = new BsonDataWriter(zip) { CloseOutput = false })
             {
-                serializer.Serialize(bson, new MessageEnvelope<T> { Message = messages });
+                // for the moment this MUST be object so that the $type property is included
+                // If it is not, then an old receiver (eg, old tentacle) will not be able to understand messages from a new sender (server)
```
MissedTheMark, line 319: Good pickup.  Would be awful to suddenly not support old receivers!

MissedTheMark (approved):
LGTM overall.  Couple of minor questions - Happy for this to be merged once you look at those questions, and as long as it definitely works ok for older tentacles etc.

source/Halibut.Tests/UsageFixture.cs
andrew-at-octopus: It's either that or tell resharper to ignore the capture variable scope warning thing... this seemed easier.

source/Halibut.Tests/TestServices/SupportedServices.cs
andrew-at-octopus: Yeah, that's the "change" for the round-trip verification - I'll add a comment.

PR #138, opened 2021-11-14
Change HalibutRuntime to use IPendingRequestQueue interface instead of implementation 

emilol:
...and add a HalibutRuntimeBuilder so that we can provide an alternate, queue implementation.

### DIFF
--- a/source/Halibut.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress.NETCore.approved.cs
+++ b/source/Halibut.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress.NETCore.approved.cs
@@ -287,7 +287,7 @@ namespace Halibut.ServiceModel
         public void ApplyResponse(Halibut.Transport.Protocol.ResponseMessage response) { }
         public Halibut.Transport.Protocol.RequestMessage Dequeue() { }
         public Task<Halibut.Transport.Protocol.RequestMessage> DequeueAsync() { }
-        public Halibut.Transport.Protocol.ResponseMessage QueueAndWait(Halibut.Transport.Protocol.RequestMessage request, System.Threading.CancellationToken cancellationToken) { }
+        public Task<Halibut.Transport.Protocol.ResponseMessage> QueueAndWaitAsync(Halibut.Transport.Protocol.RequestMessage request, System.Threading.CancellationToken cancellationToken) { }
     }
     public interface IPendingRequestQueueFactory
     {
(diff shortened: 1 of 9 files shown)


source/Halibut.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress.NETCore.approved.cs
```diff
@@ -318,6 +333,7 @@ public class PendingRequestQueue : Halibut.ServiceModel.IPendingRequestQueue
         public Task<Halibut.Transport.Protocol.RequestMessage> DequeueAsync() { }
         public Halibut.Transport.Protocol.ResponseMessage QueueAndWait(Halibut.Transport.Protocol.RequestMessage request) { }
         public Halibut.Transport.Protocol.ResponseMessage QueueAndWait(Halibut.Transport.Protocol.RequestMessage request, System.Threading.CancellationToken cancellationToken) { }
```
uglybugger, line 335: Happy to mark both of these as `[Obsolete]`?

source/Halibut/HalibutRuntime.cs
```diff
@@ -48,17 +49,29 @@ public HalibutRuntime(IServiceFactory serviceFactory, X509Certificate2 serverCer
         {
```
uglybugger, line 49: Can we mark the old public constructors as `[Obsolete]` and have them delegate behaviour to the new, internal one?

source/Halibut/HalibutRuntimeBuilder.cs
```diff
@@ -0,0 +1,56 @@
...
+
+namespace Halibut
+{
+    public class HalibutRuntimeBuilder
+    {
+        ILogFactory logFactory = new LogFactory();
+        IPendingRequestQueueFactory queueFactory;
+        X509Certificate2 serverCertificate;
+        IServiceFactory serviceFactory = new NullServiceFactory();
+        ITrustProvider trustProvider = new DefaultTrustProvider();
```
uglybugger, line 14: Can we use `??=` to initialise these in the `Build` method rather than creating instances we potentially don't need? This will also discourage other builder methods from messing with instance fields which may be subsequently discarded.

source/Halibut/ServiceModel/PendingRequestQueue.cs
```diff
@@ -47,6 +47,12 @@ public ResponseMessage QueueAndWait(RequestMessage request, CancellationToken ca
             return pending.Response;
         }
 
+        public Task<ResponseMessage> QueueAndWaitAsync(RequestMessage request, CancellationToken cancellationToken)
```
uglybugger, line 50: Can we make this one async so that the call stack is preserved if something goes 💥 ?

uglybugger (approved):
Some suggestions inline. LGTM after those are addressed one way or another 👍

source/Halibut/HalibutRuntime.cs
```diff
@@ -181,8 +194,14 @@ public TService CreateClient<TService>(ServiceEndPoint endpoint, CancellationTok
             return proxy;
 #endif
         }
-
+        
+        // TODO: async HalibutProxy
         ResponseMessage SendOutgoingRequest(RequestMessage request, CancellationToken cancellationToken)
+        {
+            return SendOutgoingRequestAsync(request, cancellationToken).GetAwaiter().GetResult();
```
adam-mccoy, line 201: [Sync over async](https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#warning-sync-over-async). Might make sense to implement a legit sync version of this method.

emilol: One Day ™️ As discussed: an async HalibutProxy isn't within the scope of these changes. I've added a note for future us.

source/Halibut/HalibutRuntime.cs
emilol: Obsolete tags done 👍 

Added a note about why we can't just delegate to make it clearer what's going on here.

PR #140, opened 2022-03-17
Support delayed creation of DataStream.

LukeButters:
Support creating a DataStream from a `Lazy<DataStream>`, allowing for the `Length` or `writer` to be created at the time either is required.

### DIFF
--- a/source/Halibut.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress.NETFramework.approved.cs
+++ b/source/Halibut.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress.NETFramework.approved.cs
@@ -17,13 +17,14 @@ namespace Halibut
     {
         public DataStream() { }
         public DataStream(long length, Action<Stream> writer) { }
-        public DataStream(Func<Halibut.DataStream> delayedStreamCreation) { }
         public Guid Id { get; set; }
         public long Length { get; set; }
         private Action<Stream> writer {  }
         public bool Equals(Halibut.DataStream other) { }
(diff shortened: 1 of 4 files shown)


source/Halibut/DataStream.cs
```diff
@@ -20,14 +29,56 @@ public DataStream(long length, Action<Stream> writer)
...
+
+        private void EnsureLengthAndStreamInitializedFromDelayedCreation()
+        {
+            if (_delayedStreamCreation != null)
+            {
+                try
+                {
+                    var r = _delayedStreamCreation();
+                    this._writer = r.writer;
+                    this._length = r.Length;
```
emilol, line 74: This looks suspiciously like a `DataStream` mapper. If we want to map between two DataStreams, can we do that outside of the dto class, so that we're no mixing concerns?

source/Halibut.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress.NETFramework.approved.cs
```diff
@@ -17,8 +17,10 @@ public class DataStream : IEquatable<Halibut.DataStream>
     {
         public DataStream() { }
         public DataStream(long length, Action<Stream> writer) { }
+        public DataStream(Guid guid, Func<Halibut.DataStream> delayedStreamCreation) { }
```
emilol, line 20: similar to previous comment: looks like we're passing a `DataStream` to a `DataStream` ctor here, which makes me wonder if this new requirement is being introduced at the right abstraction. Could we achieve something similar by making the calling code accept a `Lazy<DataStream>` instead?

Issue #141, opened 2022-04-13
Exceptions in the DataStream writer causes the receiver to block with 100% CPU load.


LukeButters:
The issue can be replicated with the following:

-  An agreed upon service that sends a DataStream.
- The remote (tentacle) is setup as polling.
- The server sends a `DataStream` that will throw for example: `new DataStream(10000, stream => throw new Exception("Oh noes"))`

The result appears to be that the remote wont service any more requests and the CPU will go to 100%.

The issue also happens when the remote is listening, the CPU will go above 100% until some thread limit is exhausted and then will stop servicing any more requests.

Issue #117, opened 2020-06-26
Strange issue when using WebSockets


willemda:
Hello,

First of all, we've been using Halibut for a long time and we love it! (See our PR regarding routing)

We've recently switched to using WebSockets as it enable us sharing the port with other components running on the same machine easily.

From that moment, we've experienced some strange issues where our Polling Client would suddenly lose connectivity (sometimes having 100% CPU).

Here's what our setup looks like:
![image](https://user-images.githubusercontent.com/642347/85841215-0f70fc00-b79e-11ea-9d29-97a7f3a175d2.png)

The Tentacle is a windows service running under net 4.8 (running latest standard Halibut, no routing)
The Octopus is a windows service running under netcoreapp3.1 (running our Halibut routing)

Logs and dump/trace analysis on the Tentacle suggests that 3 of the PollingThread seems stuck forever (no mention of reconnection or activity in the logs).

We're wondering if it's something you've also experienced since you're using Halibut on such a large infrastructure?

We've also noticed the following issue that looks very similar to what we're experiencing but are unsure if this is applicable to our setup: https://github.com/dotnet/runtime/issues/16810

We do have dumps, logs & dotTrace exports to help if needed, we've tried a lot of things but couldn't figure out where this come from with certitude.

janpieterz:
Regarding the above, we're running websockets for 14 days without issue now. We will switch our whole production fleet end of the month and, if that stays stable we can close. No clear insights into the cause of this issue, back then a couple of other small changes were made but none that have any reasonable connection to the websocket connections.

jburger:
Thanks for letting us know @janpieterz keep us posted :crossed_fingers:

willemda:
Looks you guys found the issue a while ago already so I'll close. 
I've been following the past month's commits and there are really cool updates being done, love it 👍

PR #142, opened 2022-04-13
Fixes a bug caused by exceptions in DataStreams.

LukeButters:
This change fixes a bug where the remote would have 100% CPU load and
eventually stop servicing request when the sender throws an exception in
the `DataStream`.

The issue was the return value of the `read()` method was not
respected. The doc for this method says:
```
<returns>The number of bytes read into <paramref name="buffer" />. This might be less than the number of bytes requested if that many bytes are not available, or it might be zero if the end of the stream is reached.</returns>
```
The fix is to simply detect a return value of `0` indicating the stream
is closed rather than continue to try and read from the stream.

Fixes: https://github.com/OctopusDeploy/Halibut/issues/141

May be related to: https://github.com/OctopusDeploy/Halibut/issues/117

### DIFF
--- a/source/Halibut/Transport/Protocol/MessageExchangeStream.cs
+++ b/source/Halibut/Transport/Protocol/MessageExchangeStream.cs
@@ -261,13 +261,15 @@ namespace Halibut.Transport.Protocol
         TemporaryFileStream CopyStreamToFile(Guid id, long length, BinaryReader reader)
         {
             var path = Path.Combine(Path.GetTempPath(), string.Format("{0}_{1}", id.ToString(), Interlocked.Increment(ref streamCount)));
+            long bytesLeftToRead = length;
             using (var fileStream = new FileStream(path, FileMode.Create, FileAccess.Write))
             {
                 var buffer = new byte[1024 * 128];
-                while (length > 0)
+                while (bytesLeftToRead > 0)
                 {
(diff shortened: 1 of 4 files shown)


sburmanoctopus (approved):
This looks like it should solve the problem. Nice work.

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
```diff
@@ -261,13 +261,15 @@ void ReadStream(StreamCapture capture)
...
             using (var fileStream = new FileStream(path, FileMode.Create, FileAccess.Write))
             {
                 var buffer = new byte[1024 * 128];
-                while (length > 0)
+                while (bytesLeftToRead > 0)
                 {
-                    var read = reader.Read(buffer, 0, (int)Math.Min(buffer.Length, length));
-                    length -= read;
+                    var read = reader.Read(buffer, 0, (int)Math.Min(buffer.Length, bytesLeftToRead));
+                    if (read == 0) throw new ProtocolException($"Stream with length {length} was closed after only reading {length - bytesLeftToRead} bytes.");
```
LukeButters, line 271: This exception is the same type as the one thrown above.

zentron (approved):
From F&M point of view this looks good 👍

PR #143, opened 2022-04-22
Fix a performance regression in Halibut by re-using the newtonsoft ContractResolver

LukeButters:
Newton soft recommends that `ContractResolver`s should be re-used see the [performance doc](https://www.newtonsoft.com/json/help/html/Performance.htm). Also the [newtonsoft example](https://www.newtonsoft.com/json/help/html/contractresolver.htm) of how to implement a contract resolver, which shows creating a static instance of the `ContractResolver`.

I created a simple test that made 1k requests to send just a Small DataStream to a polling tentacle the speed up was:
```
before: 47.7s
after: 7.6s
```

In [this change](https://github.com/OctopusDeploy/Halibut/commit/b325aeb4cce5a9ff0b663c67ccabb3bd5d04cb34) Halibut had a regression where its speed was reduced to fix a bug. I created a test `ParallelRequestsFixture` which shows the concurrent editing issue when we re-use `ContractResolver` with the existing code:

image

With the locking, the test passes.

### DIFF
--- a/source/Halibut.Tests/TestServices/IReadDataSteamService.cs
+++ b/source/Halibut.Tests/TestServices/IReadDataSteamService.cs
@@ -2,8 +2,6 @@ namespace Halibut.Tests.TestServices
 {
     public interface IReadDataSteamService
     {
-        long SendData(DataStream dataStream);
-
-        long SendDataMany(DataStream[] dataStreams);
+        long SendData(params DataStream[] dataStreams);
     }
 }
--- a/source/Halibut.Tests/TestServices/IReadDataSteamService.cs
+++ b/source/Halibut.Tests/TestServices/IReadDataSteamService.cs
@@ -3,5 +3,7 @@ namespace Halibut.Tests.TestServices
     public interface IReadDataSteamService
     {
         long SendData(DataStream dataStream);
+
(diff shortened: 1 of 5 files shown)


source/Halibut.Tests/TestServices/IReadDataSteamService.cs
```diff
@@ -3,5 +3,7 @@ namespace Halibut.Tests.TestServices
     public interface IReadDataSteamService
     {
         long SendData(DataStream dataStream);
+
+        long SendDataMany(DataStream[] dataStreams);
```
emilol, line 7: suggestion: Is there a reason this needs to be a separate additional API? I suspect we could achieve the same thing and maintain backwards compatibility using `params`

```suggestion
        long SendData(params DataStream[] dataStream);
```

LukeButters: Done, thank you!

tothegills (approved):

LukeButters:
I made a mistake the performance is previously worse, I took those measurements with another performance optimisation (a buffer on the stream). The github message is updated with the correct measurements. The change here is still better.

I will create a new PR after this one with the buffered stream optimisation.

Issue #145, opened 2022-05-12
PR #144 to add buffered stream appears to cause timeout issues.


LukeButters:
It looks like #144 has caused some timeout issues.

This was seen in the octopus server build.

A similar issues appears to occur when the buffer size is set to 40MB, that results in test `StreamsCanBeSentToPolling()` never terminating. To reproduce on windows disable `net462`.

PR #144, opened 2022-04-28
Use a buffered stream for higher performance.

LukeButters:
For sending 1000 small files the speed up us `7.4s` to `4s`.

The code to measure this is:

```
public class SpeedTest
    {
        [Test]
        public void FailWhenServerThrowsDuringADataStreamOnPolling2()
        {
            var services = new DelegateServiceFactory();
            services.Register<IReadDataSteamService>(() => new ReadDataStreamService());

            using (var octopus = new HalibutRuntime(Certificates.Octopus))
            using (var tentaclePolling = new HalibutRuntime(services, Certificates.TentaclePolling))
            {
                var octopusPort = octopus.Listen();
                octopus.Trust(Certificates.TentaclePollingPublicThumbprint);

                tentaclePolling.Poll(new Uri("poll://SQ-TENTAPOLL"), new ServiceEndPoint(new Uri("https://localhost:" + octopusPort), Certificates.OctopusPublicThumbprint));

                var echo = octopus.CreateClient<IEchoService>("poll://SQ-TENTAPOLL", Certificates.TentaclePollingPublicThumbprint);

                var readDataSteamService = octopus.CreateClient<IReadDataSteamService>("poll://SQ-TENTAPOLL", Certificates.TentacleListeningPublicThumbprint);

                var ds = DataStream.FromString("Hello how are ye");
                Stopwatch watch = new Stopwatch();
                watch.Start();
                for (int i = 0; i < 1000; i++)
                {
                    readDataSteamService.SendData(ds);
                    ;
                }
                Console.WriteLine(watch.Elapsed.TotalMilliseconds);

                long recieved = readDataSteamService.SendData(DataStream.FromString("hello"));
                recieved.Should().Be(5);
            }
        }
    }
```

PR #146, opened 2022-05-12
Revert "Use a buffered stream for higher performance."

LukeButters:
Reverts OctopusDeploy/Halibut#144

To fix https://github.com/OctopusDeploy/Halibut/issues/145


zentron (approved):
👍 Back to safety

PR #147, opened 2022-05-24
Convert cake to nuke

N-lson:
✔️ Works with a local octopus server/tentacle updated to Halibut `4.4.783-pull-147`

Changes:

- Removed GitLink. [GitLink](https://github.com/GitTools/GitLink) doesn't seem to be maintained and I'm unsure it's actually doing anything in the current cake build - see [TeamCity log](https://build.octopushq.com/buildConfiguration/OctopusDeploy_Halibut_HalibutBuildTestAndPackageNetCore/4487334?buildTab=log&focusLine=4732&logView=flowAware&linesState=4732) of a recent build on master. It outputs a bunch of warnings about classes being missing or changed.
- Added [Source Link](https://github.com/dotnet/sourcelink) which is an updated package which does the same thing as GitLink.
- Updated `UsageFixture` to get the DNS host name instead of computer name from environment
- Updated `MemoryFixture` to retry for 10 seconds if the number of clients is over the threshold

Package metadata with GitLink on master (missing repository information and incorrect links for classes):
image
image

Package metadata with Source Link on this branch:
image
image

TODO still:

- Update readme

TODO post-merge:

1. Disable trigger for old build chain, enable trigger for new build chain
2. Switch default branch to main
3. Add branch protection/pr rules


build/Build.cs
```diff
@@ -0,0 +1,142 @@
...
+                .SetInformationalVersion(OctoVersionInfo.InformationalVersion)
+                .EnableNoRestore()
+                .CombineWith(ss => ss
+                    .SetProjectFile(Solution.Halibut))
+                .CombineWith(ss => ss
+                    .SetProjectFile(Solution.Halibut_Tests)));
+        });
+
+    [PublicAPI]
+    Target TestWindows => _ => _
```
N-lson, line 86: I've separated windows/linux so that we can still run the dotmemory tests on windows, however dotmemory isn't supported on linux so we just run a `dotnet test` instead.

build/Build.cs
```diff
@@ -0,0 +1,142 @@
...
+
+    Target Restore => _ => _
+        .DependsOn(Clean)
+        .Executes(() =>
+        {
+            DotNetRestore(s => s
+                .CombineWith(ss => ss
+                    .SetProjectFile(Solution.Halibut))
+                .CombineWith(ss => ss
+                    .SetProjectFile(Solution.Halibut_Tests)));
```
N-lson, line 67: Don't want to restore/build the sample projects. Most of them can't build on linux.

ankithkonda (changes requested):
Looks like only netstandard is being build here. there's a difference in package size between cake build and this one 
Screen Shot 2022-05-31 at 5 03 12 pm
Screen Shot 2022-05-31 at 5 02 33 pm

source/Halibut.Tests/MemoryFixture.cs
```diff
@@ -65,12 +66,25 @@ public void TcpClientsAreDisposedCorrectly()
                     GC.WaitForPendingFinalizers();
                 }
 
-                dotMemory.Check(memory =>
+                var stopwatch = Stopwatch.StartNew();
```
N-lson, line 69: Retrying in `MemoryFixture`

ankithkonda (approved):
Nice work! 🏆

PR #149, opened 2022-05-26
re-use JSON library static defaults where possible

emilol:
Currently we new up a new `ContractResolver` for every request to Halibut. This PR brings the serializer into line with the way the base resolvers work, exposing a static instance that can be reused, in order to cache contracts.

### DIFF
--- a/source/Halibut/Transport/Protocol/MessageSerializer.cs
+++ b/source/Halibut/Transport/Protocol/MessageSerializer.cs
@@ -10,17 +10,14 @@ namespace Halibut.Transport.Protocol
 {
     public class MessageSerializer : IMessageSerializer
     {
-        static readonly HalibutContractResolver halibutContractResolver = new HalibutContractResolver();
-        
         readonly RegisteredSerializationBinder binder = new RegisteredSerializationBinder();
-
         readonly HashSet<Type> messageContractTypes = new HashSet<Type>();
         
         JsonSerializer CreateSerializer()
         {
             var jsonSerializer = JsonSerializer.Create();
             jsonSerializer.Formatting = Formatting.None;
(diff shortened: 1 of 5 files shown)


source/Halibut/Transport/Protocol/MessageSerializer.cs
```diff
@@ -10,17 +10,14 @@ namespace Halibut.Transport.Protocol
...
         readonly RegisteredSerializationBinder binder = new RegisteredSerializationBinder();
-
         readonly HashSet<Type> messageContractTypes = new HashSet<Type>();
         
         JsonSerializer CreateSerializer()
         {
             var jsonSerializer = JsonSerializer.Create();
             jsonSerializer.Formatting = Formatting.None;
-            jsonSerializer.ContractResolver = halibutContractResolver;
+            jsonSerializer.ContractResolver = HalibutContractResolver.Instance;
```
emilol, line 20: This line is the main reason this PR exists; to reuse the cache of types we've seen before. The binder will still reject contracts we haven't explicitly allowed, but this should mean the cache can be reused between individual requests.

LukeButters: The `halibutContractResolver` was already a static and doing so made a measurable improvement  https://github.com/OctopusDeploy/Halibut/pull/143 measured using the code in this https://github.com/OctopusDeploy/Halibut/pull/144 reverted PR.

Is it even faster now?

emilol: So I think what happened was I started from a stale master, thought I was fixing things, then after a rebase turns out you'd solved the same problem already 😄 I'll close the PR. The syntactic changes are already up in another branch I'm working on.

PR #151, opened 2022-05-27
Add `IMessageSerializer` to `HalibutRuntimeBuilder` 

emilol:
This allows consumers of this library to add additional config to the message serializer, e.g:

```csharp
void Configure(JsonSerializerSettings settings)
{
    settings.Converters = new List<JsonConverter>
    {
        customConverter
    };
}

var halibutRuntime = new HalibutRuntimeBuilder()
    .WithMessageSerializer(builder => builder.WithSerializerSettings(Configure))
    .Build();
```

### DIFF
--- a/source/Halibut.Tests/RegisteredSerializationBinderFixture.cs
+++ b/source/Halibut.Tests/RegisteredSerializationBinderFixture.cs
@@ -12,7 +12,8 @@ namespace Halibut.Tests
         [Test]
         public void BindMethods_WithValidClass_FindsAllMethodTypes()
         {
-            var binder = new RegisteredSerializationBinder();
+            var typeRegistry = new TypeRegistry();
+            var binder = new RegisteredSerializationBinder(typeRegistry);
             binder.Register(typeof(IExampleService));
             binder.BindToName(typeof(ExampleProperties), out var assemblyName, out var typeName);
             var t = binder.BindToType(assemblyName, typeName);
@@ -70,7 +71,8 @@ namespace Halibut.Tests
         [TestCase(typeof(IObjectExampleService))]
         public void Register_WithInvalidTypes_WillThrow(params Type[] types)
         {
-            var binder = new RegisteredSerializationBinder();
+            var typeRegistry = new TypeRegistry();
+            var binder = new RegisteredSerializationBinder(typeRegistry);
             Assert.Throws<TypeNotAllowedException>(() => { binder.Register(types); });
         }
         
@@ -97,7 +99,8 @@ namespace Halibut.Tests
         [Test]
         public void Circular_Types_CanBeResolved()
         {
-            var binder = new RegisteredSerializationBinder();
+            var typeRegistry = new TypeRegistry();
+            var binder = new RegisteredSerializationBinder(typeRegistry);
             binder.Register(typeof(IMCircular));
             binder.BindToName(typeof(CircularPart1), out var assemblyName1, out var typeName1);
             var t1 = binder.BindToType(assemblyName1, typeName1);
--- a/source/Halibut/HalibutRuntime.cs
+++ b/source/Halibut/HalibutRuntime.cs
@@ -32,6 +32,7 @@ namespace Halibut
         string friendlyHtmlPageContent = DefaultFriendlyHtmlPageContent;
         Dictionary<string, string> friendlyHtmlPageHeaders = new Dictionary<string, string>();
         readonly IMessageSerializer messageSerializer;
+        readonly ITypeRegistry typeRegistry;
 
         [Obsolete]
         public HalibutRuntime(X509Certificate2 serverCertificate) : this(new NullServiceFactory(), serverCertificate, new DefaultTrustProvider())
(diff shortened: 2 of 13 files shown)


emilol:
@uglybugger this is the PR we discussed in standup; do you mind taking a look a the APIs we're exposing?

uglybugger (approved):
As discussed - looks good to me 👍

source/Halibut/Transport/Protocol/RegisteredSerializationBinder.cs
```diff
@@ -1,109 +1,38 @@
...
-
-            lock (allowedTypes)
-            {
-                if (!allowedTypes.Add(type))
-                {
-                    // Seen this before, no need to go further
-                    return;
-                }
-            }
+        readonly Type[] protocolTypes = { typeof(ResponseMessage), typeof(RequestMessage) };
```
sburmanoctopus, line 9: It feels like these protocol types should be part of the type registry, as it's a type concern and that's what the registry is all about.

I know registering it here means we know those types exist. That's the plus of keeping it here.

But also, it feels a little out of place that we are taking this registry, and manipulating it in the constructor. It feels like an expected consequence.

source/Halibut/HalibutRuntime.cs
```diff
@@ -192,8 +197,8 @@ public TService CreateClient<TService>(ServiceEndPoint endpoint)
         
         public TService CreateClient<TService>(ServiceEndPoint endpoint, CancellationToken cancellationToken)
         {
-            messageSerializer.AddToMessageContract(typeof(TService));
-            
+            typeRegistry.AddToMessageContract(typeof(TService));
```
sburmanoctopus, line 200: The only reason the typeRegistry exists as a member variable here is so that we can ensure the service exists here.

The second 'non-obsolete' constructor expects a type registry, and this class expects it to be used in the messageSerializer.

That doesn't quite feel right to me, as someone could plug in the wrong type registry.

source/Halibut/Transport/Protocol/ITypeRegistry.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace Halibut.Transport.Protocol
+{
+    public interface ITypeRegistry
```
sburmanoctopus, line 5: The name `ITypeRegistry` feels like it has more purpose beyond serialisation. Does it have a future beyond serialization?

If not, maybe a different name might suffice? `ISeriazableTypeRegistry`?

source/Halibut/Transport/Protocol/ITypeRegistry.cs
```diff
@@ -0,0 +1,12 @@
...
+    {
+        void Register(params Type[] registeredServiceTypes);
```
sburmanoctopus, line 7: The 3 methods `Register`, `RegisterType`, and `AddToMessageContract` are all fancy ways of registering a type.

It's a little unclear what each is for looking at the interface. Except maybe `RegisterType`

A suggestion is to maybe rename `Register` to something like `RegisterTypesFromService`?

source/Halibut/Transport/Protocol/ITypeRegistry.cs
```diff
@@ -0,0 +1,12 @@
...
+        void RegisterType(Type type, string path, bool ignoreObject);
+        bool IsInAllowedTypes(Type type);
```
sburmanoctopus, line 9: The name `IsInAllowedTypes` is a little confusing, because without looking into it, the reader wonders what an 'allowed type' is.

Some suggestions:
 - Contains (keeping with standard .NET method names)
 - CanSerialize
 - IsRegistered (although this makes more sense if we rename the interface)

source/Halibut/Transport/Protocol/ITypeRegistry.cs
```diff
@@ -0,0 +1,12 @@
...
+        void AddToMessageContract(params Type[] types);
```
sburmanoctopus, line 10: I wonder if the `AddToMessageContract` and `Register` methods should be combined (and named appropriately).

`AddToMessageContract` is just a wrapper over `Register`. The only difference is that it locks `messageContractTypes` first, and only registers types that are not in the `messageContractTypes` list already.

The `messageContractTypes` list isn't used for anything else, so it's just an optimisation.

I can't help but wonder if we simply make `Register` do the checks that `AddToMessageContract` does. 

It won't affect the way the system behaves contractually. But it will be less confusing for future developers.

source/Halibut.Tests/RegisteredSerializationBinderFixture.cs
```diff
@@ -12,7 +12,8 @@ public class RegisteredSerializationBinderFixture
         [Test]
         public void BindMethods_WithValidClass_FindsAllMethodTypes()
         {
-            var binder = new RegisteredSerializationBinder();
+            var typeRegistry = new TypeRegistry();
```
sburmanoctopus, line 15: A test builder  for making a `RegisteredSerializationBinder` maybe?

sburmanoctopus (approved):
LGTM. This is a nice refactor, and will help us achieve what we need in the SQL Halibut queue.

There are some comments around naming etc., but once discussed, nothing that should stop the merge.

We now just want to get feedback from @uglybugger about expanding the contract of this package.

emilol:
@sburmanoctopus since there's nothing blocking in your review, I'm going to go ahead and get this merged so it doesn't go stale. I've had a look over the comments and they largely seem to be around improvements that would require non-additive and/or breaking changes. 

Let's aim to chat about it in stand-up tomorrow, I think it's a good direction, but as a starting point we'd really like to get this in without impacting any existing users.

PR #158, opened 2022-07-01
Only rewind decorate the message exchange on the listening side.

skermajo:
# Background

When https://github.com/OctopusDeploy/Halibut/pull/154 was introduced, it addressed issues with the `DeflateStream` buffer on the listening side of the Halibut connection. While it was only to address listeners, tests showed no harm having the client-side include the same fix to keep things simple.

This PR removes the the `DeflateStream` fix from the client-side, as it's not necessary to be there.

# Results

Halibut `SecureListener` classes now explicitly use the `DeflateStream` fix from from https://github.com/OctopusDeploy/Halibut/pull/154 inside their `MessageExchangeStream` . All other uses of `MessageExchangeStream` do not include the fix.

# How to review this PR
👀 
Quality :heavy_check_mark:

# Pre-requisites


ankithkonda (approved):
Sounds reasonable that if it's not needed, don't need to use it 👍

PR #160, opened 2022-07-06
Add support for setting timeouts to the RewindableBufferStream allowing halibut to recover from dropped tcp connections.

LukeButters:
# Background

The Halibut client maintains a pool of TCP connections which do not have keepalive enabled, sometimes these TCP connections can be killed. Ordinary Halibut handles this by checking the TCP connection with lowered TCP timeouts. Since the `RewindableBufferStream` did not support lowering the timeouts. So Halibut would wait 10 minutes which would exceed the `ConnectionErrorRetryTimeout` resulting in Halibut not creating a new TCP connection to execute the request with.

# Results

`RewindableBufferStream` now supports lowering the timeout.

## Before

```
We got response Hello
The time is: 6/7/2022 2:37:34 am
sleeping for: 271 seconds, goodnight
Woke up at: 6/7/2022 2:42:05 am
Unhandled exception. Halibut.HalibutClientException: An error occurred when sending a request to 'https://20.5.209.198:80/', before the request could begin: Unable to read data from the transport connection: Connection timed out.
 ---> Halibut.Transport.Protocol.ConnectionInitializationFailedException: Unable to read data from the transport connection: Connection timed out.
 ---> System.IO.IOException: Unable to read data from the transport connection: Connection timed out.
 ---> System.Net.Sockets.SocketException (110): Connection timed out
   --- End of inner exception stack trace ---
   at System.Net.Sockets.NetworkStream.Read(Span`1 buffer)
   at System.Net.Security.SslStream.ReadAsyncInternal[TIOAdapter](TIOAdapter adapter, Memory`1 buffer)
   at System.Net.Security.SslStream.Read(Byte[] buffer, Int32 offset, Int32 count)
   at Halibut.Transport.RewindableBufferStream.Read(Byte[] buffer, Int32 offset, Int32 count)
   at System.IO.StreamReader.ReadBuffer()
   at System.IO.StreamReader.ReadLine()
   at Halibut.Transport.Protocol.MessageExchangeStream.ReadLine()
   at Halibut.Transport.Protocol.MessageExchangeStream.ExpectProceeed()
   at Halibut.Transport.Protocol.MessageExchangeProtocol.PrepareExchangeAsClient()
   --- End of inner exception stack trace ---
   at Halibut.Transport.Protocol.MessageExchangeProtocol.PrepareExchangeAsClient()
   at Halibut.Transport.Protocol.MessageExchangeProtocol.ExchangeAsClient(RequestMessage request)
[...]

PR #161, opened 2022-07-06
Adds a test showing we can recover from tcp connections that die in the pool

LukeButters:
# Background

This adds a test for the fix in https://github.com/OctopusDeploy/Halibut/pull/160 showing that we can recover from TCP connections that become unresponsive in the pool.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
     "sdk": {
-        "version": "6.0.300",
+        "version": "5.0.100",
         "rollForward": "latestFeature"
     }
 }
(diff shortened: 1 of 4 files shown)


N-lson (approved):
The test itself makes sense to me

skermajo (commented):
The concern here is the classes added to test this; it's a fair bit of extra maintenance on untested code. Needing tests for the tests is never great and reduces confidence, no matter how well they're written or simple they may appear.

LukeButters:
That is a valid risk, hopefully the issues the test prevents saves us more time compared to the test maintenance. At worst we can always back this out if the test is unreliable.

PR #172, opened 2022-09-06
Pass the ServiceEndPoint for the RequestMessage to the ApplyResponse method of the PendingRequestQueue

nathanwoctopusdeploy:
# Background

When using Halibut in Server as part of the Halibut SQL Queue Item we need to know the timeouts for a response message in a custom implementation of IPendingRequestQueue.

# Results

This PR exposes the ServiceEndPoint when handling a response in the IPendingRequestQueue

The PR introduces a breaking change in the public API so we plan to bump the semver major version number.

The breaking changes are isolated to server or tentacle only when using the new version so do not impact compatibility between server and tentacle running different versions of Halibut

Quality :heavy_check_mark:

- Tested with server running Halibut with these changes and tentacle (workers and deployment targets) running Halibut before the changes.

# Pre-requisites


nathanwoctopusdeploy:
[sc-9641]

tothegills (approved):
Looks ok to me. Thanks for checking backwards compatibility.

PR #198, opened 2022-12-08
Fix timeouts seen with net6

LukeButters:
# Background

Fixes a bug in which halibut under NET6 would throw: `System.IO.InvalidDataException: The archive entry was compressed using an unsupported compression` exceptions or the client would just get stuck and no more messages would be processed.

# Results

Fixes: https://github.com/OctopusDeploy/Halibut/issues/199

The issue was related to timings and wrapped streams with buffers.

The message coming down the wire looks something like:
```
NEXT\r\n
<Zipped BSON><datastreams>
```

halibut was using two streams to read off the wire:
* The **inner stream**: which is used to read the zip.
* The **StreamReader**: which wraps the inner stream to read the control messages such as `NEXT\r\n`.

The issue occurs when both the control message `NEXT\r\n`  and the `<Zipped BSON>` arrived at the same time. What happens is the `StreamReader` which has its own buffer of `1024` bytes would read not only the `NEXT\r\n` message from the inner stream it would read some of the following `<Zipped BSON>` or perhaps even read the entire `<Zipped BSON><datastreams>` section. This is how we get the `System.IO.InvalidDataException` or the timeouts.

The fix is to, remove the `StreamReader` and read directly from the stream. Reading of short control messages is now down byte by byte which is slower, by an unmeasured and likely non noticeable amount.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Transport/Protocol/ControlMessageReader.cs
+++ b/source/Halibut/Transport/Protocol/ControlMessageReader.cs
@@ -5,7 +5,7 @@ using System.Threading.Tasks;
 
 namespace Halibut.Transport.Protocol
 {
-    public class ControlMessagesReader
+    public class ControlMessageReader
     {
         internal string ReadUntilNonEmptyControlMessage(Stream stream)
         {
@@ -21,26 +21,28 @@ namespace Halibut.Transport.Protocol
             StringBuilder sb = new StringBuilder();
             while (true)
             {
-                var line = stream.ReadByte();
-                if (line == -1) throw new EndOfStreamException();
-                if (line == '\r')
+                var nextByte = stream.ReadByte();
+                if (nextByte == -1) throw new EndOfStreamException();
+                // Control messages must end with \r\n and must not contain \r or \n
+                if (nextByte == '\r')
                 {
-                    line = stream.ReadByte();
-                    if (line == -1)
+                    nextByte = stream.ReadByte();
+                    if (nextByte == -1)
                     {
                         throw new EndOfStreamException();
                     }
 
-                    if (line != '\n')
+                    if (nextByte != '\n')
                     {
-                        throw new Exception($"It is not clear what should be done with this character: dec value {line}");
+                        var byteAsHex = nextByte.ToString("X");
+                        throw new Exception($"Unexpected byte after '\r' in control message: 0x{byteAsHex}");
                     }
 
-                    // We have found the end of the line, ie we have found \r\n
+                    // We have found the end of control message
                     return sb.ToString();
                 }
 
-                sb.Append((char) line);
+                sb.Append((char) nextByte);
             }
         }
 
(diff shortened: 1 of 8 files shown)


source/Halibut/Transport/Protocol/MessageExchangeStream.cs
```diff
@@ -168,8 +147,17 @@ public void IdentifyAsServer()
 
         public RemoteIdentity ReadRemoteIdentity()
         {
-            var line = streamReader.ReadLine();
+            var line = controlMessagesReader.ReadControlMessage(stream);
+            
+            
             if (string.IsNullOrEmpty(line)) throw new ProtocolException("Unable to receive the remote identity; the identity line was empty.");
+            
+            var emptyLine = controlMessagesReader.ReadControlMessage(stream);
```
LukeButters, line 155: Explicitly read the empty line that is sent after the identity line, rather than rely on the `StreamReader` to happen to consume that within its buffer.

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
```diff
@@ -186,7 +174,7 @@ public RemoteIdentity ReadRemoteIdentity()
             {
                 log.Write(EventType.Error, "Response:");
                 log.Write(EventType.Error, line);
-                log.Write(EventType.Error, streamReader.ReadToEnd());
+                log.Write(EventType.Error, new StreamReader(stream, new UTF8Encoding(false)).ReadToEnd());
```
LukeButters, line 177: I wonder if a short timeout should be set here, in this error case we don't know how much has been sent over the wire. Potentially we could block for a long time here. In a sense this is now more likely since we consume the `emptyLine` that is sent after the identity line, so in the case the identity line is wrong we could hang.

source/Halibut/Transport/Protocol/ControlMessageReader.cs
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        internal string ReadControlMessage(Stream stream)
+        {
+            StringBuilder sb = new StringBuilder();
+            while (true)
+            {
+                var nextByte = stream.ReadByte();
+                if (nextByte == -1) throw new EndOfStreamException();
+                // Control messages must end with \r\n and must not contain \r or \n
```
veochen-octopus, line 26: I'm curious how is this ensured from the sender's side? `StreamWriter` seems to use `Environment.NewLine` as the line break.

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
veochen-octopus: My understanding is that `ReadToEnd` on a network stream will block forever until the connection's closed or the read timeout is reached.

source/Halibut/Transport/Protocol/ControlMessageReader.cs
LukeButters: It looks like `\r\n` is always sent as part of https://github.com/OctopusDeploy/Halibut/blob/9fce5a8b3de869a3af7acaa0547172432326cae1/source/Halibut/Transport/Protocol/MessageExchangeStream.cs#L37

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
LukeButters: I think that is true, so I guess what it does now is ok.

veochen-octopus (approved):
LGTM

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
veochen-octopus: It's not any different than before. I think we can put this in and if need be deal with any blocking issues that arise.

PR #200, opened 2022-12-11
Adds a test

LukeButters:
# Background

Adds a test for the timeout and errors seen in https://github.com/OctopusDeploy/Halibut/pull/198

# Results

The port forwarder is re-used to add latency and change how data arrives. The new behaviour allows configuring the port forwarder to buffer up data before sending until no data has been received in some number of milliseconds. This means if a sender sends the following:
```
write("foo");
flush();
write("bar");
flush();
```

Ordinarily the receiver would read `"foo"` from the stream and later read `"bar"`.  With the delay and buffering added in this PR, the sender will end up sending both messages in a single write so the above turns in to:

```
write("foobar")
```

This likely ends up with the reader reading `"foobar"` off the wire in a single read call.

This is how we reproduce problems in which halibut consumes too much of the wire when the data is available.

## Before

We did not have a test that reproduced the problem seen in https://github.com/OctopusDeploy/Halibut/pull/198

## After

Now we have a test that reproduces the problem seen in https://github.com/OctopusDeploy/Halibut/pull/198

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/UsageFixture.cs
+++ b/source/Halibut.Tests/UsageFixture.cs
@@ -10,6 +10,7 @@ using System.Threading.Tasks;
 using FluentAssertions;
 using Halibut.ServiceModel;
 using Halibut.Tests.TestServices;
+using Halibut.Tests.Util;
 using NUnit.Framework;
 
 namespace Halibut.Tests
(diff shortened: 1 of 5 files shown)


veochen-octopus (approved):
Pretty magical stuff. Nice one figuring this out. Might be an obvious one but I also assume the test fails without the fix in place?

PR #201, opened 2023-01-05
Sast/provide async read write methods for rewindable buffer stream

matthew-james-octopus:
# Background

While investigating an issue related to threadpool starvation, we noticed a stack trace that indicated that the `RewindablebufferStream.Read`, a sync operation, becomes involved in an async pathway, since `SslStream`'s implementation of `Read` ultimately involves async behaviour.

image

This PR overrides the `Stream.ReadAsync` and `Stream.WriteAsync` methods in `RewindableBufferStream` so that the synchronous default implementations of `Stream` are not relied on. It is hoped that specifically this will avoid thread exhaustion due to async-over-sync scenarios appearing, but is seen in any case to be in general a Good Thing by providing async overrides.

[sc-29089]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Transport/RewindableBufferStreamTests.cs
+++ b/source/Halibut.Tests/Transport/RewindableBufferStreamTests.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.IO;
 using System.Text;
+using System.Threading.Tasks;
 using Halibut.Transport;
 using NUnit.Framework;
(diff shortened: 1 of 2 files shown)


LukeButters (approved):

veochen-octopus (approved):
Makes sense to me 👍

PR #209, opened 2023-02-21
Pass the cancellation token to the inner stream when reading

LukeButters:
# Background

Otherwise the read will hang forever, since under async the ReadTimeout is not respected.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Transport/RewindableBufferStreamTests.cs
+++ b/source/Halibut.Tests/Transport/RewindableBufferStreamTests.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.IO;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 using Halibut.Transport;
 using NUnit.Framework;
@@ -42,6 +43,25 @@ namespace Halibut.Tests.Transport
                 {
                     Assert.AreEqual("Test", await streamReader.ReadToEndAsync());
                 }
+            }
+        }
+        
+        [Test]
+        public async Task ReadAsyncShouldPassTheCancellationTokenToTheUnderlyingStream()
+        {
+            using (var baseStream = new MemoryStream(16))
+            using (var sut = RewindableBufferStreamBuilder.Build(baseStream))
+            {
+                var inputBuffer = Encoding.ASCII.GetBytes("Test");
+                await baseStream.WriteAsync(inputBuffer, 0, inputBuffer.Length);
+
+                baseStream.Position = 0;
+
+                var readBuffer = new byte[inputBuffer.Length];
+                var cts = new CancellationTokenSource();
+                cts.Cancel();
+                
+                Assert.ThrowsAsync<TaskCanceledException>(async () => await sut.ReadAsync(readBuffer, 0, readBuffer.Length, cts.Token));
             }
         }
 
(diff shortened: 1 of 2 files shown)


source/Halibut/Transport/RewindableBufferStream.cs
```diff
@@ -114,7 +114,7 @@ public async override Task<int> ReadAsync(byte[] buffer, int offset, int count,
                 return rewoundCount;
             }
 
-            var baseCount = await baseStream.ReadAsync(buffer, offset, count);
+            var baseCount = await baseStream.ReadAsync(buffer, offset, count, cancellationToken);
```
geofflamrock, line 117: Are there any tests that currently exist or we could potentially add that would help us to assert the expected behaviour around cancellation? E.g. Start a stream, cancel the read via the token and make sure it doesn't hang?

LukeButters: I think only one test exists in halibut that checks timeouts work but that goes down a sync path. I can take a closer look tomorrow. It is not clear to me if this async read method is ever called currently.

geofflamrock: > It is not clear to me if this async read method is ever called currently.

If we aren't sure if it's called, then are we sure this change will fix the underlying issue? Feels like it would be good to get some confidence in the change through some tests.

LukeButters: Testing the read timeout works when the client waits for the service to respond is difficult to write currently, since the read timeout comes from the static
```
public static TimeSpan TcpClientReceiveTimeout = TimeSpan.FromMinutes(10);
```
and so the test would take ten minutes to run.

I guess that could be changed to take the timeout from the ServiceEndpoint.

I guess a simpler test is one that targets the RewindableBufferStream but that doesn't really capture the behaviour and doesn't show the timeout working when waiting for the service to respond.

In any case passing the cancellation token along makes sense.

LukeButters: @geofflamrock I created a targeted test for this.

source/Halibut.Tests/Transport/RewindableBufferStreamTests.cs
```diff
@@ -44,6 +45,25 @@ public async Task ReadAsyncShouldPassThrough()
...
+                var inputBuffer = Encoding.ASCII.GetBytes("Test");
+                await baseStream.WriteAsync(inputBuffer, 0, inputBuffer.Length);
+
+                baseStream.Position = 0;
+
+                var readBuffer = new byte[inputBuffer.Length];
+                var cts = new CancellationTokenSource();
+                cts.Cancel();
+                
+                Assert.ThrowsAsync<TaskCanceledException>(async () => await sut.ReadAsync(readBuffer, 0, readBuffer.Length, cts.Token));
```
LukeButters, line 64: Amazingly this is not something that is awaited!

geofflamrock: Looks like we are already using FluentAssertions in the tests so you could use `Should().ThrowAsync<TaskCanceledException>` if you wanted to.

geofflamrock (approved):
LGTM, I've left one comment on the test for you to consider

source/Halibut.Tests/Transport/RewindableBufferStreamTests.cs
LukeButters: oh let me try that.

LukeButters: oh nice that worked.

PR #210, opened 2023-02-22
Correct the doco around the listening protocol.

LukeButters:
# Background

Fixes the doco around listening mode.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


docs/halibut-protocol.md
```diff
@@ -29,13 +30,42 @@ The final stage can be entered by either the Client or Service terminating the T
...
+    3. If the client recieves `PROCEED` the RPC call is made with this TCP connection starting at step `2.Message exchange`.
+
+
 
 ![Listening client protocol data exchange](images/listeningprotocoldata.png)
 
 ## Polling Service protocol data exchange
 
-When the **Service** is in  **Polling** mode, the **Service** identifies itself as `MX-SUBSCRIBER 1.0` while the `Client` identifies itself as `MX-SERVER 1.0`.
+To make RPC calls the steps are:
```
geofflamrock, line 56: It might be worth saying here that in polling mode a single connection is opened by the client as opposed to multiple as required, and incorporate the bit further down the bit about keeping it alive. It's noted in the connection re-use section but feels like it worth pulling up to the top of the section?

docs/halibut-protocol.md
```diff
@@ -29,13 +30,42 @@ The final stage can be entered by either the Client or Service terminating the T
...
+1. *Identification:*
+    1. **Service** identifies itself as `MX-SUBSCRIBER 1.0`
+    2. **Client** identifies itself as `MX-SERVER 1.0`
+2. *Message exchange:*
+    1. The Client sends a `Request` message containing the method to execute, its parameters and 0 or more data streams.
+    2. The Service executes the method and sends the result in a `Response` message containing the result and 0 or more data streams.
+3. *Connection re-use:*
+    1. The **Service** sends a `NEXT` control message.
+    2. The **Client** responds with a `PROCEED` control message.
+    3. The **Service** Waits for a **Request** message.
```
geofflamrock, line 67: Might be worth mentioning here that there is a timeout on this waiting and if there is no request we go around again?

geofflamrock (commented):
Looks good 👍  just a couple of minor thoughts

docs/halibut-protocol.md
LukeButters: I added a section above around the Service making the connection and that only a single connection exists at a time.

I left the No-op bit ad the end since I think it makes sense to introduce that after the process is described.

docs/halibut-protocol.md
LukeButters: Added a note.

docs/halibut-protocol.md
```diff
@@ -29,15 +30,46 @@ The final stage can be entered by either the Client or Service terminating the T
...
+1. *Identification:*
+    1. **Service** identifies itself as `MX-SUBSCRIBER 1.0`
+    2. **Client** identifies itself as `MX-SERVER 1.0`
+2. *Message exchange:*
+    1. The Client sends a `Request` message containing the method to execute, its parameters and 0 or more data streams.
+    2. The Service executes the method and sends the result in a `Response` message containing the result and 0 or more data streams.
+3. *Connection re-use:*
+    1. The **Service** sends a `NEXT` control message.
+    2. The **Client** responds with a `PROCEED` control message.
+    3. The **Service** Waits for a **Request** message. If waiting times out the connection is closed and recreated starting again from step `1.Identification`.
```
geofflamrock, line 69: Is this correct? I thought if there was no message and waiting timed out then the Client sent a null request and the process went around again, not that the connection gets recreated?

LukeButters: The bit at the end about No-ops being sent I think covers that case.

geofflamrock: Ah right, I'm still finding it a bit hard to follow as this makes it sound like the normal case will be that the connection drops and gets created. It might be worth adding both the cases here as sub-points so that the normal case of the connection being held open is clear.

nathanwoctopusdeploy (approved):
LGTM

docs/halibut-protocol.md
LukeButters: I agree I think this could be improved, but I think I will merge this as is since it is better and we can come back to it when we have time and focus.

PR #218, opened 2023-03-14
Add a test around killing a polling TCP connection mid request

LukeButters:
# Background

Adds a test showing what happens when a TCP connection to a polling tentacle is killed while waiting for the Response to come back.

The test shows that the killed connection is immediately detected and a error response is immediately put back on to the queue.

This also does some clean up around the port forwarder which I previously had done for some other investigation. The changes are:
* Support disposing of the port forwarder twice required for this change.
* Make the port forwarder respect cancellation such that it will stop pumping bytes as soon as it is disposed.
* Disposing all the things rather than stop disposing at the first exception (may not be required for this test).

[SC-43094]
# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Util/SocketPump.cs
+++ b/source/Halibut.Tests/Util/SocketPump.cs
@@ -30,7 +30,7 @@ namespace Halibut.Tests.Util
             // Only read if we have nothing to send or if data exists 
             if (readFrom.Available > 0 || buffer.Length == 0)
             {
-                var receivedByteCount = await ReadFromSocket(readFrom, buffer);
+                var receivedByteCount = await ReadFromSocket(readFrom, buffer, cancellationToken);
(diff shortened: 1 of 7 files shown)


source/Halibut.Tests/Util/SocketPump.cs
```diff
@@ -54,29 +54,40 @@ public async Task<SocketStatus> PumpBytes(Socket readFrom, Socket writeTo, Cance
 
         async Task PausePump(CancellationToken cancellationToken)
         {
+            cancellationToken.ThrowIfCancellationRequested();
             while (isPumpPaused())
             {
                 await Task.Delay(TimeSpan.FromMilliseconds(100));
```
gb-8, line 60: You could pass the cancellation token in here too.

PR #221, opened 2023-04-11
Reset the reconnection backoff strategy for a Polling Client once a connection is successfully established to the Service

nathanwoctopusdeploy:
# Background

When a Polling Client connects to a Service, it uses a backoff strategy to determine if it should wait between connection attempts.

The intent was to slow re-connection attempts to the Service when the Polling Client was having trouble connecting rather than trying to connect very frequently and possibly making the issue worse.

The backoff is based on the duration that the client has been trying to connect and ranges from 5 seconds to a maximum of 2 minutes.

A side effect of this logic means that if a polling client has been connected to the Service successfully for 2 minutes or over and then experiences a connection issue the client will wait for 2 minutes before trying to reconnect. Given the connection has been stable, possibly for a long duration, the 2-minute backoff does not seem necessary and causes unexpected delays in the polling tentacle reconnecting and picking up queued requests.

## Backoff strategy 

```
Elapsed 00:00:00 - Delay 00:00:05
Elapsed 00:00:01 - Delay 00:00:05
Elapsed 00:00:05 - Delay 00:00:08.7314902
Elapsed 00:00:10 - Delay 00:00:17.4629805
Elapsed 00:00:20 - Delay 00:00:34.9259611
Elapsed 00:00:30 - Delay 00:00:52.3889417
Elapsed 00:01:00 - Delay 00:01:44.7778834
Elapsed 00:02:00 - Delay 00:02:00
Elapsed 00:03:00 - Delay 00:02:00
```

# Results

This PR resets the backoff strategy once a connection has been successfully established. If the connection is dropped and the client needs to reconnect, the backoff strategy will start again, e.g. it will wait 5 seconds initially rather than 2 minutes.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/PollingClientConnectionHandlingFixture.cs
+++ b/source/Halibut.Tests/PollingClientConnectionHandlingFixture.cs
@@ -1,0 +1,161 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using FluentAssertions;
+using Halibut.ServiceModel;
+using Halibut.Tests.TestServices;
+using Halibut.Tests.Util;
+using NUnit.Framework;
+
+namespace Halibut.Tests
+{
+    public class PollingClientConnectionHandlingFixture
+    {
+        [Test]
+        public void PollingClientShouldConnectQuickly()
+        {
+            var started = DateTime.UtcNow;
+            var calls = new List<DateTime>();
+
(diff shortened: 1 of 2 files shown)


LukeButters (approved):
This is going to make development a lot easier.

nathanwoctopusdeploy:

PR #223, opened 2023-04-13
Bump nunit test adapter

nathanwoctopusdeploy:
Bump nunit test adapter to fix https://github.com/nunit/nunit3-vs-adapter/issues/987


LukeButters (approved):
Yep, that version is higher :p

PR #226, opened 2023-04-19
Luke/throw specific method not found exceptions

LukeButters:
# Background

This PR makes it easier for a client to determine if the service does not implement the called service or method. This is useful when newer clients are talking to older services which don't have the latest methods.

[SC-46268]
# Results

Clients can now catch a `NoMatchingServiceOrMethodHalibutClientException` to discover that the service (Tentacle):
- Doesn't know about the Service.
- Doesn't know about the method within that Service.
- Can't find an unambiguously matching method.

Also it is possible for a client to distinguish between those with:
- `ServiceNotFoundHalibutClientException`
- `MethodNotFoundHalibutClientException`
- `AmbiguousMethodMatchHalibutClientException`

The solution is also backwards compatible, this means an upgraded client can catch those exceptions and `Halibut` (client side) will do the heavy lifting of converting the old responses into the correct exception.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/ServiceModel/HalibutProxy.cs
+++ b/source/Halibut/ServiceModel/HalibutProxy.cs
@@ -166,6 +166,17 @@ namespace Halibut.ServiceModel
                 return;
 
             var realException = responseMessage.Error.Details as string;
+            if (!string.IsNullOrEmpty(responseMessage.Error.ErrorType))
+            {
+                var theType = Type.GetType(responseMessage.Error.ErrorType);
+                if (theType != null)
+                {
+                    var ctor = theType.GetConstructor(new[] { typeof(string), typeof(string) });
+                    Exception e = (Exception)ctor.Invoke(new object[] { responseMessage.Error.Message, realException });
+                    throw e;
+                }
+            }
+            
             throw new HalibutClientException(responseMessage.Error.Message, realException);
         }
     }
--- a/source/Halibut/ServiceModel/HalibutProxy.cs
+++ b/source/Halibut/ServiceModel/HalibutProxy.cs
@@ -77,19 +77,6 @@ namespace Halibut.ServiceModel
         {
             return messageRouter(requestMessage, cancellationToken);
         }
-
-        static void EnsureNotError(ResponseMessage responseMessage)
-        {
-            if (responseMessage == null)
-                throw new HalibutClientException("No response was received from the endpoint within the allowed time.");
-
-            if (responseMessage.Error == null)
-                return;
-
-            var realException = responseMessage.Error.Details as string;
-            throw new HalibutClientException(responseMessage.Error.Message, realException);
-        }
-    }
 #else
     public class HalibutProxy : DispatchProxy
     {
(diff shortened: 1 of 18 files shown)


source/Halibut/ServiceModel/HalibutProxy.cs
```diff
@@ -165,9 +153,39 @@ static void EnsureNotError(ResponseMessage responseMessage)
...
-            throw new HalibutClientException(responseMessage.Error.Message, realException);
+            ThrowExceptionFromReceivedError(responseMessage.Error);
+        }
+        
+        internal static void ThrowExceptionFromReceivedError(ServerError error)
+        {
+            var realException = error.Details as string;
+            if (!string.IsNullOrEmpty(error.HalibutErrorType))
+            {
+                var theType = Type.GetType(error.HalibutErrorType);
```
nathanwoctopusdeploy, line 164: Should we wrap this in a try-catch so we guarantee we don't break things unexpectedly and instead fallback to the previous behaviour?

source/Halibut/Transport/Protocol/ResponseMessage.cs
```diff
@@ -27,7 +27,17 @@ public static ResponseMessage FromError(RequestMessage request, string message)
 
         public static ResponseMessage FromException(RequestMessage request, Exception ex)
         {
-            return new ResponseMessage {Id = request.Id, Error = new ServerError { Message = ex.UnpackFromContainers().Message, Details = ex.ToString() }};
+            return new ResponseMessage {Id = request.Id, Error = ServerErrorFromException(ex)};
+        }
+
+        internal static ServerError ServerErrorFromException(Exception ex)
```
nathanwoctopusdeploy, line 33: Are we missing a test to guarantee that we are not breaking the current behaviour of this method, e.g. Old Halibut Client is compatible with new Halibut Service?

LukeButters: Is that the case that a new Halibut service throws an exception and so ends up here and so adds the ErrorType to the ServerError. Then Old halibut receives a ServerError which contains an additional parameter.

What would that test be? I guess the test that might be possible to write is one that verifies the message serializer can successfully ignore that extra parameter. I will add this test.

The other test would be an integration test between new Halibut and old Halibut. I am still unsure how to piratically do that.

nathanwoctopusdeploy: > The other test would be an integration test between new Halibut and old Halibut. I am still unsure how to piratically do that.

Yes this would be the nicest but we need to figure that out. Maybe skip for now but something we can come back to later

LukeButters: Added a test which shows that extra fields in ServerError are ignored by the serializer. This should at least show that for older halibut clients they will just ignore the extra field and just throw a HalibutClientException as they normaly would.

source/Halibut/ServiceModel/HalibutProxy.cs
LukeButters: If a non HalibutClientException is thrown, that error is now caught and logged. Finally the old behaviour is run of just throwing a HalibutClientException.

nathanwoctopusdeploy (approved):
LGTM 👍

nathanwoctopusdeploy:

PR #230, opened 2023-04-27
Improves the envelope was null error and invalid compression format error to instead report the actual underlying network error.

LukeButters:
# Background

Previously network errors would result in a message envelope was null error or the compression format was invalid. Both of which is wrong. This was happening because end of stream within reading messages was not considered an error in either zip or JSON stream reading.

The change here is to treat timeouts and end of streams when reading messages as an error and throw that type of error.

It works by wrapping the stream before passing that stream to the json serializer. The wrapper `ErrorRecordingStream` records if a error has occured or if the end of the stream was reached. If either of those have occured then we just need to throw some sort of IOException.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Transport/Protocol/MessageSerializerFixture.cs
+++ b/source/Halibut.Tests/Transport/Protocol/MessageSerializerFixture.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.IO;
 using System.IO.Compression;
-using System.Linq;
 using System.Text;
 using FluentAssertions;
 using Halibut.Tests.Support;
--- a/source/Halibut.Tests/Transport/Protocol/MessageSerializerFixture.cs
+++ b/source/Halibut.Tests/Transport/Protocol/MessageSerializerFixture.cs
@@ -86,7 +86,7 @@ namespace Halibut.Tests.Transport.Protocol
         }
 
         [Test]
-        public void WhenTheStreamEndsBeforeBeforeAnyBytesAreRead_AEndOfStreamExceptionIsThrown()
+        public void WhenTheStreamEndsBeforeAnyBytesAreRead_AnEndOfStreamExceptionIsThrown()
         {
             var sut = new MessageSerializerBuilder().Build();
             using (var stream = new MemoryStream(new byte[0]))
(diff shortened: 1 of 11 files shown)


source/Halibut/Transport/Protocol/MessageSerializer.cs
```diff
@@ -131,7 +170,7 @@ MessageEnvelope<T> DeserializeMessage<T>(JsonReader reader)
         
         // By making this a generic type, each message specifies the exact type it sends/expects
         // And it is impossible to deserialize the wrong type - any mismatched type will refuse to deserialize
-        class MessageEnvelope<T>
+        public class MessageEnvelope<T>
```
nathanwoctopusdeploy, line 173: Does this need to be public?

source/Halibut.Tests/Util/ArrayExtensionMethods.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+
+namespace Halibut.Tests.Util
+{
+    public static class ArrayExtensionMethods
+    {
+        public static T[] SubArray<T>(this T[] array, int offset, int length)
```
nathanwoctopusdeploy, line 7: .Skip(a).Take(b)?

source/Halibut.Tests/Transport/Protocol/MessageSerializerFixture.cs
```diff
@@ -83,7 +85,81 @@ public void BackwardsCompatibility_ExtraParametersInServerErrorAreIgnored()
             }
         }
 
-        #if !NETFRAMEWORK
+        [Test]
+        public void WhenTheStreamEndsBeforeBeforeAnyBytesAreRead_AEndOfStreamExceptionIsThrown()
```
nathanwoctopusdeploy, line 89: nit: BeforeBefore _A`n`End

nathanwoctopusdeploy (approved):
Nice

source/Halibut/Transport/Protocol/MessageSerializer.cs
LukeButters: no reverting

PR #231, opened 2023-04-28
Adds the ability to detect if an error is a network error

LukeButters:
# Background

It can be useful for clients to be able to know if the error returned by the Halibut client is a network error or not. This can be useful to know when to try again.

[SC-46263]

[SC-46263](https://app.shortcut.com/octopusdeploy/story/46263/halibut-ability-to-identify-halibut-network-related-errors)

# Results

An extension method which can tell you if an exception thrown by the halibut client is or is not a network error. Since it is currently it also has to return the `Unknown` type for many errors since it is not easy or possible to determine what is or is not a network error. This isn't great but at least clients can just depend on the halibut implementation, which can be improved as needed.

This adds specific cases around:
* proxies
* FileNotFound exceptions.

This adds a new halibut client exception type of `ServiceInvocationHalibutClientException` thrown when the service e.g. tentacle throws an exception.

This also adds a new TCP Listening helper which just listens on a port and each time a client connects it immediately closes that new tcp connection.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Diagnostics/ExceptionReturnedByHalibutProxyExtensionMethodFixture.cs
+++ b/source/Halibut.Tests/Diagnostics/ExceptionReturnedByHalibutProxyExtensionMethodFixture.cs
@@ -1,4 +1,5 @@
 using System;
+using System.IO;
 using System.Threading.Tasks;
 using FluentAssertions;
 using Halibut.Diagnostics;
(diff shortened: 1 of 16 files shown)


source/Halibut.Tests/Diagnostics/UnpackFromContainersFixture.cs
```diff
@@ -0,0 +1,48 @@
+using System;
+using System.Reflection;
+using FluentAssertions;
+using Halibut.Diagnostics;
+using NUnit.Framework;
+
+namespace Halibut.Tests.Diagnostics;
+
+public class UnpackFromContainersFixture
```
LukeButters, line 9: I didn't understand what this was doing so I wrote a test. Which I have accidentally committed. Would you like this to come along for the ride? or should I drop this test?

source/Halibut.Tests/Diagnostics/ExceptionReturnedByHalibutProxyExtensionMethodFixture.cs
```diff
@@ -0,0 +1,211 @@
...
+                    var octopusPort = octopus.Listen();
+                    octopus.Trust(Certificates.TentaclePollingPublicThumbprint);
+
+                    tentaclePolling.Poll(new Uri("poll://SQ-TENTAPOLL"), new ServiceEndPoint(new Uri("https://localhost:" + octopusPort), Certificates.OctopusPublicThumbprint));
+
+                    var echo = octopus.CreateClient<IEchoService>("poll://SQ-TENTAPOLL", Certificates.TentaclePollingPublicThumbprint);
+
+                    var dataStream = new DataStream(10, stream =>
+                    {
+                        new FileStream("DoesNotExist2497546", FileMode.Open).Dispose();
```
LukeButters, line 175: The error message from FileStream opening a file which is what Server is doing looks different to the one thrown by just throwing a FileNotFoundException. This and the above tests exists because we want to be sure we get this one correct.

source/Halibut.Tests/Util/TCPListenerWhichKillsNewConnections.cs
```diff
@@ -0,0 +1,56 @@
+using System;
+using System.Net;
+using System.Net.Sockets;
+using System.Threading;
+using System.Threading.Tasks;
+using Serilog;
+
+namespace Halibut.Tests.Util
+{
+    public class TCPListenerWhichKillsNewConnections : IDisposable
```
LukeButters, line 15: I added this since some tests want to start a TCP connection to something that is in someway immediately not working. If the tests was to talk to a random port then it could be that it by chance talks to a port that is actually listening making the test sometimes fail.

This causes the test to be predictable.

source/Halibut/Diagnostics/ExceptionReturnedByHalibutProxyExtensionMethod.cs
```diff
@@ -0,0 +1,79 @@
...
+                return HalibutNetworkExceptionType.NotANetworkError;
+            }
+
+            if (IsErrorInService(exception))
+            {
+                return HalibutNetworkExceptionType.NotANetworkError;
+            }
+
+            if (exception is UnexpectedCertificateException
+                || exception is FileNotFoundException)
```
LukeButters, line 27: This FileNotFoundException is a bit of a mess, because the DataStream itself has a "Action" that writes to the given stream (ie the network) we don't know if the error in a DataStream because writing to the given stream (ie the network) caused an error OR if something else caused an error in a DataStream.

At least right now a `FileNotFoundException` is always going to be some sort of non network error so this works for the errors we have seen in server.

source/Halibut.Tests/Diagnostics/ExceptionReturnedByHalibutProxyExtensionMethodFixture.cs
```diff
@@ -0,0 +1,211 @@
...
+using Halibut.Diagnostics;
+using Halibut.ServiceModel;
+using Halibut.Tests.TestServices;
+using Halibut.Tests.Util;
+using Halibut.Transport.Proxy;
+using NUnit.Framework;
+
+namespace Halibut.Tests.Diagnostics
+{
+    public static class ExceptionReturnedByHalibutProxyExtensionMethodFixture
```
LukeButters, line 14: I don't think it is possible to just throw a "ThisWasANetworkError" in a consistent way, so I have added lots of tests to try to verify the extension method actually works.

source/Halibut/Diagnostics/ExceptionReturnedByHalibutProxyExtensionMethod.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            return HalibutNetworkExceptionType.UnknownError;
+        }
+
+        /// <summary>
+        /// Did the exception thrown from a HalibutProxy object one which was caused by the service e.g. Tentacle
+        /// throwing an exception?
+        /// </summary>
+        /// <param name="e"></param>
+        /// <returns></returns>
```
nathanwoctopusdeploy, line 67: Nit: incomplete doco. Also, does this need to be public?

LukeButters: I think it would be handy to be able to identify tentacle errors, currently we do NOT log tentacle errors since we think they could contain sensitive information.

nathanwoctopusdeploy (changes requested):
Looking good. I like the integration tests that check some specific errors. Are we missing unit tests for the extension to do variation testing e.g. `NoMatchingServiceOrMethodHalibutClientException` should not be a Network error, inner exception variations etc.

LukeButters:
@nathanwoctopusdeploy Added tests around the network not found as unit tests, which should be fine as unit tests since those exceptions are properly thrown.

Also added a test for the service threw an exception case testing both new methods.

source/Halibut.Tests/Diagnostics/ExceptionReturnedByHalibutProxyExtensionMethodFixture.cs
```diff
@@ -0,0 +1,259 @@
...
+    {
+        public class WhenGivenA
```
LukeButters, line 17: No tests around inner exceptions here since the exception containing an inner exception is "just the way it happens to work" rather than explicit behaviour like throwing a `MethodNotFoundHalibutClientException` from a halibut client object.

nathanwoctopusdeploy (approved):
LGTM

nathanwoctopusdeploy:

PR #234, opened 2023-05-11
Add support for testing slightly older versions of Halibut

LukeButters:
# Background

Adds support for testing the current version of Halibut (ie what you are editing) with a previous version of Halibut. Allowing developers to write backwards compatibility tests between latest client and a previous version of Halibut used as a service.

[SC-49433]

## It is now possible to write a backwards compatibility test for example:

```
using (var clientAndService = await ClientAndPreviousVersionServiceBuilder.WithListeningService().WithServiceVersion("5.0.429").Build())
{
    var echo = clientAndService.CreateClient<IEchoService>(se =>
    {
        se.PollingRequestQueueTimeout = TimeSpan.FromSeconds(20);
        se.PollingRequestMaximumMessageProcessingTimeout = TimeSpan.FromSeconds(20);
    });

    var ex = Assert.Throws<ServiceInvocationHalibutClientException>(() => echo.Crash());
}
```

## To add a new version of Halibut to test against.

Halibut is a library, so we need something to actually test against. This is solved by creating a binary which has a implementation of the `IEchoService` which can be called from the tests. Doing so means that to test a previously untested version of Halibut requires some simple boilerplate.

To start testing against version 13.3.7 the steps are:

- Create a new Console project named `Halibut.TestUtils.CompatBinary.v13_3_7` note the dots are replaced with underscore.
- Update that projects csproj file to be similar to `Halibut.TestUtils.CompatBinary.v5_0_429` except make sure to depend on the version of Halibut you want to test against.
- Update the Main to call `return BackwardsCompatProgramBase.Main(args)`
- Update the csproj of Halibut.Tests to depend on that project without depending on the assemblies. In this case:
```
<ProjectReference Include="..\Halibut.TestUtils.CompatBinary.v13_3_7\Halibut.TestUtils.CompatBinary.v13_3_7.csproj" ReferenceOutputAssembly="false" />
```
- Call your version of the binary from the test e.g. `await ClientAndPreviousVersionServiceBuilder.WithListeningService().WithServiceVersion("13.3.7").Build()` 

## Solution summary

- A `Halibut.TestUtils.CompatBinary.Base` project is created which contains:
    - A `IEchoService` and implementation. This is shared between the various console projects that depend on some particular version of Halibut. Doing this means only the version of Halibut changes but the service implementation stays the same making tests easier to write.
    - A class which knows how to start Halibut as a service (listening or polling) and configures the service to make available the `IEchoService`
    - To test against a particular version of Halibut, a project with format `Halibut.TestUtils.CompatBinary.v{VersionOfHalibut}` is created which simply depends on the above project and depends on the version of Halibut that a developer wishes to test against.
- The `Halibut.Test` project:
    - Has a dependency on the `Halibut.TestUtils.CompatBinary.v{VersionOfHalibut}` projects with `ReferenceOutputAssembly="false"` Doing so means that the binary is built but we don't include its dlls in the test project.
    - It has a `ClientAndOldServiceBuilder` which can be used to run that `CompatBinary` in the test.

This solution was chosen since it is simple. It might have been possible to remove the need for any boilerplate with the use of class loaders (app domain? what is this called in dotnet?), however that seems like an overly complex solution. Since it is unclear how we would fetch not just the halibut dll but also its dependency, nor is it clear how halibut types would cross or not cross that boundary. Doing that would make it easy to test many versions, but we don't think we will need to test against many versions of Halibut or even more than one.

## Future work

Add support for testing an old halibut client calling the latest halibut service.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Halibut.Tests/BackwardsCompatibility/InvocationExceptionsAreMappedCorrectly.cs
```diff
@@ -0,0 +1,42 @@
+using System;
+using System.Threading.Tasks;
+using Halibut.Exceptions;
+using Halibut.Tests.BackwardsCompatibility.Util;
+using Halibut.Tests.TestServices;
+using NUnit.Framework;
+
+namespace Halibut.Tests.BackwardsCompatibility
+{
+    public class InvocationExceptionsAreMappedCorrectly
```
LukeButters, line 10: This is a real test, without this we have a gap in our testing.

source/Halibut.TestUtils.CompatBinary.Base/EchoService.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            return name;
+        }
+
+        public bool Crash()
+        {
+            throw new Exception();
+        }
+
+        public static Action OnLongRunningOperation { get; set; }
```
nathanwoctopusdeploy, line 19: Not static

source/Halibut.Tests/BackwardsCompatibility/Util/ClientAndOldServiceBuilder.cs
```diff
@@ -0,0 +1,101 @@
...
+using Halibut.Tests.Util;
+
+namespace Halibut.Tests.BackwardsCompatibility.Util
+{
+    public class ClientAndOldServiceBuilder
+    {
+        readonly ServiceConnectionType serviceConnectionType;
+        readonly CertAndThumbprint serviceCertAndThumbprint;
+        readonly CertAndThumbprint clientCertAndThumbprint = CertAndThumbprint.Octopus;
+        string version = "5.0.429";
```
nathanwoctopusdeploy, line 13: Throw instead of default

source/Halibut.Tests/BackwardsCompatibility/Util/ClientAndOldServiceBuilder.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        ClientAndOldServiceBuilder(ServiceConnectionType serviceConnectionType, CertAndThumbprint serviceCertAndThumbprint)
+        {
+            this.serviceConnectionType = serviceConnectionType;
+            this.serviceCertAndThumbprint = serviceCertAndThumbprint;
+        }
+
+        public static ClientAndOldServiceBuilder Polling()
```
nathanwoctopusdeploy, line 21: WithPollingService

source/Halibut.Tests/BackwardsCompatibility/Util/ClientAndOldServiceBuilder.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            return new ClientAndOldServiceBuilder(ServiceConnectionType.Polling, CertAndThumbprint.TentaclePolling);
+        }
+
+        public static ClientAndOldServiceBuilder Listening()
```
nathanwoctopusdeploy, line 26: WithListeningService

source/Halibut.Tests/BackwardsCompatibility/Util/ClientAndOldServiceBuilder.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            return new ClientAndOldServiceBuilder(ServiceConnectionType.Listening, CertAndThumbprint.TentacleListening);
+        }
+
+        public ClientAndOldServiceBuilder WithVersion(string version)
```
nathanwoctopusdeploy, line 31: WithServiceVersion

source/Halibut.Tests/Util/CertAndThumbprint.cs
```diff
@@ -5,17 +5,19 @@ namespace Halibut.Tests.Util
...
     {
-        public static CertAndThumbprint TentacleListening = new(Certificates.TentacleListening, Certificates.TentacleListeningPublicThumbprint);
-        public static CertAndThumbprint TentaclePolling = new(Certificates.TentaclePolling, Certificates.TentaclePollingPublicThumbprint);
-        public static CertAndThumbprint Octopus = new(Certificates.Octopus, Certificates.OctopusPublicThumbprint);
+        public static CertAndThumbprint TentacleListening = new(Certificates.TentacleListeningPfxPath, Certificates.TentacleListening, Certificates.TentacleListeningPublicThumbprint);
+        public static CertAndThumbprint TentaclePolling = new(Certificates.TentaclePollingPfxPath, Certificates.TentaclePolling, Certificates.TentaclePollingPublicThumbprint);
+        public static CertAndThumbprint Octopus = new(Certificates.OctopusPfxPath, Certificates.Octopus, Certificates.OctopusPublicThumbprint);
 
-        public CertAndThumbprint(X509Certificate2 certificate2, string thumbprint)
+        public CertAndThumbprint(string filePath, X509Certificate2 certificate2, string thumbprint)
```
nathanwoctopusdeploy, line 12: what filePath

nathanwoctopusdeploy (approved):
LGTM

source/Halibut.TestUtils.CompatBinary.Base/EchoService.cs
LukeButters: Removed since it can't be used across the process boundary.

PR #238, opened 2023-05-30
Luke/client proxy cancellation

LukeButters:
# Background

Adds the ability to trigger the `ConnetCancellationToken` on individual calls to the proxy.

All changes are client side meaning we wont get surprises around sending something new to existing clients.

A `HalibutProxyRequestOptions` wrapper object is used since the cancellationToken is not a normal one and it is likely we would like different cancellation tokens or options in the future. Keeping them all as one object minimises complexity in the proxy and minimises changes to halibut.

# Results

## Before

To cancel a request not yet sent:
```
var proxy = halibutRuntime.CreateService<IFoo>(serviceEndpoint, cancellationToken);
proxy.Bar(); // May be cancelled by the cts.

// To make a new request with a new token we need to make a new proxy object.
var proxy = halibutRuntime.CreateService<IFoo>(serviceEndpoint, someOtherCanellationToken);
proxy.Bar(); // May be cancelled by the cts.
```
## After

```
var proxy = halibutRuntime.CreateService<IFoo, IClientFoo>(serviceEndpoint);

proxy.Bar(new HalibutProxyRequestOptions(cancellationToken));

// To make a new request with a new token we can just pass it to the call.
proxy.Bar(new HalibutProxyRequestOptions(someOtherCanellationToken));
```

# Pre-requisites

### DIFF
--- a/source/Halibut/ServiceModel/HalibutProxy.cs
+++ b/source/Halibut/ServiceModel/HalibutProxy.cs
@@ -159,8 +159,7 @@ namespace Halibut.ServiceModel
                 return globalCancellationToken;
             }
 
-            CancellationToken ct = (CancellationToken) halibutProxyRequestOptions.ConnectCancellationToken;
-            return CancellationTokenSource.CreateLinkedTokenSource(globalCancellationToken, ct).Token;
+            return (CancellationToken) halibutProxyRequestOptions.ConnectCancellationToken;
         }
         
         void EnsureNotError(ResponseMessage responseMessage)
--- a/source/Halibut/ServiceModel/HalibutProxy.cs
+++ b/source/Halibut/ServiceModel/HalibutProxy.cs
@@ -17,17 +17,17 @@ namespace Halibut.ServiceModel
         readonly Func<RequestMessage, CancellationToken, ResponseMessage> messageRouter;
         readonly Type contractType;
         readonly ServiceEndPoint endPoint;
-        readonly CancellationToken cancellationToken;
+        readonly CancellationToken globalCancellationToken;
         long callId;
         ILog logger;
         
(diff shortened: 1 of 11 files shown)


source/Halibut/ServiceModel/HalibutProxy.cs
```diff
@@ -144,12 +146,23 @@ RequestMessage CreateRequest(MethodInfo targetMethod, object[] args)
...
-#endif
+        CancellationToken ConnectingCancellationToken(HalibutProxyRequestOptions halibutProxyRequestOptions)
+        {
+            if (halibutProxyRequestOptions == null || halibutProxyRequestOptions.ConnectCancellationToken == null)
+            {
+                return globalCancellationToken;
+            }
+
+            CancellationToken ct = (CancellationToken) halibutProxyRequestOptions.ConnectCancellationToken;
+            return CancellationTokenSource.CreateLinkedTokenSource(globalCancellationToken, ct).Token;
```
nathanwoctopusdeploy, line 163: Is this making it even more confusing e.g. we can have 2 cancellation tokens now both cancelling connects? One for the whole proxy and one for just a request.

Thoughts on changing this so you can only have one or the other e.g. if the global one is provided (anything other than CT.None) blow up when a request one is provided.

nathanwoctopusdeploy (approved):
LGTM

source/Halibut/ServiceModel/HalibutProxy.cs
LukeButters: Solved by only allowing one token to be set by the user.

LukeButters:
[Full chain with only the client upgraded](https://build.octopushq.com/buildConfiguration/OctopusDeploy_OctopusServer_TeamCorePlatform_ChainFull/8068452?buildTab=dependencies&mode=timeline&type=snapshot)

PR #240, opened 2023-06-08
Move the port forwarder to its own project

LukeButters:
# Background

So that it can be re-used.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.TestUtils.PortForwarder/Halibut.TestUtils.PortForwarder.csproj
+++ b/source/Halibut.TestUtils.PortForwarder/Halibut.TestUtils.PortForwarder.csproj
@@ -1,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <LangVersion>9.0</LangVersion>
+    </PropertyGroup>
+
+    <PropertyGroup Condition="!$([MSBuild]::IsOSUnixLike())">
+        <TargetFrameworks>net48;net6.0</TargetFrameworks>
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
+    <PropertyGroup Condition="$([MSBuild]::IsOSUnixLike())">
+        <TargetFramework>net6.0</TargetFramework>
+    </PropertyGroup>
+    <ItemGroup>
+        <PackageReference Include="Serilog" Version="2.12.0" />
+    </ItemGroup>
+    
+</Project>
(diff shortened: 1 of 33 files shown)


source/Halibut.TestUtils.PortForwarder/Halibut.TestUtils.PortForwarder.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
nathanwoctopusdeploy, line 1: Given the plan is to move this to its own solution should the project name reflect that? That way the generated nuget package name won't change when it's split from Halibut and create confusion with 2 different port forwarders in different nuget packages.

source/Halibut.Tests/Util/PortForwarderUtil.cs
```diff
@@ -0,0 +1,12 @@
+using Halibut.Tests.Util.TcpUtils;
+
+namespace Halibut.Tests.Util
+{
+    public static class PortForwarderUtil
```
nathanwoctopusdeploy, line 5: What purpose does this serve?

LukeButters: Consistent logs and less code in each tests.

Otherwise each test will need to new up their own logger and pass it in.

nathanwoctopusdeploy: Isn't that the idea of the `ClientServiceBuilder` to abstract away all the plumbing. There is a bit of legacy code in some tests that could be cleaned up but wondering if we are just adding more things, `PortForwarderBuilder`, `PortForwarderUtil` etc for people to have to understand the usage of

LukeButters: Yeah this could be removed when:
- The ClientServiceBuilder actually makes the port forwarder, currently the caller has to build it.
- All tests use that.

That could be done in a future PR and this can be deleted.

source/Halibut.TestUtils.PortForwarder/Halibut.TestUtils.PortForwarder.csproj
LukeButters: Renamed

nathanwoctopusdeploy (approved):
LGTM

PR #241, opened 2023-06-08
Add the ability for Halibut client to cache responses from the Service

nathanwoctopusdeploy:
# Background

This PR allows a Halibut client to cache responses from the Service for selected RPC calls.

The interface used by the client can have an attribute applied that instructs Halibut to cache the response from the service for the specified duration in seconds.

```
public interface IMyService
{
    [CacheResponse(600)]
    Guid CachableCall();
}
```

The responses are cached in a `System.Runtime.Caching.MemoryCache`

By default, only non-error responses are cached.

## Cache Invalidation

A best effort is made to invalidate the cache when it is deemed stale for an endpoint. This is defined as the endpoint process restarting and is designed to take into account the endpoint being upgraded or downgraded, which could mean the response is invalid.

At least one service call that is either not cached or is cached but is a cache miss must be made for the cache mechanism to determine that the process running Halibut for an endpoint has restarted since items were cached. 

## Caching Error Responses

It is possible to cache an error response by overriding the cache behaviour on an error-by-error basis. This allows for scenarios such as caching the error response when a `NoMatchingServiceOrMethodHalibutClientException` exception is thrown.

```
IHalibutRuntime halibut = ....;
halibut.OverrideErrorResponseMessageCaching = (response) =>
{
    if (response.Error.Message.Contains("CACHE ME"))
    {
        return true;
    }

    return false;
};
```

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.TestUtils.CompatBinary.Base/EchoService.cs
+++ b/source/Halibut.TestUtils.CompatBinary.Base/EchoService.cs
@@ -30,5 +30,35 @@ namespace Halibut.TestUtils.SampleProgram.Base
             File.Delete(tempFile);
             return length;
         }
+
+        public Guid NonCachableCall()
+        {
+            return Guid.NewGuid();
+        }
+
+        public Guid CachableCall()
+        {
+            return Guid.NewGuid();
+        }
+
+        public Guid CachableCall(Guid input)
+        {
+            return Guid.NewGuid();
+        }
+
+        public Guid AnotherCachableCall()
+        {
+            return Guid.NewGuid();
+        }
+
+        public Guid CachableCallThatThrowsAnException(string exceptionMessage)
+        {
+            throw new Exception(exceptionMessage + " " + Guid.NewGuid());
+        }
+
+        public Guid TwoSecondCachableCall()
+        {
+            return Guid.NewGuid();
+        }
     }
 }
--- a/source/Halibut.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress.NET6.0.approved.cs
+++ b/source/Halibut.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress.NET6.0.approved.cs
@@ -45,6 +45,7 @@ namespace Halibut
         public Halibut.Diagnostics.ILogFactory Logs { get; }
         public Func<string, string, Halibut.UnauthorizedClientConnectResponse> OnUnauthorizedClientConnect { get; set; }
         public static bool OSSupportsWebSockets { get; }
+        public Halibut.Transport.Caching.OverrideErrorResponseMessageCachingAction OverrideErrorResponseMessageCaching { get; set; }
         protected Halibut.UnauthorizedClientConnectResponse HandleUnauthorizedClientConnect(string clientName, string thumbPrint) { }
         public bool IsTrusted(string remoteThumbprint) { }
         public Halibut.ServiceEndPoint Discover(Halibut.ServiceEndPoint endpoint) { }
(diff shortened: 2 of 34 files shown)


nathanwoctopusdeploy:
[sc-46261]

source/Halibut.Tests/SecureListenerFixture.cs
```diff
@@ -44,12 +44,12 @@ public void SecureListenerDoesNotCreateHundredsOfIoEventsPerSecondOnWindows()
             using (var opsPerSec = GetCounterForCurrentProcess("Process", "IO Other Operations/sec"))
             {
                 var client = new SecureListener(
-                    new IPEndPoint(new IPAddress(new byte[]{ 127, 0, 0, 1 }), 1093), 
+                    new IPEndPoint(new IPAddress(new byte[]{ 127, 0, 0, 1 }), 0),
```
nathanwoctopusdeploy, line 47: This was flakey on my machine due to the port being in use

source/Halibut.TestUtils.CompatBinary.Base/EchoService.cs
```diff
@@ -30,5 +30,35 @@ public int CountBytes(DataStream stream)
             File.Delete(tempFile);
             return length;
         }
+
+        public Guid NonCachableCall()
```
LukeButters, line 34: You are allowed to create new services :p

source/Halibut.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress.NET6.0.approved.cs
```diff
@@ -45,6 +45,7 @@ public class HalibutRuntime : Halibut.IHalibutRuntime, IDisposable
         public Halibut.Diagnostics.ILogFactory Logs { get; }
```
LukeButters, line 45: I wonder if we can remove these checks.

source/Halibut/ServiceModel/ServiceInvoker.cs
```diff
@@ -13,10 +13,12 @@ namespace Halibut.ServiceModel
     public class ServiceInvoker : IServiceInvoker
     {
         readonly IServiceFactory factory;
+        readonly Guid halibutRuntimeProcessIdentifier;
```
LukeButters, line 16: nit: This could be renamed to "HalibutRuntimeInstanceId" as Process does sound like a OS process. We can have many HalibutRuntime instances under a single CLR and each of those have their own Id

source/Halibut/Transport/Caching/ResponseCache.cs
```diff
@@ -0,0 +1,120 @@
...
+            responseMessageCache.Add(cacheKey, wrapper, new CacheItemPolicy { AbsoluteExpiration = DateTimeOffset.UtcNow.AddSeconds(cacheDuration) });
+        }
+
+        public void InvalidateStaleCachedResponses(ServiceEndPoint endPoint, ResponseMessage response)
+        {
+            FixMissingHalibutRuntimeProcessIdentifier(response);
+
+            var cacheItems = responseMessageCache.ToList();
+
+            foreach (var item in cacheItems)
```
LukeButters, line 59: This probably needs to be faster, since scanning 24k tentacle entries on each get capabilities call will be costly.

Computer can do 3000000000 cycles per second. Reduce that to 10ms latency so 30,000,000 cycles per 10ms.
Processing 24K tentacles so 1250 cycles to check each entry. If it takes more than that it will take more then 10ms on each call to invalidate the cache.

this PR improves on that significantly https://github.com/OctopusDeploy/Halibut/pull/242/files by only looking to invalidate entries within the service.

PR #244, opened 2023-06-21
Use client service builder in more tests

LukeButters:
# Background

Updates more tests to make use of the `ClientServiceBuilder` removing the boiler plate around setting up listening/polling halibut communications.

Doing so will make expanding test coverage by making it possible to for a single test:
- vary the type of comms e.g. Polling/Listening
- vary network conditions e.g. add latency
- to be in a form that will be easier to add backwards compatibility testing to later on.

Also removes a slight trap from the `ClientServiceBuilder`
# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Util/ClientServiceBuilder.cs
+++ b/source/Halibut.Tests/Util/ClientServiceBuilder.cs
@@ -50,11 +50,6 @@ namespace Halibut.Tests.Util
             return this;
         }
 
-        public ClientServiceBuilder WithService<TContract>(TContract implementation)
-        {
-            return WithService(() => implementation);
-        }
-
         public ClientServiceBuilder WithService<TContract>(Func<TContract> implementation)
         {
             if (serviceFactory == null) serviceFactory = new DelegateServiceFactory();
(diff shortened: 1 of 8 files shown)


source/Halibut.Tests/Util/ClientServiceBuilder.cs
```diff
@@ -50,11 +50,6 @@ public ClientServiceBuilder WithServiceFactory(IServiceFactory serviceFactory)
             return this;
         }
 
-        public ClientServiceBuilder WithService<TContract>(TContract implementation)
```
LukeButters, line 53: This was the trap it would not actually be possible to load the contract unless the interface type was specified.

sburmanoctopus: Cool.

🤔 Doesn't the left over `WithService` method still have the same issue?

E.g. would this `.WithService<IEchoService>(() => new EchoService())` still compile fine as this `.WithService(() => new EchoService())` ?

sburmanoctopus (approved):
LGTM

Nice to see centralisation removing all that code 😁 Good one!

source/Halibut.Tests/Util/ClientServiceBuilder.cs
LukeButters: Hmm good point, somehow I never fell into that trap. I wonder if the method can be declared such that the type must be an interface and not a class.

sburmanoctopus: That would be amazing! But I don't think that's a constraint that can be applied to generics 😞

PR #246, opened 2023-06-22
Simulate a network in which the last n bytes arrive late.

LukeButters:
# Background

Expands the available simulated networks to allow for sending the last n bytes late. For example if `hello` was sent over the wire and `n=1` then the port forwarder would forward `hell` and `10ms` later `o` would be sent.

The idea is to find and prevent bugs where we are unintentionally dropping read bytes. No bugs were found.

This PR also:
* Adds logging to the halibut instances created by the Client and Service builder.
* Switches the existing Test halibut logging to using `SeriLog` (will be useful later when that is improved).
* Passes the logger down to the SocketPump, not used but does make desperate debugging easier. The amount of logs coming out of the socket pump is too noisy to keep, so no logging code is kept in the pump.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Util/ClientServiceBuilder.cs
+++ b/source/Halibut.Tests/Util/ClientServiceBuilder.cs
@@ -32,6 +32,19 @@ namespace Halibut.Tests.Util
             return new ClientServiceBuilder(ServiceConnectionType.Listening, CertAndThumbprint.TentacleListening);
         }
 
+        public static ClientServiceBuilder ForMode(ServiceConnectionType serviceConnectionType)
+        {
+            switch (serviceConnectionType)
+            {
+                case ServiceConnectionType.Polling:
+                    return ClientServiceBuilder.Polling();
+                case ServiceConnectionType.Listening:
+                    return ClientServiceBuilder.Listening();
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(serviceConnectionType), serviceConnectionType, null);
+            }
+        }
+
         /// <summary>
         ///     Ie no tentacle.
         ///     In the case of listening, a TCPListenerWhichKillsNewConnections will be created. This will cause connections to
@@ -72,11 +85,20 @@ namespace Halibut.Tests.Util
         public ClientAndService Build()
         {
             serviceFactory = serviceFactory ?? new DelegateServiceFactory();
-            var octopus = new HalibutRuntime(clientCertAndThumbprint.Certificate2);
+            var octopus = new HalibutRuntimeBuilder().WithServerCertificate(clientCertAndThumbprint.Certificate2)
+                .WithLogFactory(new TestContextLogFactory("Tentacle"))
+                .Build();
             octopus.Trust(serviceCertAndThumbprint.Thumbprint);
 
             HalibutRuntime? tentacle = null;
-            if (HasService) tentacle = new HalibutRuntime(serviceFactory, serviceCertAndThumbprint.Certificate2);
+            if (HasService)
+            {
+                tentacle = new HalibutRuntimeBuilder()
+                    .WithServiceFactory(serviceFactory)
+                    .WithServerCertificate(serviceCertAndThumbprint.Certificate2)
+                    .WithLogFactory(new TestContextLogFactory("Tentacle"))
+                    .Build();
+            }
 
             var disposableCollection = new DisposableCollection();
 
--- a/source/Octopus.TestPortForwarder/PortForwarderBuilder.cs
+++ b/source/Octopus.TestPortForwarder/PortForwarderBuilder.cs
@@ -11,7 +11,7 @@ namespace Octopus.TestPortForwarder
         TimeSpan sendDelay = TimeSpan.Zero;
         readonly List<Func<BiDirectionalDataTransferObserver>> observerFactory = new();
         int? listeningPort;
-        int delaySendingLastNBytes = 0;
+        int numberOfBytesToDelaySending = 0;
         readonly ILogger logger;
 
         public PortForwarderBuilder(Uri originServer, ILogger logger)
@@ -50,7 +50,7 @@ namespace Octopus.TestPortForwarder
 
         public PortForwarderBuilder WithDelaySendingLastNBytes(int numberOfBytesToDelaySending)
         {
-            this.delaySendingLastNBytes = numberOfBytesToDelaySending;
+            this.numberOfBytesToDelaySending = numberOfBytesToDelaySending;
             return this;
         }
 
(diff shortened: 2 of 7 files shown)


LukeButters:
TODO fails because of net48.

source/Octopus.TestPortForwarder/SocketPump.cs
```diff
@@ -19,13 +20,17 @@ public class SocketPump
         readonly TimeSpan sendDelay;
 
         IDataTransferObserver dataTransferObserver;
+        readonly int delaySendingLastNBytes;
```
sburmanoctopus, line 23: Nit: This variable name threw me a little when I first saw it in the other files. First I thought it was a 'delay' (and should be a TimeSpan), then it sounded like a bool (should I delay sending last n bytes?). Then I saw it was the 'last n bytes' for 'delay sending'.

Naming is hard. Maybe an alternative could be `lastNBytesToDelaySending`?

source/Octopus.TestPortForwarder/SocketPump.cs
```diff
@@ -48,8 +71,7 @@ public async Task<SocketStatus> PumpBytes(Socket readFrom, Socket writeTo, Cance
                 dataTransferObserver.WritingData(tcpPump, buffer);
 
                 await PausePump(cancellationToken);
-
-                await WriteToSocket(writeTo, buffer.GetBuffer(), (int)buffer.Length, cancellationToken);
+                await WriteToSocketDelayingSendingTheLastNBytes(writeTo, buffer.GetBuffer(), (int)buffer.Length, numberOfBytesToDelaySending, cancellationToken);
```
sburmanoctopus, line 74: Why pass this in when it's a member variable?

source/Octopus.TestPortForwarder/SocketPump.cs
```diff
@@ -69,15 +91,34 @@ async Task PausePump(CancellationToken cancellationToken)
...
+            if (sent != bufferLength)
+            {
+                throw new Exception($"Was supported to send {bufferLength} but sent {sent}");
+            }
+        }
+        
+        static async Task<int> WriteToSocket(Socket writeTo, byte[] buffer, int initialOffset, int totalBytesToSend, CancellationToken cancellationToken)
         {
-            var offset = 0;
+            ArraySegment<byte> toSend = new ArraySegment<byte>(buffer, initialOffset, totalBytesToSend);
```
sburmanoctopus, line 114: Just wondering, instead of introducing the `Slice` below, could we not simply use the code as it was, and set `offset` to `initialOffset`?

I'm likely missing a subtlety here...

LukeButters: I originally didn't use slice but the code was cleaner with slice, then later I learned slice was not a net48 thing so I added the extension below to keep this method simpler.

source/Octopus.TestPortForwarder/SocketPump.cs
```diff
@@ -69,15 +91,34 @@ async Task PausePump(CancellationToken cancellationToken)
                 cancellationToken.ThrowIfCancellationRequested();
             }
         }
-
-        static async Task WriteToSocket(Socket writeTo, byte[] inputBuffer, int totalBytesToSend, CancellationToken cancellationToken)
+        
+        async Task WriteToSocketDelayingSendingTheLastNBytes(Socket writeTo, byte[] buffer, int bufferLength, int delaySendingLastNBytes, CancellationToken cancellationToken)
+        {
+            int howMuchToSend = bufferLength - delaySendingLastNBytes;
```
sburmanoctopus, line 97: Nit: Why aren't we using 'vars' in this file? If we can, I think we should 😁

source/Octopus.TestPortForwarder/SocketPump.cs
```diff
@@ -69,15 +91,34 @@ async Task PausePump(CancellationToken cancellationToken)
...
+            if(howMuchToSend < 0) howMuchToSend = bufferLength;
+            int sent = await WriteToSocket(writeTo, buffer, 0, howMuchToSend, cancellationToken);
+            if (howMuchToSend < bufferLength)
+            {
+                await Task.Delay(10);
```
sburmanoctopus, line 102: A possible future enhancement might be to configure this delay length. But I don't know if that's a requirement. Just an idea for later 😁

source/Octopus.TestPortForwarder/PortForwarderBuilder.cs
```diff
@@ -36,19 +37,31 @@ public PortForwarderBuilder WithDataObserver(Func<BiDirectionalDataTransferObser
...
+        /// </summary>
+        /// <param name="listeningPort"></param>
+        /// <returns></returns>
         public PortForwarderBuilder ListenOnPort(int? listeningPort)
         {
             this.listeningPort = listeningPort;
             return this;
         }
 
+        public PortForwarderBuilder WithDelaySendingLastNBytes(int numberOfBytesToDelaySending)
```
sburmanoctopus, line 51: Nit: Not a big deal. But now that we renamed to `numberOfBytesToDelaySending`, would it be nicer to rename this method to something similar too?

LukeButters: yes done.

source/Octopus.TestPortForwarder/SocketPump.cs
LukeButters: It can be done when required.

source/Octopus.TestPortForwarder/SocketPump.cs
LukeButters: Not sure

source/Octopus.TestPortForwarder/SocketPump.cs
LukeButters: It used to be a static, which I have changed it back to. For the purposes of functional is a little easier to reason about.

source/Halibut.Tests/Util/ClientServiceBuilder.cs
```diff
@@ -72,11 +85,20 @@ public ClientServiceBuilder WithPortForwarding(Func<int, PortForwarder> portForw
         public ClientAndService Build()
         {
             serviceFactory = serviceFactory ?? new DelegateServiceFactory();
-            var octopus = new HalibutRuntime(clientCertAndThumbprint.Certificate2);
+            var octopus = new HalibutRuntimeBuilder().WithServerCertificate(clientCertAndThumbprint.Certificate2)
```
sburmanoctopus, line 88: Nit: `WithServerCertificate` on new line 😁

source/Octopus.TestPortForwarder/SocketPump.cs
sburmanoctopus: I personally find it more confusing having it there. But it's not a big deal.

sburmanoctopus (approved):
LGTM

PR #248, opened 2023-06-22
Logs will show in TC, also improve the log message

LukeButters:
# Background

Logs should now show up in Teamcity under the Test.

Log messages are also improved to include the context and in teamcity include the test name (useful for tests that never finish and all you have is a single log file containing ALL test logs).

# Results

## Before
image

image

## After
image

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Util/SerilogLoggerBuilder.cs
+++ b/source/Halibut.Tests/Util/SerilogLoggerBuilder.cs
@@ -28,7 +28,7 @@ namespace Halibut.Tests.Util
             
             return new LoggerConfiguration()
                 .MinimumLevel.Debug()
-                .WriteTo.Sink(new NonProgressNUnitSink(new MessageTemplateTextFormatter(OutputTemplate)))
+                .WriteTo.Sink(new NonProgressNUnitSink(new MessageTemplateTextFormatter(outputTemplate)))
                 .CreateLogger();
         }
         
--- a/source/Halibut.Tests/Util/SerilogLoggerBuilder.cs
+++ b/source/Halibut.Tests/Util/SerilogLoggerBuilder.cs
@@ -27,6 +27,7 @@ namespace Halibut.Tests.Util
             return new LoggerConfiguration()
                 .MinimumLevel.Debug()
                 .WriteTo.Sink(new NonProgressNUnitSink(new MessageTemplateTextFormatter(outputTemplate)))
+                .Enrich.WithProperty("TestName", TestContext.CurrentContext.Test.Name)
                 .CreateLogger();
         }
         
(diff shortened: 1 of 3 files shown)


source/Halibut.Tests/Util/SerilogLoggerBuilder.cs
```diff
@@ -1,19 +1,68 @@
...
+using Serilog.Events;
 using Serilog.Formatting.Display;
 using Serilog.Sinks.NUnit;
 
 namespace Halibut.Tests.Util
 {
     public class SerilogLoggerBuilder
     {
-        const string OutputTemplate = "{Timestamp:yyyy-MM-dd HH:mm:ss.fff zzz} [{Level}] {Message}{NewLine}{Exception}";
+        const string OutputTemplate = "{Timestamp:HH:mm:ss.fff zzz} {Message}{NewLine}{Exception}";
```
sburmanoctopus, line 14: I'm being blind I think, but we've updated this private const, and it looks like we aren't using it below.

source/Halibut.Tests/Util/SerilogLoggerBuilder.cs
```diff
@@ -1,19 +1,68 @@
...
         
         public ILogger Build()
         {
+            // In teamcity we need to know what test the log is for, since we can find hung builds and only have a single file containing all log messages.
```
sburmanoctopus, line 18: I don't mind this being a TeamCity specific thing.

But I was wondering, why not just have it this way in general? Maybe it'll help the local experience as well?

Just a thought.

source/Halibut.Tests/Util/SerilogLoggerBuilder.cs
```diff
@@ -1,19 +1,68 @@
...
+        
+        public class NonProgressNUnitSink : ILogEventSink
+        {
+            private readonly MessageTemplateTextFormatter _formatter;
+
+            public NonProgressNUnitSink(MessageTemplateTextFormatter formatter) => _formatter = formatter != null ? formatter : throw new ArgumentNullException(nameof(formatter));
+
+            public void Emit(Serilog.Events.LogEvent logEvent)
+            {
+                if (logEvent == null)
```
sburmanoctopus, line 43: Style police Nits:

All places I've worked don't like new lines after 'if' without curly braces.

I'd prefer curly braces, or put them on the same line without curly braces.


Also, there's a missing 'var' below (consistency)

source/Halibut.Tests/Util/SerilogLoggerBuilder.cs
LukeButters: progress is nicer locally since that will show logs in your ide immediately, otherwise you have to wait until the test completes to see what happened. This is very problematic when a mistake is made and the test never finishes :D

source/Halibut.Tests/Util/SerilogLoggerBuilder.cs
LukeButters: good point let me fix that up.

sburmanoctopus (approved):
LGTM. Some minor style preferences, but otherwise, it seems to do what's described on the tin 🙂

PR #252, opened 2023-06-26
What pump is that?

LukeButters:
# Background

It is currently hard to determine if the bytes flowing over a port forwarder are for an existing TCP connection or for a new one.

With this change we can now log out the pump number in the data observers making it easier to see which connection data is being logged about.

A simple ID appears to be sufficient for this.

With this it would be possible to log something like:
```
[Pump:1] Client sent 45 bytes
[Pump:1] Tentacle sent 38 bytes
[Pump:1] Client sent 600 bytes
[Pump:2] Tentacle sent 100 bytes <--- This is bytes being sent over a new TCP connection since each connection gets a pump.
```

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


acodrington (approved):
I can see this being handy! :shipit:

PR #254, opened 2023-07-05
Remove PublicSurfaceArea tests

acodrington:
The `PublicSurfaceAreaFixture` does not offer any value in its current state, as any changes to the API (breaking or otherwise) are communicated to users via the versioning scheme.

As this test fixture provides no value, but does have a maintenance cost, it should be deleted.

Implemented as part of [sc-53136]


nathanwoctopusdeploy (approved):
LGTM

PR #255, opened 2023-07-06
Test Cleanup 01

nathanwoctopusdeploy:
# Background

Prior to Async Halibut we are performing some test cleanup.

This PR got a bit big and includes:

Renaming abcTests to abcFixture
Move Fixtures to sub folders to mirror where the sut lives
Move Util/* to Support/* due to conflicts with a Fixture in Util
Move other files to Support that made sense
Change some tests to use ClientServiceBuilder
Change some tests to use an attribute to run for both Polling and Listening, removing duplicate tests or just added another test case where it made sense.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Halibut.Tests/Support/TeamCityDetection.cs
```diff
@@ -0,0 +1,20 @@
+using System;
```
nathanwoctopusdeploy, line 1: This was just a move 🤷

nathanwoctopusdeploy:
[sc-53136]

source/Halibut.Tests/Support/ClientServiceBuilder.cs
```diff
@@ -78,19 +78,34 @@ public ClientServiceBuilder WithPortForwarding()
...
+        }
+
         public ClientAndService Build()
         {
             serviceFactory = serviceFactory ?? new DelegateServiceFactory();
 
-            var octopus = new HalibutRuntimeBuilder().WithServerCertificate(clientCertAndThumbprint.Certificate2)
-                .WithLogFactory(new TestContextLogFactory("Tentacle"))
-                .Build();
+            var octopusLogFactory = new TestContextLogFactory("Server");
```
LukeButters, line 98: Should this be `Client`?

source/Halibut.Tests/UsageFixture.cs
```diff
@@ -35,37 +35,23 @@ public void OctopusCanDiscoverTentacle()
...
-            var services = GetDelegateServiceFactory();
-            services.Register<ISupportedServices>(() => new SupportedServices());
-            using (var clientAndService = ClientServiceBuilder.Polling().WithServiceFactory(services).Build())
-            {
-                var svc = clientAndService.CreateClient<ISupportedServices>();
-                for (var i = 1; i < 100; i++)
+                for (var i = 0; i < 2000; i++)
                 {
-                    var i1 = i;
-                    svc.GetLocation(new MapLocation { Latitude = -i, Longitude = i }).Should().Match<MapLocation>(x => x.Latitude == i1 && x.Longitude == -i1);
```
LukeButters, line 68: Did we stop testing this call? Should we add another test that tests this call?

LukeButters (approved):
LGTM, consider adding back what looked like the dropped test

source/Halibut.Tests/UsageFixture.cs
nathanwoctopusdeploy: Does it add any value? Listening was doing one thing, Polling another. I made the Listening test do both!! We can add a 
listening and polling variations of this if we want!!

LukeButters: I think we should, more tests cost very little.

source/Halibut.Tests/Support/ClientServiceBuilder.cs
nathanwoctopusdeploy: Change 👍

source/Halibut.Tests/UsageFixture.cs
nathanwoctopusdeploy: Added the test back in with a polling and listening variation 👍

PR #265, opened 2023-07-07
Speed up fail when server throws during a data stream

LukeButters:
# Background

The test was taking 10 minutes.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Transport/PollingClient.cs
+++ b/source/Halibut/Transport/PollingClient.cs
@@ -16,25 +16,27 @@ namespace Halibut.Transport
         readonly Thread thread;
         readonly CancellationToken cancellationToken;
         bool working;
+        Func<RetryPolicy> CreateRetryPolicy;
 
         [Obsolete("Use the overload that provides a logger. This remains for backwards compatibility.")]
-        public PollingClient(Uri subscription, ISecureClient secureClient, Func<RequestMessage, ResponseMessage> handleIncomingRequest)
-            : this(subscription, secureClient, handleIncomingRequest, null)
+        public PollingClient(Uri subscription, ISecureClient secureClient, Func<RequestMessage, ResponseMessage> handleIncomingRequest, Func<RetryPolicy> createRetryPolicy)
+            : this(subscription, secureClient, handleIncomingRequest, null, createRetryPolicy)
         {
         }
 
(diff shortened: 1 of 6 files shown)


source/Halibut/Util/RetryPolicy.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Halibut.Util
 {
-    class RetryPolicy
+    public class RetryPolicy
```
LukeButters, line 6: Is forced to be public, but you can't actually set it on the runtime. So I guess someone could use this class but not with halibut.

source/Halibut/Transport/PollingClient.cs
```diff
@@ -16,25 +16,27 @@ public class PollingClient : IPollingClient
         readonly Thread thread;
         readonly CancellationToken cancellationToken;
         bool working;
+        Func<RetryPolicy> CreateRetryPolicy;
 
         [Obsolete("Use the overload that provides a logger. This remains for backwards compatibility.")]
-        public PollingClient(Uri subscription, ISecureClient secureClient, Func<RequestMessage, ResponseMessage> handleIncomingRequest)
-            : this(subscription, secureClient, handleIncomingRequest, null)
+        public PollingClient(Uri subscription, ISecureClient secureClient, Func<RequestMessage, ResponseMessage> handleIncomingRequest, Func<RetryPolicy> createRetryPolicy)
```
nathanwoctopusdeploy, line 22: Why `Func<RetryPolicy> createRetryPolicy` rather than just passing the RetryPolicy?

LukeButters: Its mutable :(

LukeButters: The retry policy is holding both the backoff exponent, min and max times as well as the stopwatches used to record how long since the last call occurred. If we passed the retry policy around then this same stopwatch would be shared by all connections to the client e.g. when a tentacle is configured to communicate to multiple octopus servers.

nathanwoctopusdeploy: 😭 Ahh yeh!!

nathanwoctopusdeploy (approved):
LGTM

PR #266, opened 2023-07-10
Adding observability into the message sizes being written

sburmanoctopus:
# Background

While trying to figure out how to do JSON serialization asynchronously, we needed to know how large the data being written/read can become.

# Results

Related to https://github.com/OctopusDeploy/Issues/issues/8229

## Before
We had no way of knowing the size of the data being read from or written to the stream.

## After

We can now register an observer with the message serializer.

There was no way to find out the number of bytes read/written from the existing streams. So we had to make a "pass through" style stream that counted how much data was written/read.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


acodrington:
The changes themselves look fine, but how does this get used? As there is no provided implementation of `IMessageSerializerObserver`, is this expected to be implemented by the calling code i.e. using `ObservableStream` etc?

sburmanoctopus:
> The changes themselves look fine, but how does this get used?

Octopus Server configures the Halibut runtime using `MessageSerializerBuilder` via the `HalibutRuntimeBuilder.WithMessageSerializer` method.

So the idea is to register a new `IMessageSerializerObserver` where it configures the message serializer.

After that, is should be hooked up correctly.

acodrington (approved):

source/Halibut/Transport/Protocol/MessageSerializer.cs
```diff
@@ -62,9 +73,13 @@ public T ReadMessage<T>(Stream stream)
         T ReadCompressedMessage<T>(Stream stream)
         {
             using (var zip = new DeflateStream(stream, CompressionMode.Decompress, true))
-            using (var bson = new BsonDataReader(zip) { CloseInput = false })
+            using (var decompressedObservableStream = new ObservableStream(zip))
```
LukeButters, line 76: Would it make sense to observe the zip side as well, since we may decide to have the zipped copy in memory rather than the unzipped copy.

source/Halibut/Transport/Protocol/MessageSerializer.cs
```diff
@@ -88,6 +104,9 @@ T ReadCompressedMessageRewindable<T>(Stream stream, IRewindableBuffer rewindable
                     {
                         rewindable.CancelBuffer();
                     }
+
+                    observer.MessageRead(decompressedObservableStream.BytesRead);
```
LukeButters, line 108: Could this be pushed into the observable stream, to avoid needing to remember to call it?

source/Halibut/Transport/Protocol/MessageSerializer.cs
```diff
@@ -75,10 +90,11 @@ T ReadCompressedMessageRewindable<T>(Stream stream, IRewindableBuffer rewindable
             try
             {
                 using (var zip = new DeflateStream(stream, CompressionMode.Decompress, true))
-                using (var bson = new BsonDataReader(zip) { CloseInput = false })
+                using (var decompressedObservableStream = new ObservableStream(zip))
```
LukeButters, line 93: Interesting that the `ObservableStream` does not need to be told to not dispose the inner stream. I guess it works because we tell the `DeflateStream` not to dispose its inner stream.

source/Halibut/Transport/Observability/ObservableStream.cs
```diff
@@ -0,0 +1,94 @@
+﻿using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Halibut.Transport.Observability
+{
+    public class ObservableStream : Stream
```
LukeButters, line 8: Is `ByteCountingStream` a better name for this stream?

LukeButters (approved):
LGTM although it does feel like we are measuring the wrong thing when reading a message.

nathanwoctopusdeploy:
[sc-53477]

source/Halibut/Transport/Protocol/MessageSerializer.cs
sburmanoctopus: Yeah, it's probably worth knowing both. That'll give us some better information to make a choice over later 🙂

source/Halibut/Transport/Observability/ObservableStream.cs
sburmanoctopus: Yeah, done 😁

source/Halibut/Transport/Protocol/MessageSerializer.cs
sburmanoctopus: I might leave it where it is so that if we enhance observability beyond this stream, we can do so easily.

source/Halibut/Transport/Protocol/MessageSerializer.cs
sburmanoctopus: Luke and I had a chat about this, and decided it was a good idea to pass in a flag in the constructor to choose what happens to the underlying stream on dispose.

This will make it clear what is going to happen to the reader/future author.

source/Halibut.Tests/Support/ReadMessage.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Halibut.Tests.Support
+{
+    public class ReadMessage
```
LukeButters, line 3: nit: `ReadMessageStats`

LukeButters: or `ReceivedMessageState`

LukeButters (approved):

PR #267, opened 2023-07-10
Reduce halibut limits in test for both net48 and net6

LukeButters:
# Background

Reduce the halibut limits in both net48 and net6. Previously the limits were reduced only got net48.

This also changes some of the limits used in test, specifically making the heart beat timeout less than the regular TCP timeout (which is a closer match to reality).

[SC-53392]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


acodrington (approved):
Love it. It is actually easier to understand now that the reader doesn't have to have some esoteric knowledge about how `app.config` gets used!

LukeButters:
Attempting to bump the timeouts higher since I am seeing some failures around 
```
System.IO.InvalidDataException: The archive entry was compressed using an unsupported compression method.
```
which is either a real bug or more likely the TCP stream was killed and the message is poorly reported.

PR #268, opened 2023-07-10
Add standard test cases

LukeButters:
# Background

Adds some standard test cases sources:

- LatestClientAndServiceTestCases
- LatestAndAPreviousVersionClientAndServiceTestCases

Which allows a test to be written once but test with all combinations of:
- Service connection types
- network conditions (perfect, with latency, etc)
- Different versions of the client and service (latest -> latest, latest client with an old service, old client with the latest service)

Each test that uses these test case sources takes a single param. That single passed object takes care of building up the Client and Service builder with the above cases (service type, network condition etc). Doing so makes it easy to add additional cases such as a proxy, as the tests wont need to change.

# Adds a shared Test Contract project

This PR adds a test contract project which holds things like IEchoService. Doing this makes it possible to send complex objects to and from the compatibility binaries.

In the new project's csproj file, the `<PrivateAssets>all</PrivateAssets>` is required so that both the comparability binaries and the halibut tests can both depend on the test contracts project. If this is not done the clr will complaining about having the wrong dll or the IDE will not compile.

# Results

See the diff of the UsageFixture for an example of the impact of the change.

# Pre-requisites


source/Halibut.TestUtils.Contracts/Halibut.TestUtils.Contracts.csproj
```diff
@@ -13,7 +13,9 @@
     </PropertyGroup>
 
     <ItemGroup>
-      <ProjectReference Include="..\Halibut\Halibut.csproj" />
+        <PackageReference Include="Halibut" Version="5.0.205">
+            <PrivateAssets>all</PrivateAssets>
+        </PackageReference>
     </ItemGroup>
```
acodrington, line 19: This is cool, I didn't know we could do this. TIL!

source/Halibut.Tests/UsageFixture.cs
```diff
@@ -159,7 +159,7 @@ public async Task StreamsCanBeSent(ClientAndServiceTestCase clientAndServiceTest
                 var data = new byte[1024 * 1024 + 15];
                 new Random().NextBytes(data);
 
-                for (var i = 0; i < 100; i++)
+                for (var i = 0; i < 1; i++)
```
acodrington, line 162: Intentional?

source/Halibut.TestUtils.Contracts/Halibut.TestUtils.Contracts.csproj
```diff
@@ -13,7 +13,9 @@
     </PropertyGroup>
 
     <ItemGroup>
-      <ProjectReference Include="..\Halibut\Halibut.csproj" />
+        <PackageReference Include="Halibut" Version="5.0.205">
+            <PrivateAssets>all</PrivateAssets>
+        </PackageReference>
```
acodrington, line 18: This is cool, I didn't know we could do this. TIL!

source/Halibut.TestUtils.CompatBinary.Base/Utils/DataStreamExtensionMethods.cs
```diff
@@ -0,0 +1,14 @@
...
+{
+    public static class DataStreamExtensionMethods
+    {
+        public static void ConfigureWriterOnReceivedDataStream(this DataStream dataStream)
+        {
+            Action<Stream> writer = stream => { dataStream.Receiver().Read(dataStreamStream => dataStreamStream.CopyTo(stream)); };
+            dataStream.ReflectionSetFieldValue("writer", writer);
+        }
+    }
+}
```
acodrington, line 14: I'm going to need you to explain this to me slowy 😄

nathanwoctopusdeploy:
[sc-53455]

source/Halibut.TestUtils.CompatBinary.Base/ServiceFactoryFactory.cs
```diff
@@ -29,9 +29,13 @@ public static DelegateServiceFactory CreateServiceFactory()
         public static DelegateServiceFactory CreateProxyingServicesServiceFactory(HalibutRuntime clientWhichTalksToLatestHalibut, ServiceEndPoint realServiceEndpoint)
         {
             var forwardingEchoService = clientWhichTalksToLatestHalibut.CreateClient<IEchoService>(realServiceEndpoint);
+            var forwardingCachingService = clientWhichTalksToLatestHalibut.CreateClient<ICachingService>(realServiceEndpoint);
+            var forwardingMultipleParametersTestService = clientWhichTalksToLatestHalibut.CreateClient<IMultipleParametersTestService>(realServiceEndpoint);
             var services = new DelegateServiceFactory();
             services.Register<IEchoService>(() => new DelegateEchoService(forwardingEchoService));
+            services.Register<ICachingService>(() => new DelegateCachingService(forwardingCachingService));
```
LukeButters, line 36: This will need to be removed as it doesn't make sense, marked as future task of the PR.

source/Halibut.TestUtils.CompatBinary.Base/Utils/DataStreamExtensionMethods.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+using System.IO;
+
+namespace Halibut.TestUtils.SampleProgram.Base.Utils
+{
+    public static class DataStreamExtensionMethods
+    {
+        public static void ConfigureWriterOnReceivedDataStream(this DataStream dataStream)
```
sburmanoctopus, line 8: Nit: It would be nice if we could return a new DataStream which does exactly what you do below.

This would leave the given data stream alone, which I think is a cleaner implementation.

sburmanoctopus (approved):
LGTM

source/Halibut.TestUtils.CompatBinary.Base/Utils/DataStreamExtensionMethods.cs
LukeButters: Added a comment on what is going on.

source/Halibut.Tests/UsageFixture.cs
LukeButters: removed

source/Halibut.TestUtils.CompatBinary.Base/Utils/DataStreamExtensionMethods.cs
LukeButters: Good idea it does now.

PR #276, opened 2023-07-11
Create a Standard set of test services, which will make it easier to align all client service builders.

LukeButters:
# Background

We would like to be able to have a single interface for all three of our Client Service builders. To simplify the compatibility bins this PR adds a concept of a standard set of services available to all Client Service builders. The result is that tests can just ask for the standard set of services which will just work no matter what Client Service builder is being used.

[SC-53455]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Support/BackwardsCompatibility/ClientAndPreviousServiceVersionBuilder.cs
+++ b/source/Halibut.Tests/Support/BackwardsCompatibility/ClientAndPreviousServiceVersionBuilder.cs
@@ -62,6 +62,14 @@ namespace Halibut.Tests.Support.BackwardsCompatibility
             return this;
         }
 
+        public ClientAndPreviousServiceVersionBuilder WithStandardServices()
+        {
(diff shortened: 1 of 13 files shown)


source/Halibut.Tests/Support/BackwardsCompatibility/ClientAndPreviousServiceVersionBuilder.cs
```diff
@@ -62,6 +62,14 @@ public ClientAndPreviousServiceVersionBuilder WithPortForwarding(Func<int, PortF
             return this;
         }
 
+        public ClientAndPreviousServiceVersionBuilder WithStandardServices()
+        {
+            // TODO: EchoService is registered by default, so we don't need to do anything else here.
```
LukeButters, line 67: Will do this in a future PR.

LukeButters (commented):
LGTM @acodrington make sense to me

acodrington (approved):

PR #278, opened 2023-07-11
Add Proxy support to the Previous Client Version and Previous Service Version Builders

nathanwoctopusdeploy:
# Background

Added support for using a Proxy to the ClientAndPreviousServiceVersionBuilder and PreviousClientVersionAndServiceBuilder

.... and a little bit of code clean-up as well!!

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.TestUtils.CompatBinary.Base/SettingsHelper.cs
+++ b/source/Halibut.TestUtils.CompatBinary.Base/SettingsHelper.cs
@@ -1,0 +1,87 @@
+﻿using System;
+using System.Security.Cryptography.X509Certificates;
+using Halibut.Transport.Proxy;
+
+namespace Halibut.TestUtils.SampleProgram.Base
+{
+    static class SettingsHelper
+    {
+        public static ProxyDetails GetProxyDetails()
+        {
+            var host = GetSetting("proxydetails_host");
+
+            if (!string.IsNullOrWhiteSpace(host))
+            {
+                var password = GetSetting("proxydetails_password");
+                var userName = GetSetting("proxydetails_username");
+                var port = GetSetting("proxydetails_port");
+                var type = GetSetting("proxydetails_type");
+
+                Console.WriteLine($"Using Proxy Details Hot:{host} Port:{port} Type:{type} Username:{userName} Password:{password}");
+
+                return new ProxyDetails(host, AsInt(port), AsPortType(type), userName, password);
+
+                ProxyType AsPortType(string s)
+                {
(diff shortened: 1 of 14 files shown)


source/Halibut.TestUtils.CompatBinary.Base/SettingsHelper.cs
```diff
@@ -0,0 +1,87 @@
+﻿using System;
+using System.Security.Cryptography.X509Certificates;
+using Halibut.Transport.Proxy;
+
+namespace Halibut.TestUtils.SampleProgram.Base
+{
+    static class SettingsHelper
```
LukeButters, line 7: This is a nice change

source/Halibut.Tests/UsageFixture.cs
```diff
@@ -57,19 +56,41 @@ public async Task OctopusCanSendMessagesToTentacle_WithEchoService(ServiceConnec
...
-        public async Task OctopusCanSendMessagesToTentacle_WithEchoService_AndAProxy(ServiceConnectionType serviceConnectionType)
+        [TestCase(ServiceConnectionType.Polling, null, null)]
+        [TestCase(ServiceConnectionType.Polling, PreviousVersions.v5_0_236_Used_In_Tentacle_6_3_417, null)]
+        [TestCase(ServiceConnectionType.Polling, null, PreviousVersions.v5_0_236_Used_In_Tentacle_6_3_417)]
+        [TestCase(ServiceConnectionType.Listening, null, null)]
+        [TestCase(ServiceConnectionType.Listening, PreviousVersions.v5_0_236_Used_In_Tentacle_6_3_417, null)]
+        [TestCase(ServiceConnectionType.Listening, null, PreviousVersions.v5_0_236_Used_In_Tentacle_6_3_417)]
+#if SUPPORTS_WEB_SOCKET_CLIENT
+        // Polling Over WebSockets does not appear to support a proxy and ignores the ProxyDetails
+        [TestCase(ServiceConnectionType.PollingOverWebSocket, null, null)]
```
LukeButters, line 67: Should we not have this test case?

LukeButters (approved):
LGTM get this in soon since we also need to make further changes to one of the builders

source/Halibut.Tests/UsageFixture.cs
nathanwoctopusdeploy: I left it there in case support gets added e.g. does it still work with a proxy if support is added! Now = yes (but it doesn't use it) Future = unknown. It may be a bit misleading though, so I can remove these and leave a note that WebSocket do not currently use Proxies

PR #291, opened 2023-07-13
Ensure the child compat test binary will always exits if the parent dies.

LukeButters:
# Background

Looking at a hung build, we found that a compatibility test binary was still running when the parent tests process was not. This resulted in the teamcity build being hung. Killing that binary resulted in the tests completing and going green.

The change here is to change compatibility test binary such that if the parent dies it also dies.

# How

This is done by the parent processes creating a file and holding a lock on that, released in the dispose method (when the running compatibility test binary is disposed).

The compatibility test binary is told the location of that file and will exit if either:
* It can get a lock on that file.
* the file no longer exists.

File locks are always released when the owning process dies, this means should the test host process die the lock is released. This will result in the compatibility test binary obtaining the lock and then exiting.

# Local Development Improvement

This change also helps local development since if the parent test process is killed in some way that doesn't kill the child, the child will detect that and die. Previously we found the compat bin would stay running, preventing future compilation.

# Helps when other bits of the test are in error

This is also handy in code that is not prepared for EVERYTHING and ANYTHING to throw an exception e.g.

```
SomeDisposable build() {
    var runningCompatBin = StartCompatBin();
    var foo = somethingThatCanThrowAnException(); // If this throws the compat bin wont be disposed, and so will stay running.
    return new SomeDisposable(runningCompatBin, foo);
}
```

with these changes, in that above example the compat bin will die when the test process completes.

# How to review this PR

Try it yourself, previously the compat binary would stay running if you forget to dispose of it now it will always be killed.

# Pre-requisites

### DIFF
--- a/source/Halibut.TestUtils.CompatBinary.Base/Services/StayAliveUntilHelper.cs
+++ b/source/Halibut.TestUtils.CompatBinary.Base/Services/StayAliveUntilHelper.cs
@@ -1,0 +1,40 @@
+using System;
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Halibut.TestUtils.SampleProgram.Base.Services
+{
+    public class StayAliveUntilHelper
+    {
+        public static void WaitUntilSignaledToDie()
+        {
+            var stayAliveFile = SettingsHelper.CompatBinaryStayAliveLockFile();
+            while (true)
+            {
+                try
+                {
+                    Console.WriteLine("Waiting to lock");
+                    var fileStreamLock = new FileStream(stayAliveFile, FileMode.Open, FileAccess.Read, FileShare.None);
+                    Console.WriteLine("Got a lock, going to die now");
+                    fileStreamLock.Dispose();
+                    File.Delete(stayAliveFile);
+                    Environment.Exit(0);
+                }
+                catch (Exception e)
+                {
+                
+                }
+
+                if (!File.Exists(stayAliveFile))
+                {
+                    Environment.Exit(0);
+                }
+                
+                Thread.Sleep(2000);
+            }
+            
+        }
+    }
+}
--- a/source/Halibut.TestUtils.CompatBinary.Base/Services/StayAliveUntilHelper.cs
+++ b/source/Halibut.TestUtils.CompatBinary.Base/Services/StayAliveUntilHelper.cs
@@ -8,7 +8,7 @@ namespace Halibut.TestUtils.SampleProgram.Base.Services
 {
     public class StayAliveUntilHelper
     {
-        public static void WaitUntilSignaledToDie()
+        public static async Task WaitUntilSignaledToDie()
         {
             var stayAliveFile = SettingsHelper.CompatBinaryStayAliveLockFile();
             while (true)
@@ -16,11 +16,20 @@ namespace Halibut.TestUtils.SampleProgram.Base.Services
                 try
                 {
                     Console.WriteLine("Waiting to lock");
-                    var fileStreamLock = new FileStream(stayAliveFile, FileMode.Open, FileAccess.Read, FileShare.None);
-                    Console.WriteLine("Got a lock, going to die now");
-                    fileStreamLock.Dispose();
-                    File.Delete(stayAliveFile);
-                    Environment.Exit(0);
+                    using (var fileStreamLock = new FileStream(stayAliveFile, FileMode.Open, FileAccess.Read, FileShare.None))
+                    {
+                        Console.WriteLine("Got a lock, going to die now");
+                        fileStreamLock.Dispose();    
+                    }
+
+                    try
+                    {
+                        File.Delete(stayAliveFile);
+                    }
+                    finally
+                    {
+                        Environment.Exit(0);
+                    }
                 }
                 catch (Exception e)
                 {
--- a/source/Halibut.TestUtils.CompatBinary.Base/SettingsHelper.cs
+++ b/source/Halibut.TestUtils.CompatBinary.Base/SettingsHelper.cs
@@ -86,14 +86,14 @@ namespace Halibut.TestUtils.SampleProgram.Base
 
         public static string CompatBinaryStayAliveLockFile()
         {
-            var tentacleCertPath = GetSetting("CompatBinaryStayAliveFilePath");
-            if (tentacleCertPath == null || tentacleCertPath.Length == 0)
+            var stayAliveFilePath = GetSetting("CompatBinaryStayAliveFilePath");
+            if (string.IsNullOrEmpty(stayAliveFilePath))
             {
                 throw new Exception("Env var CompatBinaryStayAliveFilePath must be set");
             }
             
-            Console.WriteLine($"Will die when the following file can be locked or is deleted '{tentacleCertPath}'.");
-            return tentacleCertPath;
+            Console.WriteLine($"Will die when the following file can be locked or is deleted '{stayAliveFilePath}'.");
+            return stayAliveFilePath;
         }
     }
 }
--- a/source/Halibut.TestUtils.CompatBinary.Base/SettingsHelper.cs
+++ b/source/Halibut.TestUtils.CompatBinary.Base/SettingsHelper.cs
@@ -83,5 +83,17 @@ namespace Halibut.TestUtils.SampleProgram.Base
 
             throw new Exception($"Unknown service type '{s}'");
         }
+
+        public static string CompatBinaryStayAliveLockFile()
+        {
+            var tentacleCertPath = GetSetting("CompatBinaryStayAliveFilePath");
+            if (tentacleCertPath == null || tentacleCertPath.Length == 0)
+            {
+                throw new Exception("Env var CompatBinaryStayAliveFilePath must be set");
+            }
+            
+            Console.WriteLine($"Will die when the following file can be locked or is deleted '{tentacleCertPath}'.");
+            return tentacleCertPath;
+        }
     }
 }
--- a/source/Halibut.Tests/Support/BackwardsCompatibility/CompatBinaryStayAlive.cs
+++ b/source/Halibut.Tests/Support/BackwardsCompatibility/CompatBinaryStayAlive.cs
@@ -9,26 +9,27 @@ namespace Halibut.Tests.Support.BackwardsCompatibility
         public static string StayAliveFilePathEnvVarKey = "CompatBinaryStayAliveFilePath";
         
         readonly TmpDirectory tmpDirectory;
-        public readonly string lockFile;
+        public string LockFile { get; }
         readonly FileStream fileStreamLock;
         public CompatBinaryStayAlive()
         {
             tmpDirectory = new TmpDirectory();
-            lockFile = Path.Combine(tmpDirectory.FullPath, "compat-bin-lock");
+            LockFile = Path.Combine(tmpDirectory.FullPath, "compat-bin-lock");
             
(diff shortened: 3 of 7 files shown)


source/Halibut.TestUtils.CompatBinary.Base/SettingsHelper.cs
```diff
@@ -83,5 +83,17 @@ static ServiceConnectionType AsServiceConnectionType(string s)
 
             throw new Exception($"Unknown service type '{s}'");
         }
+
+        public static string CompatBinaryStayAliveLockFile()
+        {
+            var tentacleCertPath = GetSetting("CompatBinaryStayAliveFilePath");
```
sburmanoctopus, line 89: tentacleCertPath?

source/Halibut.TestUtils.CompatBinary.Base/SettingsHelper.cs
```diff
@@ -83,5 +83,17 @@ static ServiceConnectionType AsServiceConnectionType(string s)
...
+            if (tentacleCertPath == null || tentacleCertPath.Length == 0)
```
sburmanoctopus, line 90: Nit: Why not `if (string.IsNullOfWhitespace(tentacleCertPath))`?

source/Halibut.TestUtils.CompatBinary.Base/SettingsHelper.cs
```diff
@@ -83,5 +83,17 @@ static ServiceConnectionType AsServiceConnectionType(string s)
...
+            {
+                throw new Exception("Env var CompatBinaryStayAliveFilePath must be set");
+            }
+            
+            Console.WriteLine($"Will die when the following file can be locked or is deleted '{tentacleCertPath}'.");
```
sburmanoctopus, line 95: Nit: I feel this line doesn't belong to the method that gets the file name. But rather, to the place where we use this file name (i.e. in `StayAliveUntilHelper`)

source/Halibut.TestUtils.CompatBinary.Base/Services/StayAliveUntilHelper.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        public static void WaitUntilSignaledToDie()
+        {
+            var stayAliveFile = SettingsHelper.CompatBinaryStayAliveLockFile();
+            while (true)
+            {
+                try
+                {
+                    Console.WriteLine("Waiting to lock");
+                    var fileStreamLock = new FileStream(stayAliveFile, FileMode.Open, FileAccess.Read, FileShare.None);
```
sburmanoctopus, line 19: Nit: would this be a little safer?

```
                    using (var fileStreamLock = new FileStream(stayAliveFile, FileMode.Open, FileAccess.Read, FileShare.None))
                    {
                       Console.WriteLine("Got a lock, going to die now");
                    }
```

source/Halibut.Tests/Support/BackwardsCompatibility/CompatBinaryStayAlive.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        public void Dispose()
+        {
+            try
+            {
+                fileStreamLock.Dispose();
+            }
+            catch (Exception e)
+            {
+                new SerilogLoggerBuilder().Build().ForContext<CompatBinaryStayAlive>().Warning(e, "Could not release lock");
```
nathanwoctopusdeploy, line 31: nit: build once for 31 and 40 to reduce code

source/Halibut.Tests/Support/BackwardsCompatibility/HalibutTestBinaryRunner.cs
```diff
@@ -42,11 +42,13 @@ public HalibutTestBinaryRunner(ServiceConnectionType serviceConnectionType, Cert
...
         public async Task<RunningOldHalibutBinary> Run()
         {
+            var compatBinaryStayAlive = new CompatBinaryStayAlive(); 
             var settings = new Dictionary<string, string>
             {
                 { "mode", "serviceonly" },
                 { "tentaclecertpath", serviceCertAndThumbprint.CertificatePfxPath },
-                { "octopusthumbprint", clientCertAndThumbprint.Thumbprint }
+                { "octopusthumbprint", clientCertAndThumbprint.Thumbprint },
+                {CompatBinaryStayAlive.StayAliveFilePathEnvVarKey, compatBinaryStayAlive.lockFile}
```
nathanwoctopusdeploy, line 51: very nit, consistency with formating

source/Halibut.TestUtils.CompatBinary.Base/Services/StayAliveUntilHelper.cs
```diff
@@ -0,0 +1,40 @@
...
+                {
+                
+                }
+
+                if (!File.Exists(stayAliveFile))
+                {
+                    Environment.Exit(0);
+                }
+                
+                Thread.Sleep(2000);
```
sburmanoctopus, line 35: Nit: Could we make this use `await Task.Delay` instead? I think all callers of this method are async as well.

source/Halibut.Tests/Support/BackwardsCompatibility/CompatBinaryStayAlive.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace Halibut.Tests.Support.BackwardsCompatibility
+{
+    public class CompatBinaryStayAlive : IDisposable
+    {
+
+        public static string StayAliveFilePathEnvVarKey = "CompatBinaryStayAliveFilePath";
+        
+        readonly TmpDirectory tmpDirectory;
+        public readonly string lockFile;
```
nathanwoctopusdeploy, line 12: nit: LockFIie and pref a getter

source/Halibut.Tests/Support/BackwardsCompatibility/ProxyHalibutTestBinaryRunner.cs
```diff
@@ -148,25 +150,29 @@ public class RoundTripRunningOldHalibutBinary : IDisposable
...
+                ProxyClientListenPort = proxyClientListenPort;
+                this.compatBinaryStayAlive = compatBinaryStayAlive;
             }
             public int? ServiceListenPort { get; }
             public int? ProxyClientListenPort { get; }
 
             public void Dispose()
             {
+                compatBinaryStayAlive.Dispose();
+                Thread.Sleep(10000);
```
nathanwoctopusdeploy, line 175: Was it intended to sleep only for this scenario, if so is it reliable and maybe a comment as to why we sleep here

source/Halibut.Tests/Support/BackwardsCompatibility/ProxyHalibutTestBinaryRunner.cs
```diff
@@ -148,25 +150,29 @@ public class RoundTripRunningOldHalibutBinary : IDisposable
...
             readonly TmpDirectory tmpDirectory;
+            readonly CompatBinaryStayAlive compatBinaryStayAlive;
 
             public RoundTripRunningOldHalibutBinary(
                 CancellationTokenSource cts,
                 Task runningOldHalibutTask,
                 TmpDirectory tmpDirectory,
                 int? serviceListenPort,
-                int? proxyClientListenPort)
+                int? proxyClientListenPort, CompatBinaryStayAlive compatBinaryStayAlive)
```
sburmanoctopus, line 160: Nit: New line 😁

source/Halibut.Tests/Support/BackwardsCompatibility/ProxyHalibutTestBinaryRunner.cs
sburmanoctopus, line 175: Do we need this sleep?

It would be better to not wait. Or if we have to wait, then can we sense what we are waiting for has finished?

source/Halibut.TestUtils.CompatBinary.Base/Services/StayAliveUntilHelper.cs
```diff
@@ -0,0 +1,40 @@
...
+            while (true)
+            {
+                try
+                {
+                    Console.WriteLine("Waiting to lock");
+                    var fileStreamLock = new FileStream(stayAliveFile, FileMode.Open, FileAccess.Read, FileShare.None);
+                    Console.WriteLine("Got a lock, going to die now");
+                    fileStreamLock.Dispose();
+                    File.Delete(stayAliveFile);
+                    Environment.Exit(0);
```
sburmanoctopus, line 23: Does `Environment.Exit` kill the test runner as well?

nathanwoctopusdeploy (approved):
LGTM. Few nit/explain for the next reader.
The general idea makes sense. It definitely feels like we shouldn't have to do something like this, and it adds a bit more complexity to the way we run compatibility scenarios, but it seems like we need a hammer to get stability in tests for now!!

source/Halibut.TestUtils.CompatBinary.Base/Services/StayAliveUntilHelper.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Halibut.TestUtils.SampleProgram.Base.Services
+{
+    public class StayAliveUntilHelper
+    {
+        public static void WaitUntilSignaledToDie()
+        {
+            var stayAliveFile = SettingsHelper.CompatBinaryStayAliveLockFile();
```
sburmanoctopus, line 13: How does this work if we run multiple tests in parallel? As they would all use the same directory, right? Would that not cause issue?

source/Halibut.Tests/Support/BackwardsCompatibility/ProxyHalibutTestBinaryRunner.cs
LukeButters: I was sure I removed this :(

source/Halibut.TestUtils.CompatBinary.Base/Services/StayAliveUntilHelper.cs
LukeButters: This is an external process so no

source/Halibut.TestUtils.CompatBinary.Base/Services/StayAliveUntilHelper.cs
LukeButters: This binary is started by a test in a block like this:
```
            using (var clientAndService = await PreviousClientVersionAndServiceBuilder.ForServiceConnectionType(serviceConnectionType)
                       .WithClientVersion(PreviousVersions.v5_0_429)
                       .WithEchoServiceService(echoService)
                       .Build())
```

Multiple tests in parallel would be starting multiple binaries each with which would get their own unique file.

source/Halibut.TestUtils.CompatBinary.Base/Services/StayAliveUntilHelper.cs
LukeButters: yes adding

source/Halibut.TestUtils.CompatBinary.Base/Services/StayAliveUntilHelper.cs
LukeButters: yep

source/Halibut.Tests/Support/BackwardsCompatibility/ProxyHalibutTestBinaryRunner.cs
LukeButters: It was not, i removed it one place but not both.

PR #292, opened 2023-07-13
Background thread is background thread

LukeButters:
# Background

When a halibut runtime fails to be disposed, e.g. because some exception is thrown in the builder perhaps because:
- the compat binary can not be started
-  the proxy can not be started
- the port forwarder can not be started

The test process stays running waiting for the SecureListener thread to die, which it never does.

The change here is to make the `SecureListener` behave like other threads started in Halibut (such as the polling client thread) where it is marked as a background thread. The result is the CLR is allowed to finish even if the thread is still running.

The result should hopefully be that we are less likely to keep test CLRs running when a test fails in an unexpected way.

Tentacle itself, expects this behaviour since polling tentacles work and the source code has a concept of waiting for the user to exit tentacle see [here](https://github.com/OctopusDeploy/OctopusTentacle/blob/main/source/Octopus.Tentacle/Commands/RunAgentCommand.cs#L119).

## Async alignment

This behaviour is not something that is aligned with converting everything to async, since async contexts don't have a concept of keeping the CLR running while they still have work todo.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Transport/SecureListener.cs
+++ b/source/Halibut/Transport/SecureListener.cs
@@ -92,6 +92,7 @@ namespace Halibut.Transport
             {
                 Name = "Accept connections on " + listener.LocalEndpoint
             };
+            backgroundThread.IsBackground = true;
             backgroundThread.Start();
 
             return ((IPEndPoint)listener.LocalEndpoint).Port;


nathanwoctopusdeploy:
Can we push this through a Tentacle build chain to make sure it behaves as expected 👍

nathanwoctopusdeploy (approved):
LGTM - Just the tentacle chain to make sure it behaves as expected 👍

LukeButters:
Running a tentacle build [here](https://github.com/OctopusDeploy/OctopusTentacle/pull/545).

PR #295, opened 2023-07-13
Don't make available the caching service when it does not make sense

LukeButters:
# Background

The `ICachingService` doesn't make sense to be tested with when the client is an old client, since that old client wont have the new caching attributes we are trying to test.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Support/BackwardsCompatibility/LatestClientAndPreviousServiceVersionBuilder.cs
+++ b/source/Halibut.Tests/Support/BackwardsCompatibility/LatestClientAndPreviousServiceVersionBuilder.cs
@@ -87,6 +87,14 @@ namespace Halibut.Tests.Support.BackwardsCompatibility
             return this;
         }
 
+        IClientAndServiceBuilder IClientAndServiceBuilder.WithCachingService() => WithCachingService();
+        
+        public IClientAndServiceBuilder WithCachingService()
+        {
+            // TODO actually make the old service only have caching service if this is called.
+            return this;
+        }
+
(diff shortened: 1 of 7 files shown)


source/Halibut.Tests/Support/BackwardsCompatibility/LatestClientAndPreviousServiceVersionBuilder.cs
```diff
@@ -87,6 +87,14 @@ public LatestClientAndPreviousServiceVersionBuilder WithStandardServices()
             return this;
         }
 
+        IClientAndServiceBuilder IClientAndServiceBuilder.WithCachingService() => WithCachingService();
+        
+        public IClientAndServiceBuilder WithCachingService()
+        {
+            // TODO actually make the old service only have caching service if this is called.
```
LukeButters, line 94: This is for a future PR similar to the above.

source/Halibut.Tests/Support/LatestClientAndLatestServiceBuilderExtensionMethods.cs
```diff
@@ -26,10 +26,5 @@ public static LatestClientAndLatestServiceBuilder WithReadDataStreamService(this
         {
             return builder.WithService<IReadDataStreamService>(() => new ReadDataStreamService());
         }
-
```
LukeButters, line 29: Moved into the class so it can implement the interface

acodrington (approved):

PR #297, opened 2023-07-14
Don't register services unless asked to

LukeButters:
# Background

Some client service builders would always register services, this PR changes that so that by default they all give no services and must be explicitly registered.

Also removes a log message in the external binary that logs every 2 seconds.

[SC-53455]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Halibut.TestUtils.CompatBinary.Base/Services/StayAliveUntilHelper.cs
```diff
@@ -15,7 +15,6 @@ public static async Task WaitUntilSignaledToDie()
             {
                 try
                 {
-                    Console.WriteLine("Waiting to lock");
```
LukeButters, line 18: This was printed every 2s

acodrington (approved):

PR #302, opened 2023-07-14
Add tests for sending/receiving complex objects

acodrington:
Adds a test which sends and receives complex objects containing multiple DataStreams via Halibut.

Part of [sc-53157]

### DIFF
--- a/source/Halibut.TestUtils.Contracts/DataStreamExtensionMethods.cs
+++ b/source/Halibut.TestUtils.Contracts/DataStreamExtensionMethods.cs
@@ -1,0 +1,21 @@
+﻿using System;
+using System.IO;
+
+namespace Halibut.TestUtils.Contracts
+{
+    public static class DataStreamExtensionMethods
+    {
+        public static string ReadAsString(this DataStream stream)
+        {
+            string result = null;
+            stream.Receiver().Read(s =>
+            {
+                using (var reader = new StreamReader(s))
+                {
+                    result = reader.ReadToEnd();
+                }
+            });
+            return result;
+        }
+    }
+}
--- a/source/Halibut.TestUtils.Contracts/IComplexObjectService.cs
+++ b/source/Halibut.TestUtils.Contracts/IComplexObjectService.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 
 namespace Halibut.TestUtils.Contracts
 {
@@ -19,7 +20,7 @@ namespace Halibut.TestUtils.Contracts
     public class ComplexResponse
     {
         public string RequestId;
-        public IList<DataStream> Payloads;
+        public IDictionary<Guid, IList<DataStream>> Payloads;
 
         public ComplexObjectChild Child;
     }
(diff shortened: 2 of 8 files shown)


source/Halibut.TestUtils.Contracts/DataStreamExtensionMethods.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using System.IO;
+
+namespace Halibut.TestUtils.Contracts
+{
+    public static class DataStreamExtensionMethods
```
acodrington, line 6: This class would be better suited in `Halibut.Tests`, but it gets used in the `ComplexObjectService` itself so it's easier to keep it here.

source/Halibut.TestUtils.Contracts/IComplexObjectService.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        public ComplexObjectChild Child;
+    }
+
+    public class ComplexResponse
+    {
+        public string RequestId;
+        public IList<DataStream> Payloads;
+
+        public ComplexObjectChild Child;
```
LukeButters, line 24: This could be made more complex with `Map<Guid, Set<ComplexObjectChild>>` or `Map<Guid, Map<Enum, Set<ComplexObjectChild>>>`

LukeButters: Oh `sed s/Map/Dict/` :stuck_out_tongue:

source/Halibut.TestUtils.Contracts/IComplexObjectService.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            return DataStream.FromString(copyNumber + ": " + payload);
+        }
+
+        public static DataStream ProcessFirst(DataStream payload)
+        {
+            return DataStream.FromString("First: " + payload.ReadAsString());
+        }
+
+        public static DataStream ProcessSecond(DataStream payload)
```
LukeButters, line 67: Maybe `DataStream PrefixDataStreamWith(DataStrream payload, string prefix)` rather than a generic `Process`.

source/Halibut.Tests/UsageFixture.cs
```diff
@@ -159,5 +159,45 @@ public async Task StreamsCanBeSentWithProgressReporting(ClientAndServiceTestCase
...
+                        First = DataStream.FromString(childPayload1),
+                        Second = DataStream.FromString(childPayload2)
+                    }
+                });
+
+                result.RequestId.Should().Be(requestId);
+
+                var payloads = result.Payloads.ToArray();
+                payloads.Length.Should().Be(3);
+                payloads[0].ReadAsString().Should().Be(ComplexObjectService.ProcessPayload(1, payload).ReadAsString());
```
LukeButters, line 194: Consider making the assert against a string literal, that way the test is easier to read and we don't find out later that shenanigans are at play e.g. `ProcessPayload` has some bug of always returning empty which happens to make the test meaningless.

LukeButters (approved):

PR #303, opened 2023-07-16
Fix buffer overread on net48

LukeButters:
# Background

A message in the Halibut protocol is:
```
<DeflateStream><DataStreams with prefix and postix of size><Control Messages>
```

The protocol sends no indication of when the `DeflateStream` ends or its length. This means we must find the length of the deflate stream by decompressing the stream. The `DeflateStream` implmentation in both dotnet and netframework will both read in more bytes from the underlying stream if the compressed bytes do not end at 8KB (the buffer sized used by the deflate stream).  This can be a problem since the following `DataStreams` and/or `Control Messages` may be consumed and dropped by the deflate stream.

This was fixed for dotnet [here](https://github.com/OctopusDeploy/Halibut/pull/154) but was never fixed for netframework.

The issue of overeading bytes during normal Halibut operation in netframework could not be reproduced. It appears to be that so long as the sender makes distinct writes between the `<DeflateStream>` section and the following sections, when reading a stream in netframework it will always happen to read bytes such that bytes from distinct writes are never mixed. EXAMPLE:

Sender sends:
```
sslStream.write("hello");
sslStream.write(StringOf12Kb);
sslStream.write("cya");
sslStream.flush();
```

If in net48 after the flush has completed:
```
sslStream.read(byteArray, 0, 2); // reads "he"
sslStream.read(byteArray, 0, 99999999); // only reads "llo" despite more being available 
// and requesting lots of data. (This is why it appears to work in net48)
sslStream.read(byteArray, 0, 99999999); // Reads the 12kb string
sslStream.read(byteArray, 0, 99999999); // reads "cya"
```

This is different to net6 where the operations would be:
```
sslStream.read(byteArray, 0, 2); // reads "he"
sslStream.read(byteArray, 0, 99999999); // reads "llo" and the 12kb string and "cya"
```

This means that communications to net48 are very delicate and depend on however we happen to be creating encrypted frames of data. If For example writing of a message is changed to writing all of the data to a `MemoryStream` and then that is sent over the sslStream in one go net48 WILL overread.

[SC-53378]

# The change

The change here is to change net48 to be aligned with net6, such that it no longer depends on the chance that data is sent in a way that the sslStream which by chance happens to force distinct reads between the frames for communications to be successful.

This also add's a try catch in the reflector. Although never seen as something that is required should an underlying implementation of dotnet change halibut wont be immediately broken.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Transport/Observability/ByteCountingStream.cs
+++ b/source/Halibut/Transport/Observability/ByteCountingStream.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.IO;
+using System.Net.Security;
 using System.Threading;
 using System.Threading.Tasks;
 
--- a/source/Halibut/Transport/Observability/ByteCountingStream.cs
+++ b/source/Halibut/Transport/Observability/ByteCountingStream.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.IO;
-using System.Net.Security;
 using System.Threading;
 using System.Threading.Tasks;
 
--- a/source/Halibut.Tests/Transport/Protocol/DeflateStreamInputBufferReflectorFixture.cs
+++ b/source/Halibut.Tests/Transport/Protocol/DeflateStreamInputBufferReflectorFixture.cs
@@ -2,6 +2,8 @@
 using System.IO;
 using System.IO.Compression;
 using System.Text;
+using Halibut.Diagnostics;
+using Halibut.Transport;
 using Halibut.Transport.Protocol;
 using NUnit.Framework;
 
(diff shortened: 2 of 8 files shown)


source/Halibut.Tests/Halibut.Tests.csproj
```diff
@@ -10,7 +10,7 @@
     <Nullable>enable</Nullable>
   </PropertyGroup>
   <PropertyGroup Condition="!$([MSBuild]::IsOSUnixLike())">
-    <TargetFrameworks>net48;net6.0</TargetFrameworks>
+    <TargetFrameworks>net48</TargetFrameworks>
```
nathanwoctopusdeploy, line 13: Nope 😉

LukeButters: hahaha, for some reason my windows wouldn't show me a run all under net48 option so I did it the manual way lol.

pawelpabich:
> The change here is to change net48 to be aligned with net6, such that it no longer depends on the chance that data is sent in a way that the sslStream which by chance happens to force distinct reads between the frames for communications to be successful

💯. I will leave the details to of this PR to others, but the ideas make perfect sense to me.  Having one, more resilient, way of reading data is a great improvement.

source/Halibut/Transport/Observability/ByteCountingStream.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.IO;
+using System.Net.Security;
```
nathanwoctopusdeploy, line 3: ?

nathanwoctopusdeploy (approved):
LGTM. Hopefully this is the end of Buffer Overreading 🥳

PR #321, opened 2023-07-19
Update tests to cover more use cases

acodrington:
This PR updates the test case source attribute used by a couple of Halibut tests, so that they also cover the following use cases:
* Polling over web socket
* Previous client -> Latest server
* Latest client -> previous server

This will complete the following section of the [Hailbut test discovery doc](https://docs.google.com/document/d/1z0qff6yenIsGCdvHBseOfuVvVRUn18B6__1abHwT4dk/edit):
image

### DIFF
--- a/source/Halibut.Tests/UsageFixture.cs
+++ b/source/Halibut.Tests/UsageFixture.cs
@@ -125,7 +125,7 @@ namespace Halibut.Tests
         }
 
         [Test]
-        [LatestAndPreviousClientAndServiceVersionsTestCases(testNetworkConditions: false)]
+        [LatestAndPreviousClientAndServiceVersionsTestCases(testWebSocket:true, testNetworkConditions: false)]
         [FailedWebSocketTestsBecomeInconclusive]
         public async Task SupportsDifferentServiceContractMethods(ClientAndServiceTestCase clientAndServiceTestCase)
         {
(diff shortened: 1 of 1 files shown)


source/Halibut.Tests/UsageFixture.cs
```diff
@@ -125,7 +125,7 @@ public async Task StreamsCanBeSent(ClientAndServiceTestCase clientAndServiceTest
         }
 
         [Test]
-        [LatestAndPreviousClientAndServiceVersionsTestCases(testNetworkConditions: false)]
+        [LatestAndPreviousClientAndServiceVersionsTestCases(testWebSocket:true, testNetworkConditions: false)]
```
LukeButters, line 128: My understanding of the outcome of standup is we would be doing very little tests on websockets since they are so unreliable.

@nathanwoctopusdeploy ?

LukeButters: Oh i see so we should keep marking the tests as run on websockets but don't actually run them. Thanks will continue to do that on my changes as well.

LukeButters (approved):

source/Halibut.Tests/UsageFixture.cs
acodrington: Yeah - if we have existing tests that already pass w/ websockets (which we can verify locally, by enabling websockets in `ServiceConnectionTypes` and testing not net48), then we can mark the websocket columns as 'done' for those tests.

But I don't think we want to spend time writing new tests, or fixing old tests, to get them to pass w/ websockets.

acodrington:
[sc-53157]

PR #326, opened 2023-07-19
Lock around finding a free a tcp port for websockets and when it is used

nathanwoctopusdeploy:
# Background

This PR adds some best-effort locking to try and minimise port conflicts due to multiple tests trying to use the same port in parallel.

A race condition exists between finding a free TCP port and then using the free TCP port, which allows another test to steal the port or think the port is free.

This PR attempts to add some locking around this so it is thread-safe between tests.

This will not help when the out-of-test-process CompatBinary find and uses a TCP port directly itself as the locking is only process wide.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.TestUtils.CompatBinary.Base/ProxyServiceForwardingRequestToClient.cs
+++ b/source/Halibut.TestUtils.CompatBinary.Base/ProxyServiceForwardingRequestToClient.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Threading;
 using System.Threading.Tasks;
 using Halibut.Tests.Support;
 using Halibut.TestUtils.SampleProgram.Base.LogUtils;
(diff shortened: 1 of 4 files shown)


source/Halibut.Tests/Support/LatestClientAndLatestServiceBuilder.cs
```diff
@@ -270,12 +270,14 @@ public async Task<ClientAndService> Build(CancellationToken cancellationToken)
             }
             else if (serviceConnectionType == ServiceConnectionType.PollingOverWebSocket)
             {
+                using var tcpPortConflictLock = await TcpPortHelper.WaitForLock(cancellationToken);
```
acodrington, line 273: Funny given our conversation around `using` scopes earlier, but I think an explicit scope with `{ }` would be a bit nicer here, as you could get rid of the lock variable and the scope would be more obvious.

I suppose an explicitly named var _does_ make intent clearer...

source/Halibut.Tests/Support/BackwardsCompatibility/LatestClientAndPreviousServiceVersionBuilder.cs
```diff
@@ -194,12 +194,15 @@ public async Task<ClientAndService> Build(CancellationToken cancellationToken)
             }
             else if (serviceConnectionType == ServiceConnectionType.PollingOverWebSocket)
             {
+                using var tcpPortConflictLock = await TcpPortHelper.WaitForLock(cancellationToken);
```
acodrington, line 197: Same same, see other comment.

source/Halibut.TestUtils.CompatBinary.Base/ProxyServiceForwardingRequestToClient.cs
```diff
@@ -57,15 +58,18 @@ public static async Task Run()
                         realServiceEndpoint = new ServiceEndPoint(new Uri("poll://SQ-TENTAPOLL"), serviceCert.Thumbprint, proxyDetails);
                         break;
                     case ServiceConnectionType.PollingOverWebSocket:
-                        var webSocketListeningPort = TcpPortHelper.FindFreeTcpPort();
-                        var webSocketPath = SettingsHelper.GetSetting("websocketpath");
-                        var webSocketListeningUrl = $"https://+:{webSocketListeningPort}/{webSocketPath}";
+                        using (await TcpPortHelper.WaitForLock(CancellationToken.None))
```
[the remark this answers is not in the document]
acodrington: So is its inclusion just in case we implement at cross-process lock in `TcpPortHelper` we won't need additional changes?

acodrington (approved):

PR #334, opened 2023-07-20
Uplift parallel test to test backwards compatibility

acodrington:
This PR updates the `SendMessagesToTentacleInParallel` test so that it tests all relevant backwards compatibility scenarios.

Also added in a try/catch within the Thread operation (nUnit _really_ doesn't like unhandled exceptions being thrown within threads), and some sanity checks to ensure all threads complete successfully.

[sc-53157]

### DIFF
--- a/source/Halibut.Tests/ParallelRequestsFixture.cs
+++ b/source/Halibut.Tests/ParallelRequestsFixture.cs
@@ -5,6 +5,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using FluentAssertions;
 using Halibut.Logging;
+using Halibut.Tests.Support;
 using Halibut.Tests.Support.TestAttributes;
 using Halibut.Tests.Support.TestCases;
 using Halibut.TestUtils.Contracts;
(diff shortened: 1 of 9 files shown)


source/Halibut.Tests/ParallelRequestsFixture.cs
```diff
@@ -1,46 +1,55 @@
...
-            var services = new DelegateServiceFactory();
-            services.Register<IReadDataStreamService>(() => new ReadDataStreamService());
-
-            using (var clientAndService = await LatestClientAndLatestServiceBuilder
-                       .ForServiceConnectionType(serviceConnectionType)
-                       .WithHalibutLoggingLevel(LogLevel.Info)
-                       .WithServiceFactory(services)
-                       .WithServiceFactory(services).Build(CancellationToken))
+            using var clientAndService = await clientAndServiceTestCase.CreateTestCaseBuilder()
+                .WithHalibutLoggingLevel(LogLevel.Trace)
```
nathanwoctopusdeploy, line 22: Revert loglevel to Info otherwise this will generate a lot of log messages

source/Halibut.Tests/ParallelRequestsFixture.cs
```diff
@@ -1,46 +1,55 @@
...
                 var dataStreams = CreateDataStreams();
 
-                var messagesAreSentTheSameTimeSemaphore = new Semaphore(0, dataStreams.Length);
+                var messagesAreSentTheSameTimeSemaphore = new SemaphoreSlim(0, dataStreams.Length);
 
+                int threadCount = 64;
+                int threadCompletionCount = 0;
                 var threads = new List<Thread>();
-                for (var i = 0; i < 64; i++)
+                var exceptions = new List<Exception>();
```
nathanwoctopusdeploy, line 35: This should probably be a `ConcurrentBag<T>` to be thread safe

nathanwoctopusdeploy (approved):
LGTM. Just the log level and a slight concurrency improvement

LukeButters (approved):

PR #341, opened 2023-07-21
Fix a WebSocket bug where a SendAsync was not being awaited

nathanwoctopusdeploy:
# Background

A WebSocket SendAsync call was not being awaited, resulting in the possibility of two SendAsync calls being in flight at the same time on the WebSocket, which is not supported

# Results

## Before

`System.InvalidOperationException: There is already one outstanding 'SendAsync' call for this WebSocket instance. ReceiveAsync and SendAsync can be called simultaneously, but at most one outstanding operation for each of them is allowed at the same time.`

## After

No error

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Transport/Protocol/WebSocketStream.cs
+++ b/source/Halibut/Transport/Protocol/WebSocketStream.cs
@@ -78,11 +78,11 @@ namespace Halibut.Transport.Protocol
                 .ConfigureAwait(false).GetAwaiter().GetResult();
         }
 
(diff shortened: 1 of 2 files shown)


LukeButters (approved):

nathanwoctopusdeploy:
Fixes https://github.com/OctopusDeploy/OctopusTentacle/issues/597

PR #346, opened 2023-07-25
Improve test attribute

acodrington:
Makes the way we test different service connection types more consistent.


LukeButters (approved):
LGTM

PR #353, opened 2023-07-26
Enable blame crash

LukeButters:
# Background

Enables blame crash, as recommended by builds team.

Hang dumps are not enabled, since I wasn't sure if the timeout applied to an individual test or the entire test run.

[SC-53909]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -110,7 +110,8 @@ class Build : NukeBuild
                     .SetConfiguration(Configuration)
                     .SetFramework(framework)
                     .EnableNoBuild()
-                    .EnableNoRestore());
+                    .EnableNoRestore()
+                    .EnableBlameCrash());
 
             });
     }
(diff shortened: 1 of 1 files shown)


build/Build.cs
```diff
@@ -110,7 +110,9 @@ ITargetDefinition TestDefinition(ITargetDefinition targetDefinition, Target depe
                     .SetConfiguration(Configuration)
                     .SetFramework(framework)
                     .EnableNoBuild()
-                    .EnableNoRestore());
+                    .EnableNoRestore()
+                    .EnableBlameCrash()
+                    .SetBlameCrashDumpType("full"));
```
LukeButters, line 115: full, since mini would not contain the heap which is needed to view async stacks afaik

nathanwoctopusdeploy (approved):

PR #354, opened 2023-07-27
Making tests use ClientAndServiceTestCase

sburmanoctopus:
[sc-54985]

# Background

As part of the async work, we need to ensure existing behaviour works for both sync and async versions.

The plan for the future is to enhance the test case attributes to do both sync and async test cases.

# Results

## Before

Many tests did not use the 'test case' system for creating their "client and service" objects.

## After

We tried to make as many test cases as we could use the test case attributes, so that they will naturally test sync and async in the future.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Halibut.Tests/BackwardsCompatibility/TestOldClientWithNewService.cs
```diff
@@ -1,54 +0,0 @@
...
-using Halibut.TestUtils.Contracts;
-using NUnit.Framework;
-
-namespace Halibut.Tests.BackwardsCompatibility
-{
-    public class TestOldClientWithNewService : BaseTest
-    {
-        [Test]
-        [TestCaseSource(typeof(ServiceConnectionTypesToTestExcludingWebSockets))]
-        public async Task SimplePreviousClientTest(ServiceConnectionType serviceConnectionType)
```
sburmanoctopus, line 17: These tests were deleted because they are covered in the `UsageFixture`

source/Halibut.Tests/CancellationViaClientProxyFixture.cs
```diff
@@ -52,9 +52,11 @@ public async Task CancellationCanBeDoneViaClientProxy(ClientAndServiceTestCase c
         }
 
         [Test]
-        public async Task CannotHaveServiceWithHalibutProxyRequestOptions()
+        [LatestClientAndLatestServiceTestCases(testNetworkConditions: false, testWebSocket: false, testPolling:false)]
```
sburmanoctopus, line 55: To specifically get just the "Listening" test cases, we had to turn off everything else.

We have done it this way throughout these tests.

I wonder if there is a trap here for the future, because if we add another type, then these tests will start running those as test cases.

Maybe we should look at making it easier to specify "inclusions" rather than "exclusions"?

source/Halibut.Tests/ListeningConnectRetryFixture.cs
```diff
@@ -12,10 +14,12 @@ namespace Halibut.Tests
...
     {
         [Test]
-        public async Task ListeningRetriesAttemptsUpToTheConfiguredValue()
+        [LatestClientAndLatestServiceTestCases(testNetworkConditions: false, testWebSocket: false, testPolling:false)]
+        public async Task ListeningRetriesAttemptsUpToTheConfiguredValue(ClientAndServiceTestCase clientAndServiceTestCase)
         {
             TcpConnectionsCreatedCounter tcpConnectionsCreatedCounter = null;
-            using (var clientAndService = await LatestClientAndLatestServiceBuilder.Listening()
+            using (var clientAndService = await clientAndServiceTestCase.CreateTestCaseBuilder()
+                       .As<LatestClientAndLatestServiceBuilder>()
```
sburmanoctopus, line 22: `As<LatestClientAndLatestServiceBuilder>()` is needed to support `WithEchoService`.

I can remove `As<LatestClientAndLatestServiceBuilder>()` and replace it with `WithStandardServices`.

I think this could be a nicer implementation, but am unsure if this will result in any unforeseen consequences.

FYI, sometimes there are other reasons for needing `As<LatestClientAndLatestServiceBuilder>()`. This is specifically talking about the echo service.

source/Halibut.Tests/Timeouts/TimeoutsApplyDuringHandShake.cs
```diff
@@ -20,10 +21,14 @@ namespace Halibut.Tests.Timeouts
     public class TimeoutsApplyDuringHandShake : BaseTest
     {
         [Test]
+        [LatestClientAndLatestServiceTestCasesPOC(testNetworkConditions: false, testWebSocket: false, additionalParameters: new object[] { true, 1 })]
```
sburmanoctopus, line 24: To get away from needing `ServiceConnectionTypesToTestExcludingWebSockets`, I had to find a way to pass the parameters into the method.

This was fiddly, and this POC works.

Do we like this as an approach? If so, we could add the POC logic into the real one.

What better options do we have?

I thought about adding a list of 'value sources', but that starts looking quite ugly...

source/Halibut.Tests/BackwardsCompatibility/InvocationExceptionsAreMappedCorrectlyFixture.cs
```diff
@@ -12,13 +11,11 @@ namespace Halibut.Tests.BackwardsCompatibility
...
         [Test]
-        [TestCaseSource(typeof(ServiceConnectionTypesToTest))]
+        [LatestClientAndPreviousServiceVersionsTestCases(testNetworkConditions: false)]
         [FailedWebSocketTestsBecomeInconclusive]
-        public async Task OldInvocationExceptionMessages_AreMappedTo_ServiceInvocationHalibutClientException(ServiceConnectionType serviceConnectionType)
+        public async Task OldInvocationExceptionMessages_AreMappedTo_ServiceInvocationHalibutClientException(ClientAndServiceTestCase clientAndServiceTestCase)
         {
-            using (var clientAndService = await LatestClientAndPreviousServiceVersionBuilder
-                       .ForServiceConnectionType(serviceConnectionType)
-                       .WithServiceVersion(PreviousVersions.v5_0_429.ServiceVersion.ForServiceConnectionType(serviceConnectionType))
```
sburmanoctopus, line 21: This test used a specific version of Halibut that is later than the version used in `LatestClientAndPreviousServiceVersionsTestCases`. After [this discussion](https://octopusdeploy.slack.com/archives/C04JFJXECS1/p1690424471049419), we decided that v5_0_429 was not used anywhere after this PR, and we could delete that project.

source/Halibut.Tests/PollingClientConnectionHandlingFixture.cs
```diff
@@ -15,16 +17,17 @@ namespace Halibut.Tests
     public class PollingClientConnectionHandlingFixture : BaseTest
     {
         [Test]
-        [TestCase(ServiceConnectionType.Polling)]
-#if SUPPORTS_WEB_SOCKET_CLIENT
-        [TestCase(ServiceConnectionType.PollingOverWebSocket)]
+        [LatestClientAndLatestServiceTestCases(testNetworkConditions: false, testListening: false
+#if !SUPPORTS_WEB_SOCKET_CLIENT
```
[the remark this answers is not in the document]
LukeButters: You can remove this as its done in `ServiceConnectionTypes`

source/Halibut.Tests/ListeningConnectRetryFixture.cs
LukeButters: Either or :D

source/Halibut.Tests/Timeouts/TimeoutsApplyDuringHandShake.cs
LukeButters: Either that or call the method within `LatestClientAndLatestServiceTestCases` to be an `IEnumerator`.

source/Halibut.Tests/CancellationViaClientProxyFixture.cs
LukeButters: Leave as is to force people to review existing tests, safety first!

source/Halibut.Tests/PollingClientConnectionHandlingFixture.cs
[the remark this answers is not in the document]
sburmanoctopus: Done!

source/Halibut.Tests/Timeouts/TimeoutsApplyDuringHandShake.cs
sburmanoctopus: I went with enhancing `LatestClientAndLatestServiceTestCases`. Mainly because I think the test cases look neater this way. Plus, it's not too much of an extension to `LatestClientAndLatestServiceTestCases` 🙂

LukeButters (approved):

PR #355, opened 2023-07-28
Halibut Memory Optimisations

nathanwoctopusdeploy:
# Background

Change the MemoryCache in the HalibutRuntime to only be created when needed. While the impact on Tentacleof creating this is minimal, it creates memory noise when running hundreds of HalibutRuntimes in the same Process.

CancellationTokenSource memory leak. Fix some issues where cancellation token sources were not being disposed of.

## Before

This is potentially contrived results as it is running 1 Halibut Client and 100 Halibut Services in the same process but there is an obvious CancellationTokenSource disposal issue.

image

image

## After

Gone

image

image

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Transport/Protocol/ControlMessageReader.cs
+++ b/source/Halibut/Transport/Protocol/ControlMessageReader.cs
@@ -59,7 +59,7 @@ namespace Halibut.Transport.Protocol
 
         internal async Task<string> ReadControlMessageAsync(Stream stream)
         {
-            var cts = GetCancellationTokenSourceFromStreamReadTimeout(stream);
+            using var cts = GetCancellationTokenSourceFromStreamReadTimeout(stream);
             StringBuilder sb = new StringBuilder();
             while (true)
             {
--- a/source/Halibut/Transport/SecureWebSocketListener.cs
+++ b/source/Halibut/Transport/SecureWebSocketListener.cs
@@ -226,6 +226,7 @@ namespace Halibut.Transport
         public void Dispose()
         {
             cts.Cancel();
+            cts.Dispose();
             log.Write(EventType.ListenerStopped, "Listener stopped");
         }
     }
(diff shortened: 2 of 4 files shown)


sburmanoctopus (approved):
LGTM. Nice catch 🙂 

I did a quick look at the solution, and noticed heaps of cases in our tests where we aren't disposing either. This isn't important for this PR, but thought it worth mentioning for future PR/sharpening.

source/Halibut/Transport/Protocol/ControlMessageReader.cs
```diff
@@ -59,7 +59,7 @@ internal async Task<string> ReadUntilNonEmptyControlMessageAsync(Stream stream)
 
         internal async Task<string> ReadControlMessageAsync(Stream stream)
         {
-            var cts = GetCancellationTokenSourceFromStreamReadTimeout(stream);
+            using var cts = GetCancellationTokenSourceFromStreamReadTimeout(stream);
```
sburmanoctopus, line 62: In previous versions of VS, you could setup code analysis to catch instances where disposable objects were not being disposed.

Maybe this is something we should think about doing in this solution some time in the future?

LukeButters: Oops this was me

source/Halibut/Transport/SecureWebSocketListener.cs
```diff
@@ -226,6 +226,7 @@ static void EnsureCertificateIsValidForListening(X509Certificate2 certificate)
         public void Dispose()
         {
             cts.Cancel();
+            cts.Dispose();
             log.Write(EventType.ListenerStopped, "Listener stopped");
```
LukeButters, line 230: If this log message is important it might be best to put it in a finally.

LukeButters (approved):

source/Halibut/Transport/Protocol/ControlMessageReader.cs
LukeButters: https://github.com/dotnet/runtime/issues/29970 This person talks about disposing these timer CTS as being important.

PR #356, opened 2023-07-28
Add basic support of testing async and sync clients in one test.

LukeButters:
# Background

Adds support for testing both async and sync Halibut clients in a single test as well as adds a first cut support for async clients (although sync calls are made under the hood).

Tests that wish to test both paths must:
 - Make use of `clientAndService.CreateClient<ISomeService, IAsyncClientSomeService>()`.
 - use the async methods on `IAsyncClientSomeService` and await them.
 - For now, must opt in with `testAsyncAndSyncClients: true` in the attribute.
 
The `LatestAndPreviousClientAndServiceVersionsTestCases` and `LatestClientAndLatestServiceTestCases` will force sync and async clients when `testAsyncAndSyncClients: true` is set. Under the hood the relevant builder will ask the `HalibutRuntime` to create either a sync or async proxy depending on the test case. When a `sync` proxy is created the `AdaptSyncProxyToAsyncProxy` is used to wrap that `sync` proxy into a `async` proxy. In this way, a single test can just make async calls on async client side services, and the test can map that to a `sync` proxy returned by the `HalibutProxy`.

## Async proxy support added.

This PR adds async proxy support by creating a new `CreateAsyncClient` call, which uses the new `HalibutProxyWithAsync` class. In this way the new async code is complete separate from the existing sync code.

Async methods on Interfaces require their async methods to end in `Async` e.g. `SayHello()` on `IEchoService` must be `SayHelloAsync` on `IAsyncClientEchoService`. The `HalibutProxyWithAsync` will trim off the `Async` part and convert the return type to its non async versions e.g. `Task<string>` becomes `string` before sending over the wire. This means the calls are backwards compatible since we never send anything that suggests the client is async to the service.

The new `CreateClientV2<TService, TClientService>` method call must be used to create async clients. The `TClientService` may contain the `async` methods while `TService` must be the interface the `Service` actually implements. A call to this method might be:
```
CreateAsyncClient<IEchoService, IAsyncClientEchoService>
```
Where `IEchoService` is the sync version of the interface which the service implements and `IAsyncClientEchoService` is a async version of the interface where each method is changed to return a `Task` or `Task<T>` and ends in `Async`.

Note that the `HalibutProxyWithAsync` calls into the existing code which quickly becomes sync. TODO in future PRs is to make that all async.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Support/BackwardsCompatibility/LatestClientAndPreviousServiceVersionBuilder.cs
+++ b/source/Halibut.Tests/Support/BackwardsCompatibility/LatestClientAndPreviousServiceVersionBuilder.cs
@@ -5,6 +5,7 @@ using System.Threading.Tasks;
 using Halibut.Logging;
 using Halibut.TestProxy;
 using Halibut.Tests.Support.Logging;
+using Halibut.Tests.TestServices.AsyncSyncCompat;
 using Halibut.Transport.Proxy;
 using Octopus.TestPortForwarder;
 using Serilog.Extensions.Logging;
@@ -22,6 +23,7 @@ namespace Halibut.Tests.Support.BackwardsCompatibility
         LogLevel halibutLogLevel;
         OldServiceAvailableServices availableServices = new(false, false);
         bool hasService = true;
+        ForceClientProxyType? forceClientProxyType;
 
         LatestClientAndPreviousServiceVersionBuilder(ServiceConnectionType serviceConnectionType, CertAndThumbprint serviceCertAndThumbprint)
         {
@@ -157,6 +159,17 @@ namespace Halibut.Tests.Support.BackwardsCompatibility
         {
             return NoService();
         }
+        
+        IClientAndServiceBuilder IClientAndServiceBuilder.WithForcingClientProxyType(ForceClientProxyType forceClientProxyType)
+        {
+            return WithForcingClientProxyType(forceClientProxyType);
+        }
+        
+        public LatestClientAndPreviousServiceVersionBuilder WithForcingClientProxyType(ForceClientProxyType forceClientProxyType)
+        {
+            this.forceClientProxyType = forceClientProxyType;
+            return this;
+        }
 
         async Task<IClientAndService> IClientAndServiceBuilder.Build(CancellationToken cancellationToken)
         {
@@ -283,7 +296,7 @@ namespace Halibut.Tests.Support.BackwardsCompatibility
                 throw new NotSupportedException();
             }
 
-            return new ClientAndService(octopus, runningOldHalibutBinary, serviceUri, serviceCertAndThumbprint, portForwarder, disposableCollection, proxy, proxyDetails, cancellationTokenSource);
+            return new ClientAndService(octopus, runningOldHalibutBinary, serviceUri, serviceCertAndThumbprint, portForwarder, disposableCollection, proxy, proxyDetails, forceClientProxyType, cancellationTokenSource);
         }
 
         public class ClientAndService : IClientAndService
@@ -293,6 +306,7 @@ namespace Halibut.Tests.Support.BackwardsCompatibility
             readonly CertAndThumbprint serviceCertAndThumbprint; // for creating a client
             readonly DisposableCollection disposableCollection;
             readonly ProxyDetails? proxyDetails;
+            readonly ForceClientProxyType? forceClientProxyType;
             readonly CancellationTokenSource cancellationTokenSource;
 
             public ClientAndService(HalibutRuntime client,
@@ -303,6 +317,7 @@ namespace Halibut.Tests.Support.BackwardsCompatibility
                 DisposableCollection disposableCollection,
                 HttpProxyService? httpProxy,
                 ProxyDetails? proxyDetails,
+                ForceClientProxyType? forceClientProxyType,
                 CancellationTokenSource cancellationTokenSource)
             {
                 Client = client;
@@ -313,6 +328,7 @@ namespace Halibut.Tests.Support.BackwardsCompatibility
                 HttpProxy = httpProxy;
                 this.disposableCollection = disposableCollection;
                 this.proxyDetails = proxyDetails;
+                this.forceClientProxyType = forceClientProxyType;
                 this.cancellationTokenSource = cancellationTokenSource;
             }
 
@@ -320,9 +336,9 @@ namespace Halibut.Tests.Support.BackwardsCompatibility
             public PortForwarder? PortForwarder { get; }
             public HttpProxyService? HttpProxy { get; }
 
-            public TService CreateClient<TService>(CancellationToken? cancellationToken = null, string? remoteThumbprint = null)
+            public TService CreateClient<TService>(CancellationToken? cancellationToken = null)
             {
-                return CreateClient<TService>(s => { }, cancellationToken ?? CancellationToken.None, remoteThumbprint);
+                return CreateClient<TService>(s => { }, cancellationToken ?? CancellationToken.None);
             }
 
             public TService CreateClient<TService>(Action<ServiceEndPoint> modifyServiceEndpoint)
@@ -330,11 +346,11 @@ namespace Halibut.Tests.Support.BackwardsCompatibility
                 return CreateClient<TService>(modifyServiceEndpoint, CancellationToken.None);
             }
 
-            public TService CreateClient<TService>(Action<ServiceEndPoint> modifyServiceEndpoint, CancellationToken cancellationToken, string? remoteThumbprint = null)
+            public TService CreateClient<TService>(Action<ServiceEndPoint> modifyServiceEndpoint, CancellationToken cancellationToken)
             {
-                var serviceEndpoint = new ServiceEndPoint(serviceUri, remoteThumbprint ?? serviceCertAndThumbprint.Thumbprint, proxyDetails);
+                var serviceEndpoint = new ServiceEndPoint(serviceUri, serviceCertAndThumbprint.Thumbprint, proxyDetails);
                 modifyServiceEndpoint(serviceEndpoint);
-                return Client.CreateClient<TService>(serviceEndpoint, cancellationToken);
+                return new AdaptToSyncOrAsyncTestCase().Adapt<TService>(forceClientProxyType, Client, serviceEndpoint, cancellationToken);
             }
 
             public TClientService CreateClient<TService, TClientService>()
@@ -346,7 +362,7 @@ namespace Halibut.Tests.Support.BackwardsCompatibility
             {
                 var serviceEndpoint = new ServiceEndPoint(serviceUri, serviceCertAndThumbprint.Thumbprint, proxyDetails);
                 modifyServiceEndpoint(serviceEndpoint);
-                return Client.CreateClient<TService, TClientService>(serviceEndpoint);
+                return new AdaptToSyncOrAsyncTestCase().Adapt<TService, TClientService>(forceClientProxyType, Client, serviceEndpoint);
             }
 
             public void Dispose()
--- a/source/Halibut.Tests/Support/TestAttributes/LatestClientAndLatestServiceTestCasesAttribute.cs
+++ b/source/Halibut.Tests/Support/TestAttributes/LatestClientAndLatestServiceTestCasesAttribute.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections;
-using System.Collections.Generic;
 using System.Linq;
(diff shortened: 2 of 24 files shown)


source/Halibut.Tests/Support/BackwardsCompatibility/LatestClientAndPreviousServiceVersionBuilder.cs
```diff
@@ -157,6 +159,17 @@ IClientAndServiceBuilder IClientAndServiceBuilder.NoService()
         {
             return NoService();
         }
+        
+        IClientAndServiceBuilder IClientAndServiceBuilder.WithForcingClientProxyType(ForceProxyClientType forceProxyClientType)
```
nathanwoctopusdeploy, line 163: `ClientProxyType` and `ProxyClientType` - should these be consistent?

source/Halibut.Tests/Support/TestAttributes/LatestClientAndLatestServiceTestCasesAttribute.cs
```diff
@@ -48,11 +48,18 @@ public static IEnumerable GetEnumerator(bool testWebSocket, bool testNetworkCond
                 {
                     serviceConnectionTypes.Remove(ServiceConnectionType.Polling);
                 }
+                
+                ForceProxyClientType[] proxyClientTypesToTest = new ForceProxyClientType[0];
+                if (testAsyncAndSyncClients)
+                {
+                    proxyClientTypesToTest = ForceProxyClientTypeValues.Values;
```
nathanwoctopusdeploy, line 55: We have an existing pattern e.g. ServiceConnection`Types.All` - should we keep these copnsistent

source/Halibut.Tests/Support/TestCases/ClientAndServiceTestCase.cs
```diff
@@ -35,13 +38,18 @@ public IClientAndServiceBuilder CreateTestCaseBuilder()
...
+            }
+
             return builder;
         }
         
         public override string ToString()
         {
             // This is used as the test parameter name, so make this something someone can understand in teamcity or their IDE.
-            return $"{ServiceConnectionType}, {ClientAndServiceTestVersion.ToString(ServiceConnectionType)}, {NetworkConditionTestCase}, RecommendedIterations: {RecommendedIterations}";
+            return $"{ServiceConnectionType}, {ClientAndServiceTestVersion.ToString(ServiceConnectionType)}, {NetworkConditionTestCase}, RecommendedIters: {RecommendedIterations}, {ForceProxyClientType?.ToString()??""}";
```
nathanwoctopusdeploy, line 52: nit: when null is this leaving us with `, ` on the end of tests

source/Halibut.Tests/Support/TestCases/ClientAndServiceTestCasesBuilder.cs
```diff
@@ -35,7 +39,24 @@ public IEnumerable<ClientAndServiceTestCase> Build()
...
-                        yield return new ClientAndServiceTestCase(serviceConnectionType, networkConditionTestCase, recommendedIterations, clientServiceTestVersion);
+                        if (!forceProxyClientTypes.Any())
+                        {
+                            yield return new ClientAndServiceTestCase(serviceConnectionType, networkConditionTestCase, recommendedIterations, clientServiceTestVersion, null);
+                        }
+                        else
+                        {
+                            if (clientServiceTestVersion.IsPreviousClient())
+                            {
+                                yield return new ClientAndServiceTestCase(serviceConnectionType, networkConditionTestCase, recommendedIterations, clientServiceTestVersion, null);
```
nathanwoctopusdeploy, line 50: Do we end up with 2 tests that test the same thing for previous client in this case?

source/Halibut.Tests/TestServices/Async/IAsyncEchoService.cs
```diff
@@ -0,0 +1,15 @@
+using System.Threading.Tasks;
+
+namespace Halibut.Tests.TestServices.Async
+{
+    public interface IAsyncEchoService
```
nathanwoctopusdeploy, line 5: Do we now end up with 4 different interfaces to support async and client cancellation if we add new interfaces rather than reusing existing e.g. `IEchoService`, `IClientEchoService`, `IAsyncEchoService` and `IAsyncClientEchoService`

If we just added async overloads to `IClientEchoService` we don't add all these variations.

source/Halibut/ServiceModel/IMessageRouter.cs
```diff
@@ -0,0 +1,39 @@
+using System;
+using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
+using Halibut.Transport.Protocol;
+
+namespace Halibut.ServiceModel
+{
+    public interface IMessageRouter
```
nathanwoctopusdeploy, line 9: This seems like complexity that adds no real value. e.g. can't this just be

`(proxy as HalibutProxyWithAsync).Configure(SendOutgoingRequest, SendOutgoingRequestAsync, typeof(TService), endpoint, logger, CancellationToken.None);`

Sync is on the removal path eventually so this plumbing to abstract the fact there is a sync and an async version of a method that Async Proxy needs to know about for a medium duration doesn't seem necessary.

source/Halibut.Tests/TestServices/AsyncSyncCompat/AdaptToTestCase.cs
```diff
@@ -0,0 +1,69 @@
+using System;
+using System.Reflection;
+using System.Threading;
+using Halibut.Tests.Support;
+
+namespace Halibut.Tests.TestServices.AsyncSyncCompat
+{
+    public class AdaptToTestCase
```
nathanwoctopusdeploy, line 8: Maybe a hint this is about Async and Sync Adaption in the class name

source/Halibut.Tests/TestServices/AsyncSyncCompat/AdaptToTestCase.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// <param name="halibutRuntime"></param>
+        /// <param name="serviceEndpoint"></param>
+        /// <typeparam name="TService"></typeparam>
+        /// <returns></returns>
+        /// <exception cref="Exception"></exception>
+        public TService Adapt<TService>(ForceProxyClientType? forceProxyClientType, HalibutRuntime halibutRuntime, ServiceEndPoint serviceEndpoint, CancellationToken cancellationToken)
+        {
+            if (forceProxyClientType == null)
+            {
+                // TODO what are we calling here?
```
nathanwoctopusdeploy, line 23: ??

source/Halibut/IHalibutRuntime.cs
```diff
@@ -45,7 +45,23 @@ public interface IHalibutRuntime : IDisposable
         /// TClientService would be: IClientFoo { void Bar(string, HalibutProxyRequestOptions); }
         /// </typeparam>
         /// <returns></returns>
-        public TClientService CreateClient<TService, TClientService>(ServiceEndPoint endpoint);
+        //public TClientService CreateClient<TService, TClientService>(ServiceEndPoint endpoint);
```
nathanwoctopusdeploy, line 48: Don't we need this for the way Tentacle currently passes the CancellationToken per request by using a IClientInterface?

nathanwoctopusdeploy: I'm surprised/confused we don't have tests for this in Halibut!!

source/Halibut.Tests/Support/BackwardsCompatibility/LatestClientAndPreviousServiceVersionBuilder.cs
LukeButters: Going with `ProxyClientType`

source/Halibut.Tests/Support/TestCases/ClientAndServiceTestCasesBuilder.cs
LukeButters: I think we only yield twice in the next foreach, here we yeild once with null

source/Halibut.Tests/TestServices/AsyncSyncCompat/AdaptToTestCase.cs
LukeButters: Done

source/Halibut.Tests/TestServices/Async/IAsyncEchoService.cs
LukeButters: Renamed this to IAsynceClientEchoService

source/Halibut/ServiceModel/IMessageRouter.cs
LukeButters: Removed

source/Halibut.Tests/WhenCallingAServiceThatDoesNotExist.cs
```diff
@@ -20,10 +21,10 @@ public async Task AServiceNotFoundHalibutClientExceptionShouldBeRaisedByTheClien
                        .WithHalibutLoggingLevel(LogLevel.Info)
                        .Build(CancellationToken))
             {
-                var echo = clientAndService.CreateClient<IEchoService>();
-                Func<string> readAsyncCall = () => echo.SayHello("Say hello to a service that does not exist.");
+                var echo = clientAndService.CreateClient<IEchoService, IAsyncClientEchoService>();
```
nathanwoctopusdeploy, line 24: Did you want to leave this testing only testing sync for now or did you miss the attribute override?

source/Halibut/ServiceModel/HalibutProxyWithAsync.cs
```diff
@@ -0,0 +1,172 @@
...
+            this.contractType = contractType;
+            this.endPoint = endPoint;
+            this.globalCancellationToken = cancellationToken;
+            this.configured = true;
+            this.logger = logger;
+        }
+
+        public override object Invoke(MethodInfo targetMethod, object[] args)
+        {
+            throw new NotImplementedException();
```
nathanwoctopusdeploy, line 34: nit: NotSupportedException or something along those lines with a helpful message to the caller

source/Halibut/ServiceModel/HalibutProxyWithAsync.cs
```diff
@@ -0,0 +1,172 @@
...
+            }
+
+            return (T) result;
+        }
+
+        async Task<(MethodInfo, object)> MakeRpcCall(MethodInfo asyncMethod, object[] args)
+        {
+            var serviceMethod = AsyncCompatibilityHelper.FindMatchingSyncMethod(asyncMethod, contractType);
+
+            if (!configured)
```
nathanwoctopusdeploy, line 59: nit: should this be first?

source/Halibut/ServiceModel/HalibutProxyWithAsync.cs
```diff
@@ -0,0 +1,172 @@
...
+            return (serviceMethod, response.Result);
+        }
+
+        RequestMessage CreateRequest(MethodInfo targetMethod, object[] args)
+        {
+            var activityId = Guid.NewGuid();
+
+            var request = new RequestMessage
+            {
+                Id = contractType.Name + "::" + targetMethod.Name + "[" + Interlocked.Increment(ref callId) + "] / " + activityId,
```
nathanwoctopusdeploy, line 81: Would it be useful to hint that this was done via an async call somehow? The activity id gets used in the LogMessage so may be useful for diagnostic purposes

nathanwoctopusdeploy (approved):
Nice job! Looks great

source/Halibut/ServiceModel/HalibutProxyWithAsync.cs
LukeButters: Good idea! TIL

source/Halibut/ServiceModel/HalibutProxyWithAsync.cs
LukeButters: yes

source/Halibut.Tests/WhenCallingAServiceThatDoesNotExist.cs
LukeButters: missed it thanks

source/Halibut/ServiceModel/HalibutProxyWithAsync.cs
LukeButters: good idea, done!

PR #359, opened 2023-07-31
Add MultipleRequestsCanBeInFlightInParallel test

acodrington:
This PR adds a test to verify that many requests can be "in-flight" simultaneously.

[sc-53157]

# How to review this PR

The test _does not_ clean up the 'file signal' file that gets created by `LockService`. I've tried multiple approaches, but each potential solution has caused test failures due to either race conditions or 'file is being used by another process' exceptions. The files are all 0-bytes, so this won't cause any storage problems on the build agents or local dev environments, but it would be better to clean up after ourselves here - any ideas?

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/ParallelRequestsFixture.cs
+++ b/source/Halibut.Tests/ParallelRequestsFixture.cs
@@ -7,6 +7,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using FluentAssertions;
 using Halibut.Logging;
+using Halibut.Tests.Support;
 using Halibut.Tests.Support.TestAttributes;
 using Halibut.Tests.Support.TestCases;
 using Halibut.TestUtils.Contracts;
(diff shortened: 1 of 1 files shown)


source/Halibut.Tests/ParallelRequestsFixture.cs
```diff
@@ -15,54 +17,116 @@ namespace Halibut.Tests
...
+                    {
+                        exceptions.Add(e);
+                    }
+                });
+                thread.Start();
+                threads.Add(thread);
+            }
+
+            // Wait for all requests to be started
+            while (requestStartedFilePaths.Any(p => !File.Exists(p)))
```
LukeButters, line 66: You can use `Wait.For()` here, or at least pass a cancellation token to `Task.Delay`

LukeButters (approved):
LGTM

PR #360, opened 2023-07-31
Rename Async methods that lie to SynchronouslyAsync

nathanwoctopusdeploy:
# Background

In preparation for adding Asyn Halibut we need to rename some methods that are already 'Async' but really perform sync io under the covers.

This PR renames those methods from `Async` to `SynchronouslyAsync`

These will eventually be removed when Halibut is Async only.

## AsyncEx

This PR also marks some very old code copied from [AsyncEx](https://github.com/StephenCleary/AsyncEx) as Obsolete so that it is not used in new code as it should be removed in favour of built in ways of doing this or an updated version of the code from AsyncEx (as a reference to the library if needed)

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


LukeButters (approved):

nathanwoctopusdeploy:
[sc-55196]

PR #366, opened 2023-08-01
Async pending request queue

sburmanoctopus:
[sc-55192]

# Background

The pending request queue needs to become async as part of the Halibut async change

# Results

## Before
The pending request queue would do most of its work with blocking synchronous code.

**There were also some bugs:**

If we queued and processed an item, we would "Set" the waiter to "have items". But it would never reset this. This results in the next Tentacle to dequeue to not wait the 20 seconds it was supposed to.

There was a memory leak when queueing an item, and an exception occurred. The item would remain in the `inProgress` list forever.

## After
An async version of the pending request queue now exists.

We now reset the "has items in the queue" status so that the next Tentacle will wait the full 20 seconds.

We now remove the item from the `inProgress` queue when an exception happens.

Now that we have the `SyncOrAsync` enum, the `PendingRequestQueueFixture` tests were refactored to use that instead of the `AsyncTestCase` (which is now deleted)

The existing tests used the sync version of the queue by default. They have been reworked to choose the async/sync version based off the given test case.

Also, we converted `Assert.ThrowsAsync` to use `AssertionExtensions.Should(...).ThrowsAsync`, as that deals with exception polymorphism. 

# How to review this PR
In the async pending request queue, please have a look at where we should be respecting the cancellation token or not. The places I did not use it was because a cancelation at that point should not break the logic.

Quality :heavy_check_mark:

# Pre-requisites


source/Halibut.Tests/ServiceModel/PendingRequestQueueFixture.cs
```diff
@@ -99,27 +99,31 @@ public async Task QueueAndWait_WhenPollingRequestQueueTimeoutIsReached_WillStopW
...
+        public async Task QueueAndWait_WhenRequestIsDequeued_ButPollingRequestQueueTimeoutIsReached_AndPollingRequestMaximumMessageProcessingTimeoutIsReached_WillStopWaitingAndClearRequest(SyncOrAsync syncOrAsync)
         {
             // Arrange
             const string endpoint = "poll://endpoint001";
 
-            var sut = new PendingRequestQueueBuilder().WithEndpoint(endpoint).Build();
+            var sut = new PendingRequestQueueBuilder()
+                .WithAsync(syncOrAsync)
+                .WithEndpoint(endpoint)
+                .WithPollingQueueWaitTimeout(TimeSpan.Zero) // Remove delay, otherwise we wait the full 20 seconds for DequeueAsync at the end of the test
```
sburmanoctopus, line 116: Turns out the "Set" bug was allowing some of these tests to finish quickly and ignore the "Delay" (as they weren't waiting like they should). 

It's the `DequeueAsync` at the end that caused the delay.

So when we fixed it, these started taking 20 seconds to finish... These tests are not exercising that logic, so make the delay 0 so the tests do not take longer than they should.

source/Halibut.Tests/ServiceModel/PendingRequestQueueFixture.cs
```diff
@@ -333,23 +348,54 @@ public async Task DequeueAsync_WithNothingQueued_WillWaitPollingQueueWaitTimeout
         }
 
         [Test]
-        [TestCaseSource(typeof(AsyncTestCase))]
-        public async Task DequeueAsync_WillContinueWaitingUntilItemIsQueued(bool async)
+        public async Task DequeueAsync_WithNothingQueued_WillWaitPollingQueueWaitTimeout_AfterPreviousRequestWasQueuedAndDequeued()
```
sburmanoctopus, line 351: The test for the "Set" bug fix

source/Halibut.Tests/ServiceModel/PendingRequestQueueFixture.cs
```diff
@@ -360,54 +406,111 @@ public async Task DequeueAsync_WillContinueWaitingUntilItemIsQueued(bool async)
...
+            await queueAndWaitTask;
+
+            var singleDequeuedRequest = dequeueTasks.Should().ContainSingle(t => t.Result != null).Subject.Result;
+            singleDequeuedRequest.Should().Be(request);
+
+        }
+
+        [Test]
+        [SyncAndAsync]
+        public async Task ApplyResponse_AfterRequestHasBeenCollected_AndWaitingHasBeenCancelled_ResponseIsIgnored(SyncOrAsync syncOrAsync)
```
sburmanoctopus, line 447: There was no real way to test the memory leak fix. This was the best usecase I could think of that came close.

source/Halibut/ServiceModel/PendingRequestQueueAsync.cs
```diff
@@ -0,0 +1,250 @@
...
+                    }
+                }
+                else
+                {
+                    log.Write(EventType.MessageExchange, "Request {0} timed out before it could be collected by the polling endpoint", request);
+                    SetResponse(ResponseMessage.FromException(request, new TimeoutException($"A request was sent to a polling endpoint, but the polling endpoint did not collect the request within the allowed time ({request.Destination.PollingRequestQueueTimeout}), so the request timed out.")));
+                }
+            }
+
+            async Task<bool> WaitForResponseToBeSet(TimeSpan timeout, CancellationToken cancellationToken)
```
sburmanoctopus, line 210: This used to be a simple call to `waiter.Wait(request.Destination.PollingRequestQueueTimeout, cancellationToken)`.

But the new libraries do not support a timeout.

This code works, but I am open to better ways of doing this.

source/Halibut/ServiceModel/PendingRequestQueueAsync.cs
```diff
@@ -0,0 +1,250 @@
...
+        {
+            this.log = log;
+            this.pollingQueueWaitTimeout = pollingQueueWaitTimeout;
+        }
+        
+        public async Task<ResponseMessage> QueueAndWaitAsync(RequestMessage request, CancellationToken cancellationToken)
+        {
+            var pending = new PendingRequest(request, log);
+
+            using (await queueLock.LockAsync(cancellationToken))
```
sburmanoctopus, line 36: As per [Luke's question](https://octopusdeploy.slack.com/archives/C04JFJXECS1/p1690869232305179), if we decide we should, then I should retro `ConfigureAwait` in this PR...

source/Halibut/ServiceModel/PendingRequestQueueAsync.cs
```diff
@@ -0,0 +1,250 @@
...
+using Halibut.Transport.Protocol;
+using Nito.AsyncEx;
+
+namespace Halibut.ServiceModel
+{
+    public class PendingRequestQueueAsync : IPendingRequestQueue
+    {
+        readonly ConcurrentQueue<PendingRequest> queue = new();
+        readonly Dictionary<string, PendingRequest> inProgress = new();
+        readonly SemaphoreSlim queueLock = new(1, 1);
```
LukeButters, line 16: Nice, name! its easy to understand what it is locking.

source/Halibut/ServiceModel/PendingRequestQueueAsync.cs
LukeButters: Possible using a `WaitHandle` with `WaitOne` could avoid this. However I am not sure which direction I trust more.

In any case this is sufficient.

source/Halibut.Tests/Support/PendingRequestQueueFactoryBuilder.cs
```diff
@@ -0,0 +1,57 @@
...
+            this.logFactory = logFactory;
+        }
+
+        public PendingRequestQueueFactoryBuilder WithSyncOrAsync(SyncOrAsync syncOrAsync)
+        {
+            this.syncOrAsync = syncOrAsync;
+            return this;
+        }
+
+        public PendingRequestQueueFactoryBuilder WithDecorator(Func<ILogFactory, IPendingRequestQueueFactory, IPendingRequestQueueFactory> createDecorator)
```
LukeButters, line 27: Nice

source/Halibut.Tests/PollingTentacleDequeuesRequestsInOrderFixture.cs
```diff
@@ -28,20 +29,23 @@ public async Task QueuedUpRequestsShouldBeDequeuedInOrder(ClientAndServiceTestCa
                        {
                            return new FuncPendingRequestQueueFactory(uri =>
                            {
-                               pendingRequestQueue = new PendingRequestQueue(logFactory.ForEndpoint(uri), TimeSpan.FromSeconds(1));
+                               pendingRequestQueue = new PendingRequestQueueBuilder()
+                                   .WithLog(logFactory.ForEndpoint(uri))
+                                   .WithPollingQueueWaitTimeout(TimeSpan.FromSeconds(1))
+                                   .WithSyncOrAsync(clientAndServiceTestCase.SyncOrAsync)
```
LukeButters, line 35: Good idea.

LukeButters (approved):

LukeButters:
Something to consider adding is a test which tests the queue can really handle 5K concurrent in flight requests, where the existing queue should fail to do that.

PR #367, opened 2023-08-01
Make IConnectionFactory support async

LukeButters:
# Background

Makes IConnectionFactory support async, nothing calls these async methods yet.

[SC-53211]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Transport/Proxy/HttpProxyClient.cs
+++ b/source/Halibut/Transport/Proxy/HttpProxyClient.cs
@@ -250,6 +250,51 @@ namespace Halibut.Transport.Proxy
                 throw new ProxyException($"Connection to proxy host {ProxyHost} on port {ProxyPort} failed: {ex.Message}", ex, true);
             }
         }
+        
+        public async Task<TcpClient> CreateConnectionAsync(string destinationHost, int destinationPort, TimeSpan timeout, CancellationToken cancellationToken)
+        {
+            try
+            {
+                // if we have no connection, create one
+                if (TcpClient == null)
+                {
+                    if (string.IsNullOrEmpty(ProxyHost))
+                        throw new ProxyException("ProxyHost property must contain a value", false);
+
+                    if (ProxyPort <= 0 || ProxyPort > 65535)
+                        throw new ProxyException("ProxyPort value must be greater than zero and less than 65535", false);
+
+                    if(ProxyHost.Contains("://"))
+                        throw new ProxyException("The proxy's hostname cannot contain a protocol prefix (eg http://)", false);
+
+
+                    //  create new tcp client object to the proxy server
+                    TcpClient = tcpClientFactory();
+
+                    // attempt to open the connection
+                    log.Write(EventType.Diagnostic, "Connecting to proxy at {0}:{1}", ProxyHost, ProxyPort);
+                    await TcpClient.ConnectWithTimeoutAsync(ProxyHost, ProxyPort, timeout, cancellationToken);
+                }
+
+                //  send connection command to proxy host for the specified destination host and port
+                await SendConnectionCommandAsync(destinationHost, destinationPort, cancellationToken);
+
+                // return the open proxied tcp client object to the caller for normal use
+                return TcpClient;
+            }
+            catch (AggregateException ae)
+            {
+                var se = ae.InnerExceptions.OfType<SocketException>().FirstOrDefault();
+                if (se != null)
+                    throw new ProxyException($"Connection to proxy host {ProxyHost} on port {ProxyPort} failed: {se.Message}", se, true);
+
+                throw;
+            }
+            catch (SocketException ex)
+            {
+                throw new ProxyException($"Connection to proxy host {ProxyHost} on port {ProxyPort} failed: {ex.Message}", ex, true);
+            }
+        }
 
 
         void SendConnectionCommand(string host, int port)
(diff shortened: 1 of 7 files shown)


source/Halibut/Transport/Proxy/HttpProxyClient.cs
```diff
@@ -287,6 +332,42 @@ void SendConnectionCommand(string host, int port)
...
+            //ignore all of them]
+            //<CR><LF>    // Last Empty Line
+
+            // create an byte response array  
+            var response = new byte[TcpClient.ReceiveBufferSize];
+            var sbuilder = new StringBuilder();
+
+            do
+            {
+                var bytes = await stream.ReadAsync(response, 0, TcpClient.ReceiveBufferSize, cancellationToken);
```
acodrington, line 361: If the stream is longer than the receive buffer size, won't this do/while get stuck in a loop as the `offset` value is hardcoded to 0?

LukeButters: 0 is the offset into the `byte[] response`, the returned bytes is the count of how many bytes were read into `response`. The next read will write to the start of the buffer which is read from in the next line into the string builder.

acodrington: Ah OK, I thought `offset` was the 'start reading from here' point in the stream, but I guess the stream position is incremented automatically. No worries!

acodrington (approved):

source/Halibut/Transport/TcpConnectionFactory.cs
```diff
@@ -47,6 +43,27 @@ public IConnection EstablishNewConnection(ExchangeProtocolBuilder exchangeProtoc
...
+
+            var certificateValidator = new ClientCertificateValidator(serviceEndpoint);
+            var client = await CreateConnectedTcpClientAsync(serviceEndpoint, log, cancellationToken);
+            log.Write(EventType.Diagnostic, $"Connection established to {client.Client.RemoteEndPoint} for {serviceEndpoint.BaseUri}");
+
+            var stream = client.GetStream();
+
+            log.Write(EventType.SecurityNegotiation, "Performing TLS handshake");
+            var ssl = new SslStream(stream, false, certificateValidator.Validate, UserCertificateSelectionCallback);
+            await ssl.AuthenticateAsClientAsync(serviceEndpoint.BaseUri.Host, new X509Certificate2Collection(clientCertificate), SslProtocols.Tls | SslProtocols.Tls11 | SslProtocols.Tls12, false);
```
nathanwoctopusdeploy, line 59: No cancellation support?

source/Halibut/Transport/Proxy/HttpProxyClient.cs
```diff
@@ -287,6 +332,42 @@ void SendConnectionCommand(string host, int port)
...
         }
+        
+        async Task SendConnectionCommandAsync(string host, int port, CancellationToken cancellationToken)
+        {
+            var stream = TcpClient.GetStream();
+            var connectCmd = GetConnectCmd(host, port);
+            var request = Encoding.ASCII.GetBytes(connectCmd);
+
+            // send the connect request
+            await stream.WriteAsync(request, 0, request.Length, cancellationToken);
```
nathanwoctopusdeploy, line 343: Does this stream need to be wrapped with the yet-to-merge `TimeoutStream` so it can timeout and consistent cancellation?

source/Halibut/Transport/TcpConnectionFactory.cs
LukeButters: Not with net48, yes with net6 although the method signature is different.

nathanwoctopusdeploy: Should we "patch" net48 so we can support cancellation across both?

source/Halibut/Transport/Proxy/HttpProxyClient.cs
LukeButters: Yes does do we have a standard TODO for following this up?

I guess perhaps we should just never call `GetStream` and instead have some sort of `GetStreamWithTimeoutSupport` extension method and replace all `GetStream` calls with that.

source/Halibut/Transport/Proxy/HttpProxyClient.cs
```diff
@@ -344,6 +425,18 @@ void WaitForData(NetworkStream stream)
                     throw new ProxyException(string.Format("A timeout while waiting for the proxy server at {0} on port {1} to respond.", Utils.GetHost(TcpClient), Utils.GetPort(TcpClient)), true);
             }
         }
+        
+        async Task WaitForDataAsync(NetworkStream stream, CancellationToken cancellationToken)
```
nathanwoctopusdeploy, line 429: Could this become redundant with the `TimeoutStream`?

source/Halibut/Transport/Proxy/HttpProxyClient.cs
nathanwoctopusdeploy: Sounds good, I have a standard comment of  `// TODO - ASYNC ME UP!` + details if relavent

source/Halibut/Transport/Proxy/HttpProxyClient.cs
LukeButters: Yes I think so. I wont change this however since we are not changing behaviour.

source/Halibut/Transport/TcpConnectionFactory.cs
LukeButters: Did we have a standard way of patching these methods to look like they support cancellation.

LukeButters: I have currently added a `// TODO - ASYNC ME UP!` comment here. To ensure we don't forget about this.

nathanwoctopusdeploy: > Did we have a standard way of patching these methods to look like they support cancellation

Would this be a use case of the 2 tasks, the action task and a cancellation/timeout task, if the cancellation/timeout task wins, then close the stream?

LukeButters: WaitAsync doesn't exist either in net48

PR #368, opened 2023-08-01
Timeout Stream Wrapper

nathanwoctopusdeploy:
# Background

Adds a stream wrapper that provides consistent cancellation and timeout for Async Read and Async Write operations accross net48 and net6.0

# Results

`new TimeoutStreamWrapper(innerStream)`

**Exceptions Thrown on Timeout and Cancellation**

Timeout: `OperationCanceledException` The {ReadAsync|WriteAsync} operation timed out after {TimeSpan}.
Cancellation: `OperationCanceledException` The {ReadAsync|WriteAsync} operation was cancelled.

### net48
- AsyncRead and AsyncWrite *would not* timeout based on the ReadTimeout and WriteTimeout properties of the stream
- AsyncRead and AsyncWrite *would not* cancel after the operation was in progress (only if the cancellation token was cancelled when the methods were called)

### net6.0
- AsyncRead and AsyncWrite *would not* timeout based on the ReadTimeout and WriteTimeout properties of the stream
- AsyncRead and AsyncWrite *would* cancel based on the cancellation token being cancelled

### net48 and net6.0
- AsyncRead and AsyncWrite *will* timeout based on the ReadTimeout and WriteTimeout properties of the stream
- AsyncRead and AsyncWrite *will* cancel based on the cancellation token being cancelled

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/BaseTest.cs
+++ b/source/Halibut.Tests/BaseTest.cs
@@ -18,7 +18,7 @@ namespace Halibut.Tests
         {
             Logger = new SerilogLoggerBuilder().Build().ForContext(GetType());
             Logger.Information("Test started");
-            cancellationTokenSource = new CancellationTokenSource(TestTimeoutAttribute.TestTimeout() - (int) TimeSpan.FromSeconds(5).TotalMilliseconds);
+            cancellationTokenSource = new CancellationTokenSource(TestTimeoutAttribute.TestTimeoutInMilliseconds() - (int) TimeSpan.FromSeconds(5).TotalMilliseconds);
             CancellationToken = cancellationTokenSource.Token;
             CancellationToken.Register(() =>
             {
--- a/source/Halibut.Tests/Support/TestAttributes/TestTimeoutAttribute.cs
+++ b/source/Halibut.Tests/Support/TestAttributes/TestTimeoutAttribute.cs
@@ -6,11 +6,15 @@ namespace Halibut.Tests.Support.TestAttributes
 {
     public class TestTimeoutAttribute : TimeoutAttribute
     {
-        public TestTimeoutAttribute() : base(TestTimeout())
+        public TestTimeoutAttribute(int timeoutInSeconds) : base((int)TimeSpan.FromSeconds(timeoutInSeconds).TotalMilliseconds)
         {
         }
 
-        public static int TestTimeout()
+        public TestTimeoutAttribute() : base(TestTimeoutInMilliseconds())
+        {
+        }
+
+        public static int TestTimeoutInMilliseconds()
         {
             if (Debugger.IsAttached) return (int) TimeSpan.FromHours(1).TotalMilliseconds;
             return (int) TimeSpan.FromMinutes(6).TotalMilliseconds;
--- a/source/Halibut.Tests/Support/Try.cs
+++ b/source/Halibut.Tests/Support/Try.cs
@@ -1,4 +1,6 @@
 using System;
+using System.Threading;
+using System.Threading.Tasks;
 
 namespace Halibut.Tests.Support
 {
@@ -14,6 +16,37 @@ namespace Halibut.Tests.Support
             {
                 onFailure(e);
             }        
+        }
+
+        public static async Task<Exception?> CatchingError(Func<Task> tryThisAction)
+        {
+            try
+            {
+                await tryThisAction();
+            }
+            catch (Exception e)
+            {
+                return e;
+            }
+
+            return null;
+        }
+
+        public static async Task<Exception?> RunTillExceptionOrCancellation(Func<Task> action, CancellationToken cancellationToken)
+        {
+            Exception? actualException = null;
+
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                actualException = await Try.CatchingError(async () => await action());
+
+                if (actualException != null)
+                {
+                    break;
+                }
+            }
+
+            return actualException;
         }
     }
 }
--- a/source/Halibut.Tests/Transport/Streams/TimeoutStreamFixture.cs
+++ b/source/Halibut.Tests/Transport/Streams/TimeoutStreamFixture.cs
@@ -1,0 +1,313 @@
+﻿using System;
+using System.Diagnostics;
+using System.IO;
+using System.Net;
+using System.Net.Sockets;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Halibut.Tests.Support;
+using Halibut.Tests.Support.TestAttributes;
+using Halibut.Transport.Streams;
+using NUnit.Framework;
+
+namespace Halibut.Tests.Transport.Streams
+{
+    [TestTimeout(timeoutInSeconds: 20)]
+    public class TimeoutStreamFixture : BaseTest
+    {
+        [Test]
+        public async Task ReadShouldPassThrough()
+        {
+            var (disposables, sut, performListenerWrite) = await BuildTcpClientAndTcpListener(CancellationToken);
+
+            using (disposables)
+            {
+                await performListenerWrite("Test");
+
+                var buffer = new byte[19];
+                var readBytes = sut.Read(buffer, 0, 19);
+                var readData = Encoding.UTF8.GetString(buffer, 0, readBytes);
+                
+                Assert.AreEqual("Test", readData);
+            }
+        }
+        
+        [Test]
+        public async Task ReadAsyncShouldPassThrough()
+        {
+            var (disposables, sut, performListenerWrite) = await BuildTcpClientAndTcpListener(CancellationToken);
+
+            using (disposables)
+            {
+                await performListenerWrite("Test");
+
+                var buffer = new byte[19];
+                var readBytes = await sut.ReadAsync(buffer, 0, 19, CancellationToken);
+                var readData = Encoding.UTF8.GetString(buffer, 0, readBytes);
+                
+                Assert.AreEqual("Test", readData);
+            }
+        }
+
+        [Test]
+        public async Task ReadAsyncShouldCancel()
+        {
+            using (var readTokenSource = new CancellationTokenSource(TimeSpan.FromSeconds(5)))
+            {
+                var (disposables, sut, _) = await BuildTcpClientAndTcpListener(CancellationToken);
+
+                using (disposables)
+                {
+                    // Ensure the timeouts are not used
+                    sut.WriteTimeout = (int)TimeSpan.FromSeconds(120).TotalMilliseconds;
+                    sut.ReadTimeout = (int)TimeSpan.FromSeconds(120).TotalMilliseconds;
+                    
+                    var stopWatch = Stopwatch.StartNew();
+                    
+                    var actualException = await Try.CatchingError(async () => await sut.ReadAsync(new byte[19], 0, 19, readTokenSource.Token));
+                    
+                    stopWatch.Stop();
+
+                    actualException.Should().NotBeNull().And.BeOfType<OperationCanceledException>();
+                    actualException!.Message.Should().Be("The ReadAsync operation was cancelled.");
+
+                    stopWatch.Elapsed.Should().BeLessThan(TimeSpan.FromSeconds(10));
+                }
+            }
+        }
+
+        [Test]
+        public async Task ReadAsyncShouldTimeout()
+        {
+            var (disposables, sut, _) = await BuildTcpClientAndTcpListener(CancellationToken);
+
+            using (disposables)
+            {
+                // Ensure the correct timeout is used
+                sut.WriteTimeout = (int)TimeSpan.FromSeconds(120).TotalMilliseconds;
+                sut.ReadTimeout = (int)TimeSpan.FromSeconds(5).TotalMilliseconds;
+
+                var stopWatch = Stopwatch.StartNew();
+
+                var actualException = await Try.CatchingError(async () => await sut.ReadAsync(new byte[19], 0, 19, CancellationToken));
+                
+                stopWatch.Stop();
+
+                actualException.Should().NotBeNull().And.BeOfType<OperationCanceledException>();
+                actualException!.Message.Should().Be("The ReadAsync operation timed out after 00:00:05.");
+
+                stopWatch.Elapsed.Should().BeLessThan(TimeSpan.FromSeconds(10));
+            }
+        }
+
+        [Test]
+        public async Task WriteShouldPassThrough()
+        {
+            string? readData = null;
+            
+            var (disposables, sut, _) = await BuildTcpClientAndTcpListener(
+                CancellationToken,
+                onListenerRead: async data =>
+                {
+                    await Task.CompletedTask;
+                    readData += data;
+                });
+
+            using (disposables)
+            {
+                var buffer = Encoding.UTF8.GetBytes("Test");
+                sut.Write(buffer, 0, buffer.Length);
+
+                while ((readData?.Length ?? 0) < 4 && !CancellationToken.IsCancellationRequested)
+                {
+                    await Task.Delay(10, CancellationToken);
+                }
+
+                Assert.AreEqual("Test", readData);
+            }
+        }
+        
+        [Test]
+        public async Task WriteAsyncShouldPassThrough()
+        {
+            string? readData = null;
+            
+            var (disposables, sut, _) = await BuildTcpClientAndTcpListener(
+                CancellationToken,
+                onListenerRead: async data =>
+                {
+                    await Task.CompletedTask;
+                    readData += data;
+                });
+
+            using (disposables)
+            {
+                var buffer = Encoding.UTF8.GetBytes("Test");
+                await sut.WriteAsync(buffer, 0, buffer.Length, CancellationToken);
+
+                while ((readData?.Length ?? 0) < 4 && !CancellationToken.IsCancellationRequested)
+                {
+                    await Task.Delay(10, CancellationToken);
+                }
+
+                Assert.AreEqual("Test", readData);
+            }
+        }
+
+        [Test]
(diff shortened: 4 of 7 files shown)


source/Halibut.Tests/Transport/Streams/TimeoutStreamWrapperFixture.cs
```diff
@@ -0,0 +1,369 @@
...
+                {
+                    // Ensure the timeouts are not used
+                    sut.WriteTimeout = (int)TimeSpan.FromSeconds(120).TotalMilliseconds;
+                    sut.ReadTimeout = (int)TimeSpan.FromSeconds(120).TotalMilliseconds;
+                    
+                    var stopWatch = Stopwatch.StartNew();
+                    Exception? actualException = null;
+
+                    // Brute force attempt to get the Write to be slow
+                    while (!CancellationToken.IsCancellationRequested)
```
nathanwoctopusdeploy, line 83: TODO - Is there a better way of making a write take a long time?

source/Halibut.Tests/Transport/Streams/TimeoutStreamWrapperFixture.cs
```diff
@@ -0,0 +1,369 @@
...
+
+                    var data = new byte[655360];
+                    var r = new Random();
+                    r.NextBytes(data);
+
+                    var stopWatch = Stopwatch.StartNew();
+                    Exception? actualException = null;
+
+                    // Brute force attempt to get the Write to be slow
+                    while (!CancellationToken.IsCancellationRequested)
```
nathanwoctopusdeploy, line 221: TODO - Is there a better way of making a write take a long time?

source/Halibut.Tests/Transport/Streams/TimeoutStreamWrapperFixture.cs
```diff
@@ -0,0 +1,369 @@
...
+
+                var data = new byte[655360];
+                var r = new Random();
+                r.NextBytes(data);
+
+                var stopWatch = Stopwatch.StartNew();
+                Exception? actualException = null;
+
+                // Brute force attempt to get the Write to be slow
+                while (!CancellationToken.IsCancellationRequested)
```
nathanwoctopusdeploy, line 265: TODO - Is there a better way of making a write take a long time?

source/Halibut/Transport/Streams/TimeoutStreamWrapper.cs
```diff
@@ -0,0 +1,178 @@
...
+            var timeout = ReadTimeout;
+            using var timeoutCancellationTokenSource = new CancellationTokenSource(timeout);
+            using var linkedCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, timeoutCancellationTokenSource.Token);
+            
+#if NETFRAMEWORK
+            using var registration = linkedCancellationTokenSource.Token.Register(() =>
+            {
+                try
+                {
+                    inner.Close();
```
nathanwoctopusdeploy, line 34: net48 🏆 😢

source/Halibut/Transport/Streams/TimeoutStreamWrapper.cs
```diff
@@ -0,0 +1,178 @@
...
+                {
+                    throw new OperationCanceledException("The WriteAsync operation was cancelled.", e);
+                }
+
+                throw new Exception("Here");
+            }
+        }
+
+#if !NETFRAMEWORK
+        public override async ValueTask<int> ReadAsync(Memory<byte> buffer, CancellationToken cancellationToken = default)
```
nathanwoctopusdeploy, line 98: These need to be implemented

nathanwoctopusdeploy:
[sc-55197]

source/Halibut/Transport/Streams/TimeoutStreamWrapper.cs
```diff
@@ -0,0 +1,192 @@
...
+        public override async Task FlushAsync(CancellationToken cancellationToken)
+        {
+            await WrapWithCancellationAndTimeout(
+                async ct =>
+                {
+                    await inner.FlushAsync(ct).ConfigureAwait(false);
+                    return 0;
+                },
+                CanTimeout ? WriteTimeout : int.MaxValue,
+                nameof(WriteAsync),
```
sburmanoctopus, line 29: FlushAsync?

source/Halibut/Transport/Streams/TimeoutStreamWrapper.cs
```diff
@@ -0,0 +1,192 @@
...
+        public override void Write(ReadOnlySpan<byte> buffer) => inner.Write(buffer);
+#endif
+
+#if NETFRAMEWORK
+        public override ObjRef CreateObjRef(Type requestedType) => inner.CreateObjRef(requestedType);
+
+        public override object? InitializeLifetimeService() => inner.InitializeLifetimeService();
+#endif
+
+        public override int ReadTimeout
```
sburmanoctopus, line 169: Just wondering, why did we go with an int instead of a TimeSpan?

If we stay with the int, did we want to rename this ReadTimeoutMilliseconds to be explicit?

source/Halibut.Tests/Transport/Streams/TimeoutStreamWrapperFixture.cs
```diff
@@ -0,0 +1,369 @@
...
+                read.Should().Throw<ObjectDisposedException>("Because the stream is closed");
+            }
+        }
+
+        async Task<(IDisposable Disposables, Stream SystemUnderTest, Func<string, Task> PerformServiceWriteFunc)> BuildTcpClientAndTcpListener(
+            CancellationToken cancellationToken, 
+            Func<string, Task>? onListenerRead = null)
+        {
+            Func<string, Task>? performServiceWriteFunc = null;
+            var disposableCollection = new DisposableCollection();
```
sburmanoctopus, line 320: We run a small risk of not disposing things we add to this list if an exception happens half way through this method.

I wonder if we should pass the `disposableCollection` as a parameter instead, and declare it in the calling method with 'using' as a way to guarantee disposal?

sburmanoctopus: Or another idea, could we make this a disposable class instead?

That could clean up some of the lambdas and tuples.

Maybe. Just a thought 🙂

source/Halibut.Tests/Transport/Streams/TimeoutStreamWrapperFixture.cs
```diff
@@ -0,0 +1,369 @@
...
+            
+            var (disposables, sut, performListenerWrite) = await BuildTcpClientAndTcpListener(CancellationToken);
+
+            using (disposables)
+            {
+                await performListenerWrite("Test");
+
+                var buffer = new byte[19];
+                var readBytes = sut.Read(buffer, 0, 19);
+                if (readBytes > 0)
```
sburmanoctopus, line 33: Should we remove the `if`?

This test will fail if it does not read anything, so I'm wondering if it's worth guarding for.

sburmanoctopus: Or if it is important, should it be in a while loop until it reads everything?

source/Halibut/Transport/Streams/TimeoutStreamWrapper.cs
```diff
@@ -0,0 +1,192 @@
...
+        {
+            this.inner = inner;
+        }
+
+        public override async Task FlushAsync(CancellationToken cancellationToken)
+        {
+            await WrapWithCancellationAndTimeout(
+                async ct =>
+                {
+                    await inner.FlushAsync(ct).ConfigureAwait(false);
```
LukeButters, line 25: Still doing `ConfigureAwait`?

source/Halibut.Tests/Transport/Streams/TimeoutStreamWrapperFixture.cs
```diff
@@ -0,0 +1,369 @@
...
+        }
+
+        [Test]
+        public async Task ReadAsyncShouldCancel()
+        {
+            using (var readTokenSource = new CancellationTokenSource(TimeSpan.FromSeconds(5)))
+            {
+                var (disposables, sut, _) = await BuildTcpClientAndTcpListener(
+                    CancellationToken,
+                    onListenerRead: async _ => await Task.Delay(TimeSpan.FromDays(1), CancellationToken));
```
sburmanoctopus, line 71: This lambda gets called straight after we "Read". And will effectively wait until the test finishes (because it waits a day, and uses the test's CancellationToken).

This stops the 'read' loop.

I wonder if it's worth a comment maybe? It wasn't obvious to me when I first looked at it (might be a reflection on me maybe 😁 )

source/Halibut.Tests/Transport/Streams/TimeoutStreamWrapperFixture.cs
```diff
@@ -0,0 +1,369 @@
...
+                using (disposables)
+                {
+                    // Ensure the timeouts are not used
+                    sut.WriteTimeout = (int)TimeSpan.FromSeconds(120).TotalMilliseconds;
+                    sut.ReadTimeout = (int)TimeSpan.FromSeconds(120).TotalMilliseconds;
+                    
+                    var stopWatch = Stopwatch.StartNew();
+                    Exception? actualException = null;
+
+                    // Brute force attempt to get the Write to be slow
```
sburmanoctopus, line 82: This comment is used below for both writes and reads.

In this case, it looks like it's to make 'Reads' slow?

source/Halibut.Tests/Transport/Streams/TimeoutStreamWrapperFixture.cs
```diff
@@ -0,0 +1,369 @@
...
+            using (disposables)
+            {
+                // Ensure the correct timeout is used
+                sut.WriteTimeout = (int)TimeSpan.FromSeconds(120).TotalMilliseconds;
+                sut.ReadTimeout = (int)TimeSpan.FromSeconds(5).TotalMilliseconds;
+
+                var stopWatch = Stopwatch.StartNew();
+                Exception? actualException = null;
+
+                // Brute force attempt to get the Write to be slow
```
sburmanoctopus, line 121: Nit: Seeing this pattern a few times, I wonder if it might be nicer to have a method like `RunTillException` that takes a lambda and CancellationToken, and returns the exception. Or even takes a generic argument of the expected exception type. 

Then we could use it for reads and writes.

Just a thought.

source/Halibut.Tests/Transport/Streams/TimeoutStreamWrapperFixture.cs
sburmanoctopus: We could even make a method that does this, and call it `DelayForADayToDelayReading` or something like that (and use it in the other tests as well). Just a thought.

source/Halibut.Tests/Transport/Streams/TimeoutStreamWrapperFixture.cs
```diff
@@ -0,0 +1,369 @@
...
+                    await Task.CompletedTask;
+                    readData += data;
+                });
+
+            using (disposables)
+            {
+                var buffer = Encoding.UTF8.GetBytes("Test");
+                sut.Write(buffer, 0, buffer.Length);
+
+                while (readData == null && !CancellationToken.IsCancellationRequested)
```
sburmanoctopus, line 162: Is it guaranteed to read all the data in one block?

If so, should the onListenerRead be `readData = data`?

If not, then would this test not be a risk of being flaky with the while loop only checking for `readData == null`?

source/Halibut/Transport/Streams/TimeoutStreamWrapper.cs
```diff
@@ -0,0 +1,192 @@
...
+using System;
+using System.IO;
+using System.Runtime.Remoting;
+using System.Threading;
+using System.Threading.Tasks;
+using Halibut.Util;
+
+namespace Halibut.Transport.Streams
+{
+    class TimeoutStreamWrapper : Stream
```
sburmanoctopus, line 11: TimeoutStreamWrapper or just TimeoutStream?

Just wondering if the word 'Wrapper' is redundant (as streams often wrap other streams)

source/Halibut/Transport/Streams/TimeoutStreamWrapper.cs
```diff
@@ -0,0 +1,192 @@
...
+            var actionTask = action(linkedCancellationTokenSource.Token);
+            var cancellationTask = linkedCancellationTokenSource.Token.AsTask<T>();
+
+            try
+            {
+                var completedTask = await Task.WhenAny(actionTask, cancellationTask).ConfigureAwait(false);
+
+                if (completedTask == cancellationTask)
+                {
+                    actionTask.IgnoreUnobservedExceptions();
```
sburmanoctopus, line 93: Is there a way we can get `actionTask` to die if we get here? I can't think of a nice way.

Maybe if we create another task using Task.Run, and pass in the cancellation token to that. Would that kill the process running the actual thread from `action`?

source/Halibut/Transport/Streams/TimeoutStreamWrapper.cs
nathanwoctopusdeploy: This overrides the Stream properties, out of our control.

sburmanoctopus (approved):
LGTM. It's annoying we have to do this, but we need this functionality.

Just some nits scattered throughout the PR. Otherwise, all good.

source/Halibut/Transport/Streams/TimeoutStreamWrapper.cs
nathanwoctopusdeploy: No you can't make actionTask stop apart from making sure it cancels co-operatively!

source/Halibut.Tests/Transport/Streams/TimeoutStreamWrapperFixture.cs
nathanwoctopusdeploy: Yep it's not needed any more 👍

source/Halibut/Transport/Streams/TimeoutStreamWrapper.cs
nathanwoctopusdeploy: Removed

source/Halibut.Tests/Transport/Streams/TimeoutStreamWrapperFixture.cs
nathanwoctopusdeploy: Nice to have, it could leak, i could do it better, it's a test, it's ok'ish

source/Halibut.Tests/Transport/Streams/TimeoutStreamWrapperFixture.cs
nathanwoctopusdeploy: Good spot, the reads should not have been wrapped in the while loop as we can predictably delay read (we just don't write anything). The writes are the bit we are struggling to predictably delay

source/Halibut.Tests/Transport/Streams/TimeoutStreamWrapperFixture.cs
nathanwoctopusdeploy: We have a Try.CatchingError helper that I will adapt to async and use here to remove boilerplate

source/Halibut/Transport/Streams/TimeoutStreamWrapper.cs
nathanwoctopusdeploy: Changed to TimeoutStream

source/Halibut.Tests/Transport/Streams/TimeoutStreamWrapperFixture.cs
nathanwoctopusdeploy: We have a Try.CatchingError helper that I will adapt to async and use here to remove boilerplate

source/Halibut.Tests/Transport/Streams/TimeoutStreamWrapperFixture.cs
nathanwoctopusdeploy: I don't know if its guaranteed so I've made it more robust in terms of the length of data it's waiting for

source/Halibut.Tests/Transport/Streams/TimeoutStreamWrapperFixture.cs
nathanwoctopusdeploy: Good spot, this was actually confusing as it was used in the read tests when it was not needed. It's only relevant for the write tests and is an attempt to delay one of the write attempts for timeout and cancellation testing. Cleaned up and `DelayForeverToTryAndDelayWriting` added for clarity of the intent of the delay

source/Halibut/Transport/Streams/TimeoutStreamWrapper.cs
sburmanoctopus: Ah, of course 🤦

LukeButters (approved):

source/Halibut/Transport/Streams/TimeoutStreamWrapper.cs
```diff
@@ -0,0 +1,192 @@
...
+
+        public override async Task FlushAsync(CancellationToken cancellationToken)
+        {
+            await WrapWithCancellationAndTimeout(
+                async ct =>
+                {
+                    await inner.FlushAsync(ct).ConfigureAwait(false);
+                    return 0;
+                },
+                CanTimeout ? WriteTimeout : int.MaxValue,
```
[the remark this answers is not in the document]
nathanwoctopusdeploy: It could 😉

PR #374, opened 2023-08-03
Add HalibutRuntime.HandleMessageAsync

acodrington:
Add an Async overload of `HalibutRuntime.HandleMessage`, as per [sc-53211].

## How to review
I converted the usages of `ExchangeActionAsync` in the two SecureListener classes to use `ExchangeActionWithCancellationAsync` instead. The SecureListener classes can then call the exchangeAction with its CancellationToken.

When HalibutRuntime goes down the async path, the CancellationToken can then be passed to the async method from `MessageExchangeProtocol`. When it goes down the sync path, the token is ignored and nothing changes.

I _think_ this makes sense, but there's a lot of passing around tokens etc, so please confirm. :)

### DIFF
--- a/source/Halibut/Transport/SecureListener.cs
+++ b/source/Halibut/Transport/SecureListener.cs
@@ -359,7 +359,7 @@ namespace Halibut.Transport
         {
             log.Write(EventType.Diagnostic, "Begin message exchange");
 
-            return exchangeAction(exchangeProtocolBuilder(stream, log));
+            return exchangeAction(exchangeProtocolBuilder(stream, log), cts.Token);
         }
 
         bool AcceptAnySslCertificate(object sender, X509Certificate clientCertificate, X509Chain chain, SslPolicyErrors sslpolicyerrors)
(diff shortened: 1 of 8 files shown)


source/Halibut/Transport/SecureListener.cs
```diff
@@ -40,22 +40,22 @@ public class SecureListener : IDisposable
         readonly Func<Dictionary<string, string>> getFriendlyHtmlPageHeaders;
         readonly CancellationTokenSource cts = new CancellationTokenSource();
         readonly TcpClientManager tcpClientManager = new TcpClientManager();
-        readonly ExchangeActionAsync exchangeAction;
```
nathanwoctopusdeploy, line 43: Is `ExchangeActionAsync` unused now or is it still used in other places?

If it's unused then `ExchangeActionWithCancellationAsync` -> `ExchangeActionAsync`

nathanwoctopusdeploy (approved):
Looks good, jsut the ExchangeActionWithCancellationAsync -> ExchangeActionAsync cahnge if `ExchangeActionAsync` is now unused.

Don't merge into the branch this is based off. Wait until that merges and then rebase onto main

source/Halibut/Transport/SecureListener.cs
acodrington: Yep, it's unused now. I will remove and rename as appropriate. 👍🏻

PR #376, opened 2023-08-03
Putting in greater measures to fix flaky test

sburmanoctopus:
# Background

Despite a [previous attempt](https://github.com/OctopusDeploy/Halibut/pull/371), the `QueueAndWait_WhenRequestIsDequeued_ButPollingRequestQueueTimeoutIsReached_ShouldWaitTillRequestRespondsAndClearRequest` test is still flaky.

Like before, the error was always Expected object not to be <null> coming from this line dequeued.Should().NotBeNull().And.Be(request);

# Results

## Before
The cause appears to still be the same. We cannot think of another reason why this would be happening.

The tricky part is that we have a separate task that does the queuing, and we have no control over when .NET schedules the task to run. In this case, it prioritises it to run completely before `DequeueAsync` can be called.

This is why we thought increasing the timeout would work, as it would encourage .NET to run the main execution path while it was waiting for the queuing task.

## After

Increasing the timeout didn't work. Since we cannot control which thread executes in which order (and that this test usually works), we decided instead to do a loop that waits until the correct condition is met.

It was noticed that the same problem may occur in the `QueueAndWait_WhenRequestIsDequeued_ButPollingRequestQueueTimeoutIsReached_AndPollingRequestMaximumMessageProcessingTimeoutIsReached_WillStopWaitingAndClearRequest` test. Although it's failure rate is lower, it has been known to fail.

So the same fix was applied there as well.

# How to review this PR
If you can think of a way to do this without a loop (or altering the `PendingRequestQueue` itself), then we would be very happy to try it.

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/ServiceModel/PendingRequestQueueFixture.cs
+++ b/source/Halibut.Tests/ServiceModel/PendingRequestQueueFixture.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


nathanwoctopusdeploy (approved):
LGTM, just a name change of the method as discussed so the reader knows this is forcing a particular scenario and not patching over potential problems

PR #377, opened 2023-08-03
Async JSON Serialization/Deserialization

sburmanoctopus:
[sc-53880]

# Background
As part of the async halibut work, we need to make JSON serialization/deserialization be async friendly as well.

# Results

## Before
Sadly, our JSON library does not support async 😞 

We explored several options ([outlined here](https://docs.google.com/document/d/1N8EwyAqhOA2FKGLcQK4tZMfbfPfZTnjFRtO7ox-VOAs/edit)), and found this to be the way to go balancing effort, maintainability, and general requirement.

## After
When writing, we write the compressed BSON to memory, and then asynchronously write it to the given network stream (as that's the blocker).

When reading, we read the decompressed BSON to memory asynchronously, then deserialize. Sadly, we cannot read the compressed data, as we do not know where the end of the stream is.

In both cases, we impose limits so we do not run out of memory. This can be tweaked from Octopus Server.

To help ensure we find the correct thresholds, the observability was enhanced to also record how much data is being written into/read from memory.

### StreamCapture
Several tests started failing as master had hooked up async, causing this code to be exercised. This uncovered a bug.

The `StreamCapture` class was using `CallContext`, which does not support async code (as it works on the current thread).

It was found that the existing code for coping with "not HAS_ASYNC_LOCAL" was very old, and no longer needed for net4.8.

Removing this old code caused the failing tests to pass.

This ended up being the last usage of `HAS_ASYNC_LOCAL`, so I removed it from the project definitions as well.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Transport/Streams/ReadIntoMemoryBufferStreamFixture.cs
+++ b/source/Halibut.Tests/Transport/Streams/ReadIntoMemoryBufferStreamFixture.cs
@@ -136,7 +136,8 @@ namespace Halibut.Tests.Transport.Streams
                     return await sut.ReadAsync(readBuffer, offset, count, CancellationToken);
                 case StreamMethod.Sync:
                     return sut.Read(readBuffer, offset, count);
-                case StreamMethod.BeginEnd:
+                case StreamMethod.LegacyAsync:
+                    // This is the way async reading was done in earlier version of .NET
                     int bytesRead = -1;
                     sut.BeginRead(readBuffer, offset, count, AsyncCallback, sut);
                     void AsyncCallback(IAsyncResult result)
--- a/source/Halibut.Tests/Transport/Streams/ReadIntoMemoryBufferStreamFixture.cs
+++ b/source/Halibut.Tests/Transport/Streams/ReadIntoMemoryBufferStreamFixture.cs
@@ -93,8 +93,7 @@ namespace Halibut.Tests.Transport.Streams
             readBuffer.Should().BeEquivalentTo(bytesToWrite);
             sut.BytesReadIntoMemory.Should().Be(bytesToWrite.Length - 1);
         }
-
-
+        
         [Test]
         public async Task PartiallyReadsFromMemoryBuffer_IfBufferingWasApplied_ReadingOneByteAtATime([Values] StreamMethod streamMethod)
         {
@@ -138,14 +137,17 @@ namespace Halibut.Tests.Transport.Streams
                 case StreamMethod.Sync:
                     return sut.Read(readBuffer, offset, count);
                 case StreamMethod.BeginEnd:
-                    int bytesRead = 0;
-                    var beginReadResult = sut.BeginRead(readBuffer, offset, count, AsyncCallback, sut);
+                    int bytesRead = -1;
+                    sut.BeginRead(readBuffer, offset, count, AsyncCallback, sut);
                     void AsyncCallback(IAsyncResult result)
                     {
                         bytesRead = sut.EndRead(result);
                     }
 
-                    beginReadResult.AsyncWaitHandle.WaitOne().Should().BeTrue();
+                    while (bytesRead < 0 && !CancellationToken.IsCancellationRequested)
+                    {
+                        await Task.Delay(10);
+                    }
                     return bytesRead;
                 default:
                     throw new ArgumentOutOfRangeException(nameof(streamMethod), streamMethod, null);
--- a/source/Halibut.Tests/Transport/Streams/ReadIntoMemoryBufferStreamFixture.cs
+++ b/source/Halibut.Tests/Transport/Streams/ReadIntoMemoryBufferStreamFixture.cs
@@ -1,0 +1,155 @@
+﻿using System;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Halibut.Transport.Observability;
+using Halibut.Transport.Streams;
+using NUnit.Framework;
+
+namespace Halibut.Tests.Transport.Streams
+{
+    public class ReadIntoMemoryBufferStreamFixture : BaseTest
+    {
+        [Test]
+        public async Task ReadsFromSourceStream_IfBufferingWasNotApplied([Values]StreamMethod streamMethod)
+        {
+            // Arrange
+            var bytesToWrite = Encoding.ASCII.GetBytes("Some bytes for testing");
+            using var memoryStream = CreateMemoryStreamPopulatedWith(bytesToWrite);
+            using var sut = new ReadIntoMemoryBufferStream(memoryStream, 1000, OnDispose.LeaveInputStreamOpen);
+
+            // Act
+            var readBuffer = new byte[bytesToWrite.Length];
+            var bytesRead = await ReadFromStream(streamMethod, sut, readBuffer, 0, bytesToWrite.Length);
+            
+            // Assert
+            bytesRead.Should().Be(bytesToWrite.Length);
+            readBuffer.Should().BeEquivalentTo(bytesToWrite);
+            sut.BytesReadIntoMemory.Should().Be(0);
+        }
+        
+        [Test]
+        public async Task ReadsFromMemoryBuffer_IfBufferingWasApplied_WithLimitEqualToDataSize([Values] StreamMethod streamMethod)
+        {
+            // Arrange
+            var bytesToWrite = Encoding.ASCII.GetBytes("Some bytes for testing");
+            using var memoryStream = CreateMemoryStreamPopulatedWith(bytesToWrite);
+            using var sut = new ReadIntoMemoryBufferStream(memoryStream, bytesToWrite.Length, OnDispose.LeaveInputStreamOpen);
+
+            await sut.BufferIntoMemoryFromSourceStreamUntilLimitReached(CancellationToken);
+
+            // Act
+            var readBuffer = new byte[bytesToWrite.Length];
+            var bytesRead = await ReadFromStream(streamMethod, sut, readBuffer, 0, bytesToWrite.Length);
+
+            // Assert
+            bytesRead.Should().Be(bytesToWrite.Length);
+            readBuffer.Should().BeEquivalentTo(bytesToWrite);
+            sut.BytesReadIntoMemory.Should().Be(bytesToWrite.Length);
+        }
+
+        [Test]
+        public async Task ReadsFromMemoryBuffer_IfBufferingWasApplied_WithLimitGreaterThanDataSize([Values] StreamMethod streamMethod)
+        {
+            // Arrange
+            var bytesToWrite = Encoding.ASCII.GetBytes("Some bytes for testing");
+            using var memoryStream = CreateMemoryStreamPopulatedWith(bytesToWrite);
+            using var sut = new ReadIntoMemoryBufferStream(memoryStream, bytesToWrite.Length + 1, OnDispose.LeaveInputStreamOpen);
+
+            await sut.BufferIntoMemoryFromSourceStreamUntilLimitReached(CancellationToken);
+
+            // Act
+            var readBuffer = new byte[bytesToWrite.Length];
+            var bytesRead = await ReadFromStream(streamMethod, sut, readBuffer, 0, bytesToWrite.Length);
+
+            // Assert
+            bytesRead.Should().Be(bytesToWrite.Length);
+            readBuffer.Should().BeEquivalentTo(bytesToWrite);
+            sut.BytesReadIntoMemory.Should().Be(bytesToWrite.Length);
+        }
+
+        [Test]
+        public async Task PartiallyReadsFromMemoryBuffer_IfBufferingWasApplied_WithLimitLessThanDataSize([Values] StreamMethod streamMethod)
+        {
+            // Arrange
+            var bytesToWrite = Encoding.ASCII.GetBytes("Some bytes for testing");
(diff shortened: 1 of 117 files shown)


source/Halibut.Tests/Transport/Streams/ReadIntoMemoryBufferStreamFixture.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        async Task<int> ReadFromStream(StreamMethod streamMethod, ReadIntoMemoryBufferStream sut, byte[] readBuffer, int offset, int count)
+        {
+            switch (streamMethod)
+            {
+                case StreamMethod.Async:
+                    return await sut.ReadAsync(readBuffer, offset, count, CancellationToken);
+                case StreamMethod.Sync:
+                    return sut.Read(readBuffer, offset, count);
+                case StreamMethod.BeginEnd:
```
LukeButters, line 139: I don't exactly understand what is happening here, could a comment be added and/or the enum give more detail on what is happening.

source/Halibut/Transport/Protocol/MessageSerializer.cs
```diff
@@ -61,17 +67,28 @@ public void WriteMessage<T>(Stream stream, T message)
...
+            using var compressedInMemoryBuffer = new WriteIntoMemoryBufferStream(compressedByteCountingStream, writeIntoMemoryLimitBytes, OnDispose.LeaveInputStreamOpen);
 
-            WriteMessage(stream, message);
-        }
+            using (var zip = new DeflateStream(compressedInMemoryBuffer, CompressionMode.Compress, true))
+            using (var bson = new BsonDataWriter(zip) { CloseOutput = false })
+            {
+                // for the moment this MUST be object so that the $type property is included
+                // If it is not, then an old receiver (eg, old tentacle) will not be able to understand messages from a new sender (server)
+                // Once ALL sources and targets are deserializing to MessageEnvelope<T>, (ReadBsonMessage) then this can be changed to T
```
LukeButters, line 83: This comment I am sure is misleading, we send a MessageEnvelope type over the wire. So changing that would break compatibility. I know its pre-existing but it could just be deleted.

source/Halibut.Tests/Transport/Streams/ReadIntoMemoryBufferStreamFixture.cs
sburmanoctopus: This is the legacy way .net used to do async.

I tested this as it was still a supported way to 'read' from a stream.

To help with understanding, I added a comment and renamed the enum to `LegacyAsync`

source/Halibut/Transport/Protocol/MessageSerializer.cs
```diff
@@ -61,17 +67,28 @@ public void WriteMessage<T>(Stream stream, T message)
...
-            observer.MessageWritten(compressedByteCountingStream.BytesWritten);
+            observer.MessageWritten(compressedByteCountingStream.BytesWritten, 0);
         }
 
         public async Task WriteMessageAsync<T>(Stream stream, T message, CancellationToken cancellationToken)
         {
-            // TODO - ASYNC ME UP!
-            await Task.CompletedTask;
+            using var compressedByteCountingStream = new ByteCountingStream(stream, OnDispose.LeaveInputStreamOpen);
+            using var compressedInMemoryBuffer = new WriteIntoMemoryBufferStream(compressedByteCountingStream, writeIntoMemoryLimitBytes, OnDispose.LeaveInputStreamOpen);
```
LukeButters, line 76: Could this just be a [BufferedStream](https://learn.microsoft.com/en-us/dotnet/api/system.io.bufferedstream?view=net-7.0)?

source/Halibut/Transport/Protocol/MessageSerializer.cs
sburmanoctopus: I don't have much context on the validity of this comment (hence leaving it 😁 ).

If you're sure we can delete, I'm happy to do so.

Delete it?

source/Halibut/Transport/Streams/WriteIntoMemoryBufferStream.cs
```diff
@@ -0,0 +1,123 @@
...
+                // We tried our best, but will no longer fit in memory. Transition over to use the sinkStream.
+                memoryStream.Position = 0;
+                memoryStream.CopyTo(sinkStream);
+                usingSinkStream = true;
+            }
+
+            sinkStream.Write(buffer, offset, count);
+        }
+
+        public async Task WriteAnyUnwrittenDataToSinkStream(CancellationToken cancellationToken)
```
LukeButters, line 114: Maybe a name like `WriteCacheToUnderlyingStream`, writing to sink stream is confusing since at this point sink/async doesn't mean much.

source/Halibut/Transport/Streams/WriteIntoMemoryBufferStream.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Halibut.Transport.Observability;
+
+namespace Halibut.Transport.Streams
+{
+    public class WriteIntoMemoryBufferStream : Stream
+    {
+        readonly MemoryStream memoryStream;
+        readonly Stream sinkStream;
```
LukeButters, line 12: Rename to `inner` because all I read is "stinkStream"

source/Halibut/Transport/Streams/WriteIntoMemoryBufferStream.cs
```diff
@@ -0,0 +1,123 @@
...
+        public override bool CanRead => false;
+        public override bool CanWrite => sinkStream.CanWrite;
+        public override bool CanSeek => false;
+        public override bool CanTimeout => false;
+
+        public override long Length => throw new NotSupportedException();
+
+        protected override void Dispose(bool disposing)
+        {
+            base.Dispose(disposing);
```
LukeButters, line 36: May as well write the rest of the cache if any to the underlying stream here as well, so no one can get it wrong.

source/Halibut/Transport/Protocol/MessageSerializer.cs
sburmanoctopus: It seems like it could.

Looking through the BufferedStream code, I think it would work, as it tends to write to the memory buffer until it reaches the buffer size, then it writes out to the underlying stream. 

But then once the buffer is written to the stream when buffer is full, it then clears and repeats. I'm not sure if this is what we want. I don't think it really matters though.

The BufferedStream does not expose how much was written through the buffer. So to be able to support `compressedInMemoryBuffer.BytesWrittenIntoMemory`, we'd have to calculate that by using the min of `compressedByteCountingStream.BytesWritten` and the cap.

The above is quite minor. The only thing making me pause on cutting over is that we lose control of how this works, and can't tweak as easily in future.

But if you think it worth doing, I can do it 🙂

source/Halibut/Transport/Streams/WriteIntoMemoryBufferStream.cs
```diff
@@ -0,0 +1,123 @@
...
+
+namespace Halibut.Transport.Streams
+{
+    public class WriteIntoMemoryBufferStream : Stream
+    {
+        readonly MemoryStream memoryStream;
+        readonly Stream sinkStream;
+        readonly long writeIntoMemoryLimitBytes;
+        readonly OnDispose onDispose;
+        bool usingSinkStream;
```
LukeButters, line 15: Rename to something that suggests buffering.

source/Halibut/Transport/Streams/WriteIntoMemoryBufferStream.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Halibut.Transport.Observability;
+
+namespace Halibut.Transport.Streams
+{
+    public class WriteIntoMemoryBufferStream : Stream
+    {
+        readonly MemoryStream memoryStream;
```
LukeButters, line 11: nit: could be called buffer since it is a buffer...

source/Halibut/Transport/Streams/WriteIntoMemoryBufferStream.cs
LukeButters: or rename to flush which take's a bool which controls if the underlying stream has flushed called on it.

source/Halibut/Transport/Streams/ReadIntoMemoryBufferStream.cs
```diff
@@ -0,0 +1,126 @@
...
+namespace Halibut.Transport.Streams
+{
+    public class ReadIntoMemoryBufferStream : Stream
+    {
+        readonly MemoryStream memoryStream;
+        readonly Stream sourceStream;
+        readonly long readIntoMemoryLimitBytes;
+        readonly OnDispose onDispose;
+
+        bool limitReached;
```
LukeButters, line 16: I deleted this field and the `ReadIntoMemoryBufferStreamFixture` passed.

is it needed?

source/Halibut/Transport/Streams/ReadIntoMemoryBufferStream.cs
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        public override long Seek(long offset, SeekOrigin origin) => throw new NotSupportedException();
+
+        public override void SetLength(long value) => throw new NotSupportedException();
+        public override void Write(byte[] buffer, int offset, int count) => throw new NotSupportedException();
+
+        public async Task BufferIntoMemoryFromSourceStreamUntilLimitReached(CancellationToken cancellationToken)
+        {
+            var buffer = new byte[81920];
```
LukeButters, line 107: why `81920`?

source/Halibut/Transport/Protocol/MessageSerializer.cs
LukeButters: I guess the difference is once the buffer is exceeded the current stream wont do any more buffering where BufferedStream will.

LukeButters (approved):
LGTM

source/Halibut/Transport/Protocol/MessageSerializer.cs
sburmanoctopus: For now, I'm inclined to leave it as is, and then we can play with it in subsequent PRs

source/Halibut/Transport/Streams/WriteIntoMemoryBufferStream.cs
sburmanoctopus: I went with `WriteBufferToUnderlyingStream`. I took your suggestion, but replaced `Cache` with `Buffer`, as that's the nomenclature used in this class.

I did think of using a method like `Flush` initially, but felt it was just a bit too different from what `Flush` actually does.

source/Halibut/Transport/Streams/WriteIntoMemoryBufferStream.cs
sburmanoctopus: Lol, sure. I was going for "source and sink", but am happy to rename 😁

source/Halibut/Transport/Streams/WriteIntoMemoryBufferStream.cs
sburmanoctopus: Sure. I wrote a test for this as well 😁

source/Halibut/Transport/Streams/WriteIntoMemoryBufferStream.cs
sburmanoctopus: Went with `usingMemoryBuffer`, and flipped the boolean.

source/Halibut/Transport/Streams/WriteIntoMemoryBufferStream.cs
sburmanoctopus: Sure.

source/Halibut/Transport/Streams/ReadIntoMemoryBufferStream.cs
sburmanoctopus: Huh, yeah, it isn't needed. Removed!

source/Halibut/Transport/Streams/ReadIntoMemoryBufferStream.cs
sburmanoctopus: This is the value of `Stream._DefaultCopyBufferSize` (which is not public 😑)

LukeButters: This might not be what we want since now we alloc 81k for each in flight request we are waiting for, bring it down to `8192` or what the buffered rewindable stream has this set to which is under HalibutLimits.

The rewindable stream will actually only fetch up to its buffer size, so it wont be any slower.

LukeButters: Wait no actually the rewindable buffer stream is on the compressed bytes but this will be looking at uncompressed bytes.

I think the memory concerns still hold, so dropping this buffer size down is probably going to reduce memory pressure without much or any performance.

sburmanoctopus: I moved this down to 8192 here, as well as in the `WriteIntoMemoryBufferStream`

PR #380, opened 2023-08-04
Make SecureListeners async

acodrington:
This PR converts any remaining sync code to async, in the following classes:
* `SecureListener`
* `SecureWebSocketListener`

Implemented as part of [sc-53211]


source/Halibut/Transport/SecureListener.cs
```diff
@@ -310,16 +315,32 @@ void SendFriendlyHtmlPage(Stream stream)
...
-                foreach (var header in headers)
-                    writer.WriteLine($"{header.Key}: {header.Value}");
-                writer.WriteLine();
-                writer.WriteLine(message);
-                writer.WriteLine();
-                writer.Flush();
-                stream.Flush();
+                if (asyncHalibutFeature.IsEnabled())
+                {
+                    await writer.WriteLineAsync("HTTP/1.0 200 OK");
```
nathanwoctopusdeploy, line 320: nit: We could write to the stream directly (see sync and async version of [this](https://github.com/OctopusDeploy/Halibut/blob/93b2f5c0252c816d1586d4b8ae08e2c2b8472c38/source/Halibut/Transport/Protocol/MessageExchangeStream.cs#L90) ) if we wanted to be able to support cancellation

nathanwoctopusdeploy (changes requested):
LGM. We should convert [these tests](https://github.com/OctopusDeploy/Halibut/blob/93b2f5c0252c816d1586d4b8ae08e2c2b8472c38/source/Halibut.Tests/FriendlyHtmlPageTests.cs#L19) so they run for sync and async and test all the changes to the friendly error page which you now have async logic for

source/Halibut.Tests/FriendlyHtmlPageTests.cs
```diff
@@ -34,7 +34,7 @@ static DelegateServiceFactory GetDelegateServiceFactory()
         public async Task SupportsHttpsGet(string uriFormat)
         {
             var services = GetDelegateServiceFactory();
-            using (var octopus = new HalibutRuntime(services, Certificates.Octopus))
+            using (var octopus = new HalibutRuntimeBuilder().WithServiceFactory(services).WithServerCertificate(Certificates.Octopus).WithAsyncHalibutFeatureEnabled().Build())
```
nathanwoctopusdeploy, line 37: Do these tests now only test Async and so now were missing sync coverage?

source/Halibut/Transport/SecureListener.cs
```diff
@@ -209,7 +214,7 @@ async Task ExecuteRequest(TcpClient client)
                     if (req.Substring(0, 2) != "MX")
                     {
                         log.Write(EventType.Diagnostic, "Appears to be a web browser, sending friendly HTML response");
-                        SendFriendlyHtmlPage(ssl);
+                        await SendFriendlyHtmlPage(ssl);
```
nathanwoctopusdeploy, line 217: nit/alt: you could have also approached this by leaving the existing methods alone and adding a new version e.g. `SendFriendlyHtmlPageAsync` and do this if..else.. here

That way you would not have to convert the existing method to async

source/Halibut/Transport/SecureListener.cs
```diff
@@ -301,25 +306,40 @@ void SafelyRemoveClientFromTcpClientManager(TcpClient client, EndPoint clientNam
...
 
-            // This could fail if the client terminates the connection and we attempt to write to it
-            // Disposing the StreamWriter will close the stream - it owns the stream
-            using (var writer = new StreamWriter(stream, new UTF8Encoding(false)) { NewLine = "\r\n" })
+            if (asyncHalibutFeature.IsEnabled())
             {
-                writer.WriteLine("HTTP/1.0 200 OK");
-                writer.WriteLine("Content-Type: text/html; charset=utf-8");
-                writer.WriteLine("Content-Length: " + message.Length);
+                await stream.WriteControlLineAsync("HTTP/1.0 200 OK", cts.Token);
```
nathanwoctopusdeploy, line 316: While `WriteControlLineAsync` has the desired behaviour, it looks strange being used here. The reader may wonder what's going on as this is not control lines, just HTML. It may be worth adding a new extension for sending this over the stream.

nathanwoctopusdeploy: E.g. just `WriteLine`. It will duplicate code but I think that's ok!!

source/Halibut/Transport/Streams/StreamExtensionMethods.cs
```diff
@@ -21,5 +21,14 @@ public static async Task WriteEmptyControlLineAsync(this Stream stream, Cancella
         {
             await stream.WriteControlLineAsync(string.Empty, cancellationToken);
         }
+
+        public static async Task<int> ReadByteAsync(this Stream stream, CancellationToken cancellationToken)
```
nathanwoctopusdeploy, line 25: 👍

source/Halibut/Transport/SecureListener.cs
acodrington: Given that `WriteControlLineAsync` doesn't have anything "control line"-specific in the implementation, can we just rename the methods to `WriteLineAsync` and `WriteEmptyLineAsync`?

Duplication is fine if you want to keep the specific "Control Line" naming though.

source/Halibut.Tests/FriendlyHtmlPageTests.cs
acodrington: 🤦🏻‍♂️ 

I've created some new test attributes to deal with sync/async when using `TestCase`, so we can define multiple `TestCase` attributes on a tests which will then get run for both sync and async.

source/Halibut/Transport/SecureListener.cs
acodrington: Yeah, it was a bit 6 of one, half dozen of the other tbh.

I figured if I change the existing method now, then we don't have to change it when we remove the sync code paths 🤷🏻‍♂️

source/Halibut/Transport/SecureListener.cs
nathanwoctopusdeploy: Leave `WriteControlLineAsync` as this can never change and just add a new extension that is WriteLineAsync etc.

source/Halibut/Transport/SecureListener.cs
nathanwoctopusdeploy: We wouldn't have to change the existing sync method when we remove sync, we would just remove it 😉

source/Halibut.Tests/Support/TestAttributes/FriendlyHtmlSyncAndAsyncTestCaseAttribute.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Collections;
+using Halibut.Tests.Support.TestCases;
+using NUnit.Framework;
+
+namespace Halibut.Tests.Support.TestAttributes
+{
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = true, Inherited = false)]
+    public class FriendlyHtmlSyncAndAsyncTestCasesAttribute : TestCaseSourceAttribute
+    {
+        public FriendlyHtmlSyncAndAsyncTestCasesAttribute(string html, string expected)
```
nathanwoctopusdeploy, line 11: Consider something more generic to avoid an attribute per test e.g.

```
[AttributeUsage(AttributeTargets.Method, AllowMultiple = true, Inherited = false)]
    public class SyncAndAsyncTestCaseAttribute : TestCaseSourceAttribute
    {
        public SyncAndAsyncTestCaseAttribute(params object[] parameters) :
            base(
                typeof(SyncAndAsyncTestCase),
                nameof(SyncAndAsyncTestCase.GetTestCases),
                new object []{ parameters.ToArray() })
        {
        }

        static class SyncAndAsyncTestCase
        {
            public static IEnumerable GetTestCases(object[] parameters)
            {
                var syncParameterList = new List<object>(parameters) { SyncOrAsync.Sync };
                var asyncParameterList = new List<object>(parameters) { SyncOrAsync.Async };

                yield return syncParameterList.ToArray();
                yield return asyncParameterList.ToArray();
            }
        }
    }
```

image

source/Halibut.Tests/FriendlyHtmlPageTests.cs
```diff
@@ -152,5 +152,16 @@ static async Task<string> DownloadStringIgnoringCertificateValidation(string uri
                 }
             }
         }
+
+        HalibutRuntime GetHalibutRuntime(IServiceFactory serviceFactory, SyncOrAsync syncOrAsync)
```
nathanwoctopusdeploy, line 156: nit: `IServiceFactory serviceFactory` is never customised, it's always `GetDelegateServiceFactory()` s this param could be removed to reduce plumbing code.

source/Halibut.Tests/Transport/SecureListenerFixture.cs
```diff
@@ -60,7 +61,8 @@ public void SecureListenerDoesNotCreateHundredsOfIoEventsPerSecondOnWindows()
                     null,
                     thumbprint => true,
                     new LogFactory(),
-                    () => ""
+                    () => "",
+                    AsyncHalibutFeature.Disabled
```
nathanwoctopusdeploy, line 65: Should this also test Async?

source/Halibut/Transport/Streams/StreamExtensionMethods.cs
```diff
@@ -16,5 +17,20 @@ public static async Task WriteControlLineAsync(this Stream stream, string s, Can
             var bytes = Encoding.GetBytes(s + ControlMessageNewLine);
             await stream.WriteAsync(bytes, 0, bytes.Length, cancellationToken);
         }
+        
+        public static async Task WriteLineAsync(this Stream stream, string s, CancellationToken cancellationToken)
+        {
+            var bytes = Encoding.GetBytes(s + ControlMessageNewLine);
```
nathanwoctopusdeploy, line 23: nit: ControlMessageNewLine doesn't make sense here.

The reason for having Control* stuff defined separately as this is an area that can never change e.g. for backward compatibility, we can never change the behaviour.  The thought was to leave these alone and not use them so it's clear that there is a specific contract for how control messages are sent over the stream.

nathanwoctopusdeploy (approved):
Looking good. Consider a more generic test case attribute to avoid having to write a test case source attribute per test.

source/Halibut.Tests/Support/TestAttributes/FriendlyHtmlSyncAndAsyncTestCaseAttribute.cs
acodrington: Ah perfect, this is what I wanted to do but I was stuck on the idea of needing a 'test case class' that wraps all the params i.e. `ClientAndServiceTestCase`. If I don't need that kind of class, and can instead use `params object[]` along with the assumption that the sync/async field is either the first or last param, then it becomes much simpler.

source/Halibut.Tests/FriendlyHtmlPageTests.cs
acodrington: I flip-flopped on whether to do this or not, ultimately deciding to leave it as-is. I will update and remove the unecessary plumbing.

source/Halibut.Tests/Transport/SecureListenerFixture.cs
acodrington: Yes, I will update.

source/Halibut/Transport/Streams/StreamExtensionMethods.cs
acodrington: Similar to the other nit, I flip-flopped on whether to use `Environment.NewLine` here instead, for the exact reasons you've stated. Ultimately I decided to leave it as I thought purposely _not using_ `ControlMessageNewLine` here,  instead replacing with `Environment.NewLine`, might be confusing. Too much second-guessing! 😵‍💫 

I will update.

nathanwoctopusdeploy: `Environment.NewLine` will change the behaviour depending on the OS hosting the Halibut Runtime. I would suggest maintaining the behaviour of the sync code that uses `using (var writer = new StreamWriter(stream, new UTF8Encoding(false)) { NewLine = "\r\n" })`

PR #382, opened 2023-08-04
Convert ConnectionManager to async

LukeButters:
# Background

So we can do more.

Rather than try to convert each `sync` method to `async` in place. I think it was just easier to adjust it slightly.

[SC-53211]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Transport/ConnectionManager.cs
+++ b/source/Halibut/Transport/ConnectionManager.cs
@@ -28,7 +28,6 @@ namespace Halibut.Transport
             var connection = await TryGetExistingConnectionAsync(exchangeProtocolBuilder, connectionFactory, serviceEndpoint, log, cancellationToken);
             if (connection != null) return connection;
             
-            // 
             return await GetConnectionByCreatingItAsync(exchangeProtocolBuilder, connectionFactory, serviceEndpoint, log, cancellationToken);
         }
 
--- a/source/Halibut/Transport/ConnectionManager.cs
+++ b/source/Halibut/Transport/ConnectionManager.cs
@@ -25,12 +25,11 @@ namespace Halibut.Transport
 
         public async Task<IConnection> AcquireConnectionAsync(ExchangeProtocolBuilder exchangeProtocolBuilder, IConnectionFactory connectionFactory, ServiceEndPoint serviceEndpoint, ILog log, CancellationToken cancellationToken)
         {
-            // TODO - ASYNC ME UP!
-            await Task.CompletedTask;
-
-#pragma warning disable CS0612
-            return AcquireConnection(exchangeProtocolBuilder, connectionFactory, serviceEndpoint, log, cancellationToken);
-#pragma warning restore CS0612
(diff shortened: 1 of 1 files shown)


source/Halibut/Transport/ConnectionManager.cs
```diff
@@ -25,12 +25,11 @@ public IConnection AcquireConnection(ExchangeProtocolBuilder exchangeProtocolBui
...
-            // TODO - ASYNC ME UP!
-            await Task.CompletedTask;
-
-#pragma warning disable CS0612
-            return AcquireConnection(exchangeProtocolBuilder, connectionFactory, serviceEndpoint, log, cancellationToken);
-#pragma warning restore CS0612
+            var connection = await TryGetExistingConnectionAsync(exchangeProtocolBuilder, connectionFactory, serviceEndpoint, log, cancellationToken);
+            if (connection != null) return connection;
+            
+            //
```
nathanwoctopusdeploy, line 31: nit: `//`

source/Halibut/Transport/ConnectionManager.cs
```diff
@@ -58,6 +80,12 @@ public async Task<IConnection> AcquireConnectionAsync(ExchangeProtocolBuilder ex
                 var c = lazyConnection.Value;
             });
         }
+        
+        async Task<IConnection> CreateNewConnectionWithIOAsync(ExchangeProtocolBuilder exchangeProtocolBuilder, IConnectionFactory connectionFactory, ServiceEndPoint serviceEndpoint, ILog log, CancellationToken cancellationToken)
```
LukeButters, line 84: IO is intentional since the other calls "Create" a connection but don't actually and do the IO in some other place.

source/Halibut/Transport/ConnectionManager.cs
```diff
@@ -47,6 +45,30 @@ public async Task<IConnection> AcquireConnectionAsync(ExchangeProtocolBuilder ex
                 return openableConnection;
             }
         }
+        
+        async Task<IConnection> TryGetExistingConnectionAsync(ExchangeProtocolBuilder exchangeProtocolBuilder, IConnectionFactory connectionFactory, ServiceEndPoint serviceEndpoint, ILog log, CancellationToken cancellationToken)
+        {
+            await Task.CompletedTask;
+            lock (activeConnections)
```
nathanwoctopusdeploy, line 52: Does this even work? a lock in an async!!

LukeButters, line 52: I don't think we need this to be async, since no IO should ever occur within this lock. That is the reason the complex Tuple logic exists.

LukeButters: hmm I had understood it would use a sync lock, which I am less concerned about since the lock shouldn't be held while doing IO.

We could change it but it does mean the sync code must also change.

nathanwoctopusdeploy: No ignore, you can't call async inside a lock. My misunderstanding here 👍

nathanwoctopusdeploy (approved):
LGTM

PR #398, opened 2023-08-07
Convert WhenCallingServicesSimilarToTheOnesInTentacle to use async code path

acodrington:
This PR converts the `WhenCallingServicesSimilarToTheOnesInTentacle` test class to exercise the async code paths, as part of [sc-54458]

NB: Further async-friendly improvements can be made in these tests once `DataStream` is converted to be async, but that is outside the scope of this PR.


LukeButters (approved):
The

PR #399, opened 2023-08-08
Convert ParallelRequestsFixture to run in sync & async

LukeButters:
# Background

[SC-54458]

The tests are written twice for sync and again for async.

The reason for this is if we try to use the async test proxy over a sync call we will cause thread pool exhaustion.
Also if we try to use threads in the 'async parallel tests' we are not actually testing what we want to test. Since it should be the case we can, without making threads, run lots of parallel requests.

Eventually when everything is actually async this async side of the test should be changed to show that hundreds of concurrent requests can be made. Since everything is async, we wont have thread pool exhaustion.

# How to review this PR

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/ParallelRequestsFixture.cs
+++ b/source/Halibut.Tests/ParallelRequestsFixture.cs
@@ -73,7 +73,7 @@ namespace Halibut.Tests
             }
 
             [Test]
-            [LatestAndPreviousClientAndServiceVersionsTestCases(testPolling: false, testWebSocket: false, testNetworkConditions: false)]
+            [LatestAndPreviousClientAndServiceVersionsTestCases(testPolling: false, testWebSocket: false, testNetworkConditions: false, testAsyncAndSyncClients: false)]
             public async Task SendMessagesToTentacleInParallelAsync(ClientAndServiceTestCase clientAndServiceTestCase)
             {
                 var builder = clientAndServiceTestCase.CreateTestCaseBuilder();
--- a/source/Halibut.Tests/ParallelRequestsFixture.cs
+++ b/source/Halibut.Tests/ParallelRequestsFixture.cs
@@ -10,142 +10,248 @@ using Halibut.Logging;
 using Halibut.Tests.Support;
 using Halibut.Tests.Support.TestAttributes;
 using Halibut.Tests.Support.TestCases;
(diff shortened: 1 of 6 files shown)


source/Halibut.Tests/ParallelRequestsFixture.cs
```diff
@@ -6,147 +6,266 @@
...
-            testAsyncAndSyncClients: false // TODO - ASYNC ME UP!
-            )]
-        public async Task MultipleRequestsCanBeInFlightInParallel(ClientAndServiceTestCase clientAndServiceTestCase)
+        public class AsyncParallelRequestFixture : BaseTest
         {
-            using var clientAndService = await clientAndServiceTestCase.CreateTestCaseBuilder()
-                .WithStandardServices()
-                .Build(CancellationToken);
+            [Test]
+            [LatestAndPreviousClientAndServiceVersionsTestCases(testPolling: false, testWebSocket: false, testNetworkConditions: false, testAsyncAndSyncClients: false)]
```
sburmanoctopus, line 23: There was a `// TODO - ASYNC ME UP!` before. That's gone. Does this mean `testAsyncAndSyncClients` should be `true` now? or is it still not ready?

LukeButters: Maybe more context on the PR description would have made sense.

The tests are written twice for sync and again for async.
The reason for this is if we try to use the async test proxy over a sync call we will cause thread pool exhaustion.
Also if we true to use threads in the async parallel tests we are not actually testing what we want to test since it should be the case we can without making threads run logs of parallel requests.

Eventually when everything is actually async this async side of the test should be changed to show that hundreds of concurrent requests can be made. Since everything is async, we wont have thread pool exhaustion.

sburmanoctopus: Ok, cool. 

So then, should `SendMessagesToTentacleInParallelAsync` also have `testAsyncAndSyncClients: false`?

LukeButters: yes

LukeButters: Done

sburmanoctopus (approved):
LGTM

PR #400, opened 2023-08-08
Making timeout stream more network related

sburmanoctopus:
# Background

Tests [were failing](https://build.octopushq.com/buildConfiguration/OctopusDeploy_Halibut_ChainFullChain/9106734) on master due to a poison merge.

# Results

## Before

These were happening because more tests were being turned on to use async, which caused the async JSON serialization paths to fire, which then caused the async timeout code paths to be hit.

The exceptions being thrown from the async timeouts did not align with the synchronous versions.

## After

We made the `TimeoutStream` throw the same type of exception that would be thrown when the same network timeout was thrown during synchronous code.

We renamed it to `NetworkTimeoutStream` to help make it obvious why it was here, and to validate the types of exceptions now being thrown from it.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
+++ b/source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
@@ -27,6 +27,7 @@ namespace Halibut.Transport.Streams
                     return 0;
                 },
                 CanTimeout ? WriteTimeout : int.MaxValue,
+                "flush",
                 nameof(FlushAsync),
                 cancellationToken);
         }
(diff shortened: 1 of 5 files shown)


source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
```diff
@@ -114,7 +115,8 @@ void ThrowMeaningfulException(Exception? innerException = null)
             {
                 if (timeoutCancellationTokenSource.IsCancellationRequested)
                 {
-                    throw new OperationCanceledException($"The {methodName} operation timed out after {TimeSpan.FromMilliseconds(timeout)}.", innerException);
+                    var socketException = new SocketException(10060);
```
sburmanoctopus, line 118: This magic number is what is thrown when the synchronous version is thrown. It also generates the same message for us 😁

nathanwoctopusdeploy (approved):

PR #401, opened 2023-08-08
Make DataStream and IDataSTreamReceiver  async.

LukeButters:
# Background

[SC-53211]

Makes DataStream IDataSTreamReceiver  async.

DataStream can not be given an async writer, which will be used in async modes.

DataStream has a new constructor which can take both a async/sync writers. The DataStream's current ctor which takes only a sync writer will continue work and will result in async code calling sync code. This has been marked as obsolete.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.TestUtils.CompatBinary.Base/Services/FileTransferService.cs
+++ b/source/Halibut.TestUtils.CompatBinary.Base/Services/FileTransferService.cs
@@ -2,7 +2,7 @@
 using System.IO;
 using Octopus.Tentacle.Contracts;
 
-namespace Halibut.TestUtils.Contracts.Tentacle.Services
+namespace Halibut.TestUtils.SampleProgram.Base.Services
 {
     public class FileTransferService : IFileTransferService
     {
--- a/source/Halibut.Tests/DataStreamFixture.cs
+++ b/source/Halibut.Tests/DataStreamFixture.cs
@@ -18,7 +18,7 @@ namespace Halibut.Tests
         [LatestClientAndLatestServiceTestCases(testNetworkConditions: false)]
         [LatestClientAndPreviousServiceVersionsTestCases(testNetworkConditions: false)]
         [Obsolete]
-        public async Task ASyncDataStreamWriterCanStillBeUsed(ClientAndServiceTestCase clientAndServiceTestCase)
+        public async Task SyncDataStreamWriterCanStillBeUsed(ClientAndServiceTestCase clientAndServiceTestCase)
         {
             using (var clientAndService = await clientAndServiceTestCase.CreateTestCaseBuilder()
                        .WithStandardServices()
--- a/source/Halibut.Tests/DataStreamFixture.cs
+++ b/source/Halibut.Tests/DataStreamFixture.cs
@@ -1,0 +1,104 @@
+﻿using System;
+using System.IO;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Halibut.Tests.Support;
+using Halibut.Tests.Support.TestAttributes;
+using Halibut.Tests.Support.TestCases;
+using Halibut.Tests.TestServices.Async;
+using Halibut.TestUtils.Contracts;
+using Halibut.Transport.Streams;
+using NUnit.Framework;
+
+namespace Halibut.Tests
+{
+    public class DataStreamFixture : BaseTest
+    {
+        [Test]
+        [LatestClientAndLatestServiceTestCases(testNetworkConditions: false)]
+        [LatestClientAndPreviousServiceVersionsTestCases(testNetworkConditions: false)]
+        [Obsolete]
+        public async Task ASyncDataStreamWriterCanStillBeUsed(ClientAndServiceTestCase clientAndServiceTestCase)
+        {
+            using (var clientAndService = await clientAndServiceTestCase.CreateTestCaseBuilder()
+                       .WithStandardServices()
+                       .Build(CancellationToken))
+            {
+                
+                var echo = clientAndService.CreateClient<IEchoService, IAsyncClientEchoService>();
+
+                var data = new byte[1337];
+                new Random().NextBytes(data);
+
+                for (var i = 0; i < clientAndServiceTestCase.RecommendedIterations; i++)
+                {
+                    // Note that the DataStream is only given a sync writer.
+                    var count = await echo.CountBytesAsync(new DataStream(data.Length, stream => stream.Write(data, 0, data.Length)));
+                    count.Should().Be(data.Length);
+                }
+            }
+        }
+        
+        [Test]
+        [LatestClientAndLatestServiceTestCases(testNetworkConditions: false, testAsyncAndSyncClients: false)]
+        public async Task AsyncDataStreamsAreUsedWhenInAsync(ClientAndServiceTestCase clientAndServiceTestCase)
+        {
+            using (var clientAndService = await clientAndServiceTestCase.CreateTestCaseBuilder()
+                       .WithStandardServices()
+                       .AsLatestClientAndLatestServiceBuilder()
+                       .WithForcingClientProxyType(ForceClientProxyType.AsyncClient)
+                       .Build(CancellationToken))
+            {
+                var echo = clientAndService.Client.CreateAsyncClient<IEchoService, IAsyncClientEchoService>(clientAndService.GetServiceEndPoint());
+
+                var data = new byte[1337];
+                new Random().NextBytes(data);
+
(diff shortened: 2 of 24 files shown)


source/Halibut.TestUtils.CompatBinary.Base/Services/FileTransferService.cs
```diff
@@ -2,7 +2,7 @@
 using System.IO;
 using Octopus.Tentacle.Contracts;
 
-namespace Halibut.TestUtils.Contracts.Tentacle.Services
```
LukeButters, line 5: The implementation of the service has changed slightly because of the change in data stream, so it has been copied into the backwards compat project

source/Halibut.Tests/TestServices/FileTransferService.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+using System.IO;
+using Octopus.Tentacle.Contracts;
+
+namespace Halibut.Tests.TestServices
+{
+    public class FileTransferService : IFileTransferService
```
LukeButters, line 7: This has been moved into the Halibut test repo so it can depend on the latest version of DataStream so that it can make use of the async writer.

source/Halibut.Tests/Transport/Protocol/MessageSerializerFixture.cs
```diff
@@ -302,14 +302,14 @@ static byte[] DeflateString(string s)
             }
         }
         
-        async Task<T> ReadMessage<T>(MessageSerializerTestCase testCase, MessageSerializer messageSerializer, RewindableBufferStream rewindableBufferStream)
+        async Task<T> ReadMessage<T>(MessageSerializerTestCase testCase, MessageSerializer messageSerializer, RewindableBufferStream RewindableBufferStream)
```
sburmanoctopus, line 305: Capital "R"?

source/Halibut.TestUtils.CompatBinary.Base/Services/FileTransferService.cs
sburmanoctopus: Now that we have both versions (one for legacy, and one for regular tests), I wonder if we should rename this to something like `LegacyFileTransferService` to stop confusion. Just a thought.

source/Halibut/DataStream.cs
```diff
@@ -9,18 +11,38 @@ namespace Halibut
...
         IDataStreamReceiver receiver;
 
         [JsonConstructor]
         public DataStream()
         {
         }
-
-        public DataStream(long length, Action<Stream> writer)
+        
+        public DataStream(long length, Action<Stream> writer, Func<Stream, CancellationToken, Task> writerAsync)
```
sburmanoctopus, line 22: Could we make this internal to discourage normal users from giving both?

source/Halibut/DataStream.cs
```diff
@@ -85,18 +112,48 @@ public static DataStream FromString(string text)
...
+                {
+                    var writer = new StreamWriter(stream, encoding);
+                    // TODO - ASYNC ME UP!
+                    // Writer does not support taking a cancellation token, should we do something else here?
+                    await writer.WriteAsync(text);
+                    await writer.FlushAsync();
+                });
+        }
+
+        // TODO test
```
sburmanoctopus, line 132: Did this TODO get done?

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
```diff
@@ -465,6 +471,24 @@ TemporaryFileStream CopyStreamToFile(Guid id, long length, BinaryReader reader)
...
             return new TemporaryFileStream(path, log);
         }
+        
+        async Task<TemporaryFileStream> CopyStreamToFileAsync(Guid id, long length, Stream stream, CancellationToken cancellationToken)
+        {
+            var path = Path.Combine(Path.GetTempPath(), string.Format("{0}_{1}", id.ToString(), Interlocked.Increment(ref streamCount)));
+            long bytesLeftToRead = length;
+            using (var fileStream = new FileStream(path, FileMode.Create, FileAccess.Write))
+            {
+                var buffer = new byte[65*1024];
```
sburmanoctopus, line 481: The buffer size here is different to the sync version. Is that done to keep the size < LOH threshold?

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
```diff
@@ -465,6 +471,24 @@ TemporaryFileStream CopyStreamToFile(Guid id, long length, BinaryReader reader)
             }
             return new TemporaryFileStream(path, log);
         }
+        
+        async Task<TemporaryFileStream> CopyStreamToFileAsync(Guid id, long length, Stream stream, CancellationToken cancellationToken)
```
sburmanoctopus, line 475: The sync version uses a `BinaryReader`, and this doesn't.

From what I can see, there is nothing that the BinaryReader does that the Stream could not do.

But the fact it's there makes me wonder if there is a reason it's there that we are not aware of.

Do you think it worth looking at history and/or reaching out to find out?

sburmanoctopus, line 475: The `stream` parameter looks to be the member variable. That could get confusing in the method itself. Maybe delete `stream`?

source/Halibut.Tests/DataStreamFixture.cs
```diff
@@ -0,0 +1,104 @@
...
+
+namespace Halibut.Tests
+{
+    public class DataStreamFixture : BaseTest
+    {
+        [Test]
+        [LatestClientAndLatestServiceTestCases(testNetworkConditions: false)]
+        [LatestClientAndPreviousServiceVersionsTestCases(testNetworkConditions: false)]
+        [Obsolete]
+        public async Task ASyncDataStreamWriterCanStillBeUsed(ClientAndServiceTestCase clientAndServiceTestCase)
```
sburmanoctopus, line 21: Is this supposed to be `SyncDataStreamWriterCanStillBeUsed` instead?

I wondered why an async test was marked as `Obsolete` straight away.

source/Halibut.Tests/DataStreamFixture.cs
```diff
@@ -0,0 +1,104 @@
...
+                    var count = await echo.CountBytesAsync(new DataStream(data.Length,
+                            async (stream, ct) => await stream.WriteByteArrayAsync(data, ct))
+                    );
+                    count.Should().Be(data.Length);
+                }
+            }
+        }
+
+        [Test]
+        public async Task ASyncDataStreamWriter_CanBeUsedInAsync()
```
sburmanoctopus, line 94: The other tests in this class are more "integration style" tests, whereas this is more a "unit style" test.

Could we exercise this in an integration style way?

If not, I wonder if it would be clearer to separate this test from the others (in a new class). Personally, I like having all tests in a class being tested at a similar level 😁

sburmanoctopus (approved):
LGTM. Some minor things, but nothing to stop this merging.

source/Halibut.TestUtils.CompatBinary.Base/Services/FileTransferService.cs
LukeButters: I don't think this is referenced in Halibut tests in a way where the class can be used. So you can never use the wrong one.

source/Halibut.Tests/DataStreamFixture.cs
LukeButters: No no its "A" "Sync" as to distinct words. Do you even camel case? :p 

I will change it

source/Halibut.Tests/Transport/Protocol/MessageSerializerFixture.cs
LukeButters: oopsies I don't even know how or why

source/Halibut/DataStream.cs
LukeButters: We probably do want to give both in `Octopus` while we roll out.

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
LukeButters: yes

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
LukeButters: It might be because previously we had wrapped the stream in a binary stream, and we just happened to pass that around.

PR #402, opened 2023-08-08
HttpProxy Disposal and Pausing + ProxyFixture Tests + HttpProxyClient Async Timeout Bugs

nathanwoctopusdeploy:
# Background

Made the HttpProxy Dispose more consistently and added the ability to Pause new connections

Additional tests in ProxyFixture to cover connecting timeouts and connected but the Connect phase has stalled

Wrapped the stream int the HttpProxyClient in a NetworkTimeoutStream for consistent cancellation and timeouts across frameworks

Fixed a bug in the HttpProxyClient around read timeouts for the new Async methods

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.TestProxy/HttpProxyService.cs
+++ b/source/Halibut.TestProxy/HttpProxyService.cs
@@ -43,6 +43,8 @@ namespace Halibut.TestProxy
 
         public async Task StartAsync()
         {
+            Exception? startUpException = null;
+            
             _ = Task.Run(async () =>
             {
                 try
@@ -51,18 +53,20 @@ namespace Halibut.TestProxy
                 }
                 catch (Exception ex)
                 {
+                    startUpException = ex;
                     logger.LogError(ex, "An error has occurred running the HTTP proxy service");
-                }
-                finally
-                {
-                    // Ensure that StartAsync can complete,
-                    // regardless of whether there was an error or not
-                    started = true;
                 }
             });
 
             while (!started)
             {
+                cancellationTokenSource.Token.ThrowIfCancellationRequested();
+
+                if (startUpException != null)
+                {
+                    throw startUpException;
+                }
+                
                 await Task.Delay(TimeSpan.FromMilliseconds(10), cancellationTokenSource.Token);
             }
         }
--- a/source/Halibut.TestProxy/HttpProxyService.cs
+++ b/source/Halibut.TestProxy/HttpProxyService.cs
@@ -43,7 +43,7 @@ namespace Halibut.TestProxy
 
         public async Task StartAsync()
         {
-            var _ = Task.Run(async () =>
+            _ = Task.Run(async () =>
             {
                 try
                 {
@@ -52,6 +52,12 @@ namespace Halibut.TestProxy
                 catch (Exception ex)
                 {
                     logger.LogError(ex, "An error has occurred running the HTTP proxy service");
+                }
+                finally
+                {
+                    // Ensure that StartAsync can complete,
+                    // regardless of whether there was an error or not
+                    started = true;
                 }
             });
 
@@ -192,8 +198,9 @@ namespace Halibut.TestProxy
             }
             finally
             {
-                await reader.CompleteAsync();
-                await writer.CompleteAsync();
+                await Task.WhenAll(
+                    reader.CompleteAsync().AsTask(),
+                    writer.CompleteAsync().AsTask());
             }
         }
 
--- a/source/Halibut.TestProxy/HttpProxyService.cs
+++ b/source/Halibut.TestProxy/HttpProxyService.cs
@@ -7,7 +7,6 @@ using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 
 #if !NET5_0_OR_GREATER
@@ -26,11 +25,14 @@ namespace Halibut.TestProxy
 
     record HttpProxyConnectionRequest(ProxyEndpoint Endpoint, string HttpVersion);
 
-    public class HttpProxyService : BackgroundService
+    public class HttpProxyService : IDisposable
     {
         readonly HttpProxyOptions options;
         readonly ILogger<HttpProxyService> logger;
         readonly IProxyConnectionService proxyConnectionService;
+        readonly CancellationTokenSource cancellationTokenSource = new();
+        bool pauseNewConnections = false;
+        bool started = false;
 
         public HttpProxyService(HttpProxyOptions options, ILoggerFactory loggerFactory)
         {
@@ -39,17 +41,39 @@ namespace Halibut.TestProxy
             this.logger = loggerFactory.CreateLogger<HttpProxyService>();
         }
 
-        protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+        public async Task StartAsync()
         {
+            var _ = Task.Run(async () =>
+            {
+                try
+                {
+                    await StartInternalAsync();
+                }
+                catch (Exception ex)
+                {
+                    logger.LogError(ex, "An error has occurred running the HTTP proxy service");
+                }
+            });
+
+            while (!started)
+            {
+                await Task.Delay(TimeSpan.FromMilliseconds(10), cancellationTokenSource.Token);
+            }
+        }
+
+        public async Task StartInternalAsync()
+        {
+            var cancellationToken = cancellationTokenSource.Token;
+
             var listener = await TcpListenerHelpers.GetTcpListener(options.Listen);
             listener.Start();
             Endpoint = (IPEndPoint)listener.LocalEndpoint;
-
+            started = true;
             logger.LogInformation("Listening for HTTP proxy requests on {Listen}", listener.Server.LocalEndPoint);
 
             try
             {
-                while (!stoppingToken.IsCancellationRequested)
+                while (!cancellationToken.IsCancellationRequested)
                 {
                     try
                     {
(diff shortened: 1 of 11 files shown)


nathanwoctopusdeploy:
[sc-53211]

source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
```diff
@@ -200,5 +200,24 @@ public override long Position
             get => inner.Position;
             set => inner.Position = value;
         }
+
+        public async Task<bool> DataAvailable(CancellationToken cancellationToken)
```
nathanwoctopusdeploy, line 204: I think this can be a pass-through if supported property rather than being wrapped in the timeout and cancellation logic. Will change and ensure it doesn't regress the timeout tests

source/Halibut.TestProxy/HttpProxyService.cs
```diff
@@ -39,45 +41,78 @@ public HttpProxyService(HttpProxyOptions options, ILoggerFactory loggerFactory)
             this.logger = loggerFactory.CreateLogger<HttpProxyService>();
         }
 
-        protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+        public async Task StartAsync()
         {
+            var _ = Task.Run(async () =>
```
nathanwoctopusdeploy, line 46: On a second look this is going to hang if `StartInternalAsync` fails before `started == true`. Will revisit and ensure it bails in that condition

source/Halibut.TestProxy/ProxyConnection.cs
```diff
@@ -29,44 +28,36 @@ public ProxyConnection(ProxyEndpoint destinationEndpoint, ILogger<ProxyConnectio
 
         public async Task Connect(TcpClient source, CancellationToken cancellationToken)
         {
-            await sync.WaitAsync(cancellationToken);
```
nathanwoctopusdeploy, line 32: redundant locking

source/Halibut.TestProxy/TcpTunnel.cs
```diff
@@ -31,7 +31,7 @@ public async Task Tunnel(CancellationToken cancellationToken)
                 var fromStreamTask = fromStream.CopyToAsync(toWriter, cancellationToken);
                 var toStreamTask = toStream.CopyToAsync(fromWriter, cancellationToken);
 
-                await Task.WhenAny(fromStreamTask, toStreamTask);
+                await Task.WhenAny(fromStreamTask, toStreamTask, cancellationToken.AsTask());
```
nathanwoctopusdeploy, line 34: Jetpack when cancelled

source/Halibut.Tests/Support/BackwardsCompatibility/LatestClientAndPreviousServiceVersionBuilder.cs
```diff
@@ -380,7 +380,12 @@ public TService CreateClient<TService>(Action<ServiceEndPoint> modifyServiceEndp
             public void Dispose()
             {
                 var logger = new SerilogLoggerBuilder().Build().ForContext<ClientAndService>();
-                logger.Information("Dispose called");
+                
+                logger.Information("****** ****** ****** ****** ****** ****** ******");
```
nathanwoctopusdeploy, line 384: Trying to make the logs easier to read. We get lots of errors after Dispose as everything is shutting down. Ideally, we would suppress those, but that is more involved. This is a quick bandaid so visually we can see when we dispose.

source/Halibut/Transport/Proxy/HttpProxyClient.cs
```diff
@@ -231,6 +232,7 @@ public TcpClient CreateConnection(string destinationHost, int destinationPort, T
                     // attempt to open the connection
                     log.Write(EventType.Diagnostic, "Connecting to proxy at {0}:{1}", ProxyHost, ProxyPort);
                     TcpClient.ConnectWithTimeout(ProxyHost, ProxyPort, timeout, cancellationToken);
+                    log.Write(EventType.Diagnostic, "Connected to proxy at {0}:{1}", ProxyHost, ProxyPort);
```
nathanwoctopusdeploy, line 235: Helpful when determining where the http proxy client got stuck

source/Halibut/Transport/Proxy/HttpProxyClient.cs
```diff
@@ -431,12 +432,12 @@ void WaitForData(NetworkStream stream)
...
         
-        async Task WaitForDataAsync(NetworkStream stream, CancellationToken cancellationToken)
+        async Task WaitForDataAsync(NetworkTimeoutStream stream, CancellationToken cancellationToken)
         {
             var sleepTime = 0;
-            while (!stream.DataAvailable)
+            while (!await stream.DataAvailable(cancellationToken))
             {
-                await Task.Delay(WAIT_FOR_DATA_TIMEOUT, cancellationToken);
+                await Task.Delay(WAIT_FOR_DATA_INTERVAL, cancellationToken);
```
nathanwoctopusdeploy, line 440: Bug making the timeout take 3.8 ish mins rather than 15 seconds

source/Halibut/Transport/Proxy/HttpProxyClient.cs
```diff
@@ -431,12 +432,12 @@ void WaitForData(NetworkStream stream)
             }
         }
         
-        async Task WaitForDataAsync(NetworkStream stream, CancellationToken cancellationToken)
+        async Task WaitForDataAsync(NetworkTimeoutStream stream, CancellationToken cancellationToken)
```
nathanwoctopusdeploy, line 435: Now that we have cancellation and timeout on reads, this could probably be simplified to a read async with a short timeout set on the stream

source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
LukeButters: Where you finding that `DataAvailable` was blocking?

another trick people use is ask to read into an empty array, although I suspect our own streams will struggle to support that.

source/Halibut/Transport/Proxy/HttpProxyClient.cs
LukeButters: It looks like we **do** have test coverage in this area, so maybe we can do that instead 👍

source/Halibut/Transport/Proxy/HttpProxyClient.cs
```diff
@@ -337,9 +340,7 @@ void SendConnectionCommand(string host, int port)
         
         async Task SendConnectionCommandAsync(string host, int port, CancellationToken cancellationToken)
         {
-            // TODO - ASYNC ME UP!
-            // This stream needs to be wrapped into a TimeoutStream
-            var stream = TcpClient.GetStream();
+            var stream = new NetworkTimeoutStream(TcpClient.GetStream());
```
LukeButters, line 343: Did you have extension methods on `TcpClient` to make a timed out stream?

source/Halibut/Transport/Proxy/HttpProxyClient.cs
LukeButters: Was it getting stuck?

source/Halibut.Tests/ProxyFixture.cs
```diff
@@ -17,41 +15,75 @@ public class ProxyFixture : BaseTest
...
+
+                var echo = clientAndService.CreateClient<IEchoService, IAsyncClientEchoService>(point =>
+                {
+                    point.TcpClientConnectTimeout = TimeSpan.FromSeconds(5);
+                    point.RetryCountLimit = 2;
+                    point.PollingRequestQueueTimeout = TimeSpan.FromSeconds(10);
+                });
+
+                Func<Task<string>> action = async () => await echo.SayHelloAsync("Hello");
+                (await action.Should().ThrowAsync<HalibutClientException>()).And.Message.Should().ContainAny(
```
LukeButters, line 55: `AssertAsync` does mean you can avoid the `Func<Task` bit.

source/Halibut.TestProxy/TcpTunnel.cs
```diff
@@ -40,6 +40,10 @@ public async Task Tunnel(CancellationToken cancellationToken)
             {
                 await fromWriter.CompleteAsync();
                 await toWriter.CompleteAsync();
+                fromStream.Close();
+                toStream.Close();
```
LukeButters, line 44: Consider using that `Try` class here. Since if `fromStrean.Close()` throws the other close methods will not be called.

source/Halibut.TestProxy/HttpProxyService.cs
```diff
@@ -88,70 +123,78 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
+                }
+
+                logger.LogInformation("CONNECT request successful: {SourceEndpoint} -> {DestinationEndpoint}", sourceEndpoint, connectionRequest.Endpoint);
             }
-            catch (Exception ex)
+            finally
             {
-                logger.LogError(ex, "An error has occurred responding to CONNECT request: {SourceEndpoint} <-> {DestinationEndpoint}", sourceEndpoint, connectionRequest!.Endpoint);
-                return;
+                await reader.CompleteAsync();
```
LukeButters, line 195: Does it make sense to do a when all here?

LukeButters (approved):

source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
nathanwoctopusdeploy: > Where you finding that DataAvailable was blocking

I don't think it was the cause that turned out to be another bug. This was leftover from trying to fix the issue

source/Halibut/Transport/Proxy/HttpProxyClient.cs
nathanwoctopusdeploy: The connect was not, that timeout worked. It was the read/write after that got stuck. This is just helpful in the logs as we see it is trying to connect and then it did connect

source/Halibut/Transport/Proxy/HttpProxyClient.cs
nathanwoctopusdeploy: No, we could / we can just do this simple wrap!!

source/Halibut.TestProxy/HttpProxyService.cs
nathanwoctopusdeploy: Yes that could be good 👍

source/Halibut.TestProxy/HttpProxyService.cs
```diff
@@ -88,70 +129,79 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
+                logger.LogInformation("CONNECT request successful: {SourceEndpoint} -> {DestinationEndpoint}", sourceEndpoint, connectionRequest.Endpoint);
             }
-            catch (Exception ex)
+            finally
             {
-                logger.LogError(ex, "An error has occurred responding to CONNECT request: {SourceEndpoint} <-> {DestinationEndpoint}", sourceEndpoint, connectionRequest!.Endpoint);
-                return;
+                await Task.WhenAll(
+                    reader.CompleteAsync().AsTask(),
+                    writer.CompleteAsync().AsTask());
```
acodrington, line 203: The `CompleteAsync` calls return a `ValueTask` rather than a regular `Task`, so we need to use `AsTask` to wrap them so we can use `WhenAll`.

I imagine would have a minor cost, but unsure if that cost is signifcantly more/less than the extra time of awaitng sequentially like before.

source/Halibut.TestProxy/HttpProxyService.cs
```diff
@@ -53,6 +53,12 @@ public async Task StartAsync()
                 {
                     logger.LogError(ex, "An error has occurred running the HTTP proxy service");
                 }
+                finally
+                {
+                    // Ensure that StartAsync can complete,
+                    // regardless of whether there was an error or not
+                    started = true;
```
nathanwoctopusdeploy, line 60: I think `StartAsync` should throw if it fails to start the listener e.g. start is never set to true by `StartInternalAsync`

e.g.

```
public async Task StartAsync()
{
            Exception? startupException = null;
            var _ = Task.Run(async () =>
            {
                try
                {
                    await StartInternalAsync();
                }
                catch (Exception ex)
                {
                    startupException = ex;
                    logger.LogError(ex, "An error has occurred running the HTTP proxy service");
                }
            });

            while (!started)
            {
               cancellationTokenSource.Token.ThrowIfCancellationRequested();
            
                if(startupException != null)
                {
                    throw startupException;
                }
            
                await Task.Delay(TimeSpan.FromMilliseconds(10), cancellationTokenSource.Token);
            }
}
```

source/Halibut/Transport/Proxy/HttpProxyClient.cs
nathanwoctopusdeploy: Scope hammer this for now

source/Halibut.TestProxy/HttpProxyService.cs
acodrington: That's nicer for sure, no worries.

PR #404, opened 2023-08-08
Zero length read does not mean the end of stream was found

LukeButters:
# Background

We have never seen this cause an error however we should implement our stream wrappers correctly. This fixes the error recording stream.

https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.read?view=net-7.0

image

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Transport/Streams/ErrorRecordingStreamFixture.cs
+++ b/source/Halibut.Tests/Transport/Streams/ErrorRecordingStreamFixture.cs
@@ -114,6 +114,22 @@ namespace Halibut.Tests.Transport.Streams
         }
         
         [Test]
+        public async Task ReadingIntoAZeroLengthArrayIsNotAEndOfStream()
+        {
+            int counter = 0;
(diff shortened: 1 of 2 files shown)


source/Halibut.Tests/Transport/Streams/ErrorRecordingStreamFixture.cs
```diff
@@ -113,6 +113,22 @@ public async Task EndOfStreamEncounteredIsRecordedAsync()
             errorRecordingStream.ReadExceptions.Count.Should().Be(0);
         }
         
+        [Test]
+        public async Task ReadingIntoAZeroLengthArrayIsNotAEndOfStream()
```
sburmanoctopus, line 117: This test seems to be testing a few test cases:
- 100 length array with 0 count
- 0 length array (which has to have a 0 count)

Maybe we should split these test cases up?

Also, should we also verify the return value to be 0 as well?

sburmanoctopus (approved):
LGTM. Just a small test suggestion, but nothing that should block this.

nathanwoctopusdeploy (approved):

PR #407, opened 2023-08-09
Make PollingClient support async

LukeButters:
# Background

[SC-53211]

Polling Client now supports async, which means that the polling loop is completely async (no Threads are created). The choose between async and sync in the polling loop is based on `AsyncHalibutFeature`.

By default the tests will continue to test only against a "sync" service which in `PollingClient` terms means creates a new thread and call the old sync methods.

An option has been added to the standard attributes which allows for additionally testing async services and this has been enabled on some tests. For now we only need some tests to test this, as we wont be using the async service for some time. When we do enable async services on all tests the result will be almost doubling the number of tests run.

# Additional optional shorting of test names

With the addition of the `AsyncService/SyncService` parameter in tests, the test name was again frequently becoming longer than what rider would show. Since I needed to see what tests were running this PR also adds a new environment variable `UseShortTestNames` which when set to `true` will cause the test names to be shortened:

image

Personally I prefer those shorter test names.

The test names in teamcity wont be shortened and are locally not shortened by default.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Support/EnvironmentVariableReaderHelper.cs
+++ b/source/Halibut.Tests/Support/EnvironmentVariableReaderHelper.cs
@@ -1,0 +1,19 @@
+﻿using System;
+using Castle.Core.Internal;
+
+namespace Halibut.Tests.Support
+{
+    public class EnvironmentVariableReaderHelper
+    {
+        public static bool EnvironmentVariableAsBool(string envVar, bool defaultValue)
+        {
+            var value = Environment.GetEnvironmentVariable(envVar);
+            if (string.IsNullOrEmpty(value))
+            {
+                return defaultValue;
+            }
+
+            return value!.Equals("true");
+        }
+    }
+}
--- a/source/Halibut.Tests/Support/EnvironmentVariableReaderHelper.cs
+++ b/source/Halibut.Tests/Support/EnvironmentVariableReaderHelper.cs
@@ -8,7 +8,7 @@ namespace Halibut.Tests.Support
         public static bool EnvironmentVariableAsBool(string envVar, bool defaultValue)
         {
             var value = Environment.GetEnvironmentVariable(envVar);
-            if (string.IsNullOrEmpty(value))
+            if (string.IsNullOrWhiteSpace(value))
             {
                 return defaultValue;
             }
--- a/source/Halibut.Tests/Support/HalibutRuntimeBuilderExtensionMethods.cs
+++ b/source/Halibut.Tests/Support/HalibutRuntimeBuilderExtensionMethods.cs
@@ -1,0 +1,16 @@
+﻿namespace Halibut.Tests.Support
+{
+    public static class HalibutRuntimeBuilderExtensionMethods
+    {
+
+        public static HalibutRuntimeBuilder WithAsyncHalibutFeatureEnabledIfForcingAsync(this HalibutRuntimeBuilder halibutRuntimeBuilder, ForceClientProxyType? forceClientProxyType)
+        {
+            if (forceClientProxyType == ForceClientProxyType.AsyncClient)
+            {
+                return halibutRuntimeBuilder.WithAsyncHalibutFeatureEnabled();
+            }
+
+            return halibutRuntimeBuilder;
+        }
+    }
+}
--- a/source/Halibut.Tests/Support/HalibutRuntimeBuilderExtensionMethods.cs
+++ b/source/Halibut.Tests/Support/HalibutRuntimeBuilderExtensionMethods.cs
@@ -2,7 +2,6 @@
 {
     public static class HalibutRuntimeBuilderExtensionMethods
     {
-
         public static HalibutRuntimeBuilder WithAsyncHalibutFeatureEnabledIfForcingAsync(this HalibutRuntimeBuilder halibutRuntimeBuilder, ForceClientProxyType? forceClientProxyType)
         {
             if (forceClientProxyType == ForceClientProxyType.AsyncClient)
--- a/source/Halibut.Tests/Support/TestAttributes/LatestAndPreviousClientAndServiceVersionsTestCasesAttribute.cs
+++ b/source/Halibut.Tests/Support/TestAttributes/LatestAndPreviousClientAndServiceVersionsTestCasesAttribute.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Halibut.Tests.Support.BackwardsCompatibility;
 using Halibut.Tests.Support.TestCases;
+using Halibut.Util;
 using NUnit.Framework;
 
 namespace Halibut.Tests.Support.TestAttributes
@@ -15,18 +16,19 @@ namespace Halibut.Tests.Support.TestAttributes
             bool testNetworkConditions = true,
             bool testListening = true,
             bool testPolling = true,
-            bool testAsyncAndSyncClients = true
+            bool testAsyncAndSyncClients = true,
+            bool testAsyncServicesAsWell = false // False means only the sync service will be tested.
             ) :
             base(
                 typeof(LatestAndPreviousClientAndServiceVersionsTestCases),
                 nameof(LatestAndPreviousClientAndServiceVersionsTestCases.GetEnumerator),
-                new object[] { testWebSocket, testNetworkConditions, testListening, testPolling, testAsyncAndSyncClients})
+                new object[] { testWebSocket, testNetworkConditions, testListening, testPolling, testAsyncAndSyncClients, testAsyncServicesAsWell})
         {
         }
         
         static class LatestAndPreviousClientAndServiceVersionsTestCases
         {
-            public static IEnumerator<ClientAndServiceTestCase> GetEnumerator(bool testWebSocket, bool testNetworkConditions, bool testListening, bool testPolling, bool testAsyncAndSyncClients)
+            public static IEnumerator<ClientAndServiceTestCase> GetEnumerator(bool testWebSocket, bool testNetworkConditions, bool testListening, bool testPolling, bool testAsyncAndSyncClients, bool testAsyncServicesAsWell)
             {
                 var serviceConnectionTypes = ServiceConnectionTypes.All.ToList();
 
@@ -51,6 +53,12 @@ namespace Halibut.Tests.Support.TestAttributes
                     clientProxyTypesToTest = ForceClientProxyTypeValues.All;
                 }
                 
+                var serviceAsyncHalibutFeatureTestCases = AsyncHalibutFeatureValues.All().ToList();
+                if (!testAsyncServicesAsWell)
+                {
+                    serviceAsyncHalibutFeatureTestCases.Remove(AsyncHalibutFeature.Enabled);
+                }
+                
                 var builder = new ClientAndServiceTestCasesBuilder(
                     new[] {
                         ClientAndServiceTestVersion.Latest(),
@@ -60,7 +68,8 @@ namespace Halibut.Tests.Support.TestAttributes
                     },
                     serviceConnectionTypes.ToArray(),
                     testNetworkConditions ? NetworkConditionTestCase.All : new[] { NetworkConditionTestCase.NetworkConditionPerfect },
-                    clientProxyTypesToTest
+                    clientProxyTypesToTest,
+                    serviceAsyncHalibutFeatureTestCases
                     );
 
                 return builder.Build().GetEnumerator();
--- a/source/Halibut.Tests/Support/TestAttributes/LatestClientAndPreviousServiceVersionsTestCasesAttribute.cs
+++ b/source/Halibut.Tests/Support/TestAttributes/LatestClientAndPreviousServiceVersionsTestCasesAttribute.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Halibut.Tests.Support.BackwardsCompatibility;
 using Halibut.Tests.Support.TestCases;
+using Halibut.Util;
 using NUnit.Framework;
 
 namespace Halibut.Tests.Support.TestAttributes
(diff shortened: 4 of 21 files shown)


source/Halibut/Transport/PollingClient.cs
```diff
@@ -13,47 +15,66 @@ public class PollingClient : IPollingClient
...
         readonly ISecureClient secureClient;
         readonly Uri subscription;
-        readonly Thread thread;
-        readonly CancellationToken cancellationToken;
+        Thread? pollingClientLoopThread;
         bool working;
-        readonly Func<RetryPolicy> createRetryPolicy;
-
-        [Obsolete("Use the overload that provides a logger. This remains for backwards compatibility.")]
-        public PollingClient(Uri subscription, ISecureClient secureClient, Func<RequestMessage, ResponseMessage> handleIncomingRequest, Func<RetryPolicy> createRetryPolicy)
```
LukeButters, line 22: No one within halibut uses these and no one outside of halibut would be using this so it should be safe to rmove.

source/Halibut/Transport/PollingClient.cs
```diff
@@ -13,47 +15,66 @@ public class PollingClient : IPollingClient
...
-            : this(subscription, secureClient, handleIncomingRequest, null, createRetryPolicy)
-        {
-        }
 
-        public PollingClient(Uri subscription, ISecureClient secureClient, Func<RequestMessage, ResponseMessage> handleIncomingRequest, ILog log, Func<RetryPolicy> createRetryPolicy)
-            : this(subscription, secureClient, handleIncomingRequest, log, CancellationToken.None, createRetryPolicy)
-        {
-        }
+        Task? pollingClientLoopTask;
+        readonly CancellationTokenSource workingCancellationTokenSource;
```
LukeButters, line 22: Rather than a bool which is used to eventually stop the thread on dispose I made a linked cancellation token which is cancelled and disposed.

source/Halibut/Transport/PollingClient.cs
```diff
@@ -13,47 +15,66 @@ public class PollingClient : IPollingClient
...
+
+            if (asyncHalibutFeature == AsyncHalibutFeature.Disabled)
+            {
+                pollingClientLoopThread = new Thread(ExecutePollingLoop!);
+                pollingClientLoopThread.Name = "Polling client for " + secureClient.ServiceEndpoint + " for subscription " + subscription;
+                pollingClientLoopThread.IsBackground = true;
+            }
+            else
+            {
+                pollingClientLoopTask = Task.Run(async () => await ExecutePollingLoopAsyncCatchingExceptions());
```
nathanwoctopusdeploy, line 54: Do we really want to change it from a Thread to a Task?

LukeButters: I am confused, is this not what making it async means?

LukeButters: I think the answer is yes.

Otherwise we end up with `new Thread(() => ExecutePollingLoopAsyncCatchingExceptions().GetAwaiter().GetResult())`;

So we end up making a thread we don't need.

Doesn't matter too much in tentacle, however it will be less efficient when we have many clients e.g. in tentacle simulator or tests within Halibut.

source/Halibut/Transport/PollingClient.cs
```diff
@@ -13,47 +15,67 @@ public class PollingClient : IPollingClient
...
+            }
         }
 
         public void Dispose()
         {
             working = false;
-        }
+            try
```
sburmanoctopus, line 62: Just wondering, why would we need this wrapped in a try/catch?

The only exceptions thrown are if it has been disposed already (which it shouldn't dispose again), and via callbacks (which we don't assign).

source/Halibut/Transport/PollingClient.cs
```diff
@@ -95,5 +117,56 @@ void ExecutePollingLoop(object ignored)
...
+            catch (Exception)
+            {
+                // We may get errors about the workingCancellationTokenSource being used when it is disposed, we don't care about that.
+            }
+        }
+        async Task ExecutePollingLoopAsync()
+        {
+            var retry = createRetryPolicy();
+            var sleepFor = TimeSpan.Zero;
+            while (!workingCancellationTokenSource.Token.IsCancellationRequested)
```
sburmanoctopus, line 136: I think this will cause exceptions we don't want.

If you create the token right at the start (maybe a member variable), then you can query the `IsCancellationRequested` property all you want. Even if the `workingCancellationTokenSource` has been disposed.

But if you query `workingCancellationTokenSource.Token.IsCancellationRequested` after it has been disposed, then it will throw an exception (as it has to make a new one, and the "shop is closed").

source/Halibut/Transport/PollingClient.cs
```diff
@@ -95,5 +117,56 @@ void ExecutePollingLoop(object ignored)
...
+                    try
+                    {
+                        retry.Try();
+                        await secureClient.ExecuteTransactionAsync(async (protocol, cancellationToken) =>
+                        {
+                            // We have successfully connected at this point so reset the retry policy
+                            // Subsequent connection issues will try and reconnect quickly and then back-off
+                            retry.Success();
+                            await protocol.ExchangeAsSubscriberAsync(subscription, handleIncomingRequest, int.MaxValue, cancellationToken);
+                        }, workingCancellationTokenSource.Token);
```
sburmanoctopus, line 149: Use the above mentioned token here (and line 167) instead of `workingCancellationTokenSource.Token` 🙂

source/Halibut/Transport/PollingClient.cs
```diff
@@ -95,5 +117,56 @@ void ExecutePollingLoop(object ignored)
                 }
             }
         }
+
+        async Task ExecutePollingLoopAsyncCatchingExceptions()
+        {
+            try
```
sburmanoctopus, line 123: I suspect that this try/catch becomes redundant if we implement the token in the way I described before.

source/Halibut/Transport/PollingClient.cs
sburmanoctopus: Just wondering, how long running is this thread?

I.e., do we want it taking a thread from the thread pool, and it'll finish with it quickly? Or should we mark it as a long running task, and let it devote a thread to it properly etc.?

sburmanoctopus (approved):
LGTM

source/Halibut.Tests/Support/HalibutRuntimeBuilderExtensionMethods.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Halibut.Tests.Support
+{
+    public static class HalibutRuntimeBuilderExtensionMethods
+    {
+        public static HalibutRuntimeBuilder WithAsyncHalibutFeatureEnabledIfForcingAsync(this HalibutRuntimeBuilder halibutRuntimeBuilder, ForceClientProxyType? forceClientProxyType)
```
LukeButters, line 5: I was going down a path that needed this a many times, I think it is still worth keeping even though it is called only a couple of times. Since it keeps our big builders smaller.

source/Halibut/Transport/PollingClient.cs
LukeButters: Noted that it is a "just in case" and to avoid unobserved exceptions

source/Halibut.Tests/Support/EnvironmentVariableReaderHelper.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        public static bool EnvironmentVariableAsBool(string envVar, bool defaultValue)
+        {
+            var value = Environment.GetEnvironmentVariable(envVar);
+            if (string.IsNullOrEmpty(value))
+            {
+                return defaultValue;
+            }
+
+            return value!.Equals("true");
```
sburmanoctopus, line 16: Might be worth making this a case insensitive comparison. `bool.Parse` does this nicely 😁

source/Halibut.Tests/Support/EnvironmentVariableReaderHelper.cs
```diff
@@ -0,0 +1,19 @@
...
+using Castle.Core.Internal;
+
+namespace Halibut.Tests.Support
+{
+    public class EnvironmentVariableReaderHelper
+    {
+        public static bool EnvironmentVariableAsBool(string envVar, bool defaultValue)
+        {
+            var value = Environment.GetEnvironmentVariable(envVar);
+            if (string.IsNullOrEmpty(value))
```
sburmanoctopus, line 11: A small thing, but I find string.IsNullOrWhitespace a little better in these cases. Someone shouldn't be setting it to "  "

source/Halibut.Tests/Support/TestAttributes/LatestAndPreviousClientAndServiceVersionsTestCasesAttribute.cs
```diff
@@ -15,18 +16,19 @@ public class LatestAndPreviousClientAndServiceVersionsTestCasesAttribute : TestC
             bool testNetworkConditions = true,
             bool testListening = true,
             bool testPolling = true,
-            bool testAsyncAndSyncClients = true
+            bool testAsyncAndSyncClients = true,
+            bool testAsyncServicesAsWell = false // False means only the sync service will be tested.
```
sburmanoctopus, line 20: Nit: Not a big thing, I don't know about the `AsWell` part. Would `testAsyncServices` be fine as is?

source/Halibut.Tests/Support/TestAttributes/LatestClientAndPreviousServiceVersionsTestCasesAttribute.cs
```diff
@@ -14,18 +15,19 @@ public class LatestClientAndPreviousServiceVersionsTestCasesAttribute : TestCase
             bool testNetworkConditions = true,
             bool testListening = true,
             bool testPolling = true,
-            bool testAsyncAndSyncClients = true
+            bool testAsyncAndSyncClients = true,
+            bool testAsyncServicesAsWell = false // False means only the sync service will be tested.
```
sburmanoctopus, line 19: In `LatestClientAndPreviousServiceVersionBuilder`, we state that "The service is external and so can not be made to use async".

So therefore, should we have `testAsyncServicesAsWell` on this test case attribute? (as it is also 'previous service')

source/Halibut.Tests/Support/TestCases/ClientAndServiceTestCasesBuilder.cs
```diff
@@ -39,21 +43,29 @@ public IEnumerable<ClientAndServiceTestCase> Build()
...
+                        foreach (var serviceAsyncHalibutFeatureTestCase in serviceAsyncHalibutFeatureTestCases)
                         {
-                            yield return new ClientAndServiceTestCase(serviceConnectionType, networkConditionTestCase, recommendedIterations, clientServiceTestVersion, null);
-                        }
-                        else
-                        {
-                            if (clientServiceTestVersion.IsPreviousClient())
+                            if (clientServiceTestVersion.IsPreviousService() && serviceAsyncHalibutFeatureTestCase == AsyncHalibutFeature.Enabled)
+                            {
+                                continue;
```
sburmanoctopus, line 50: I wonder if we should remove this if statement. The `LatestClientAndPreviousServiceVersionBuilder` will throw if this happens, and this will be a good signal to the developer that they configured it wrong.

Otherwise, it will just not appear when they think it should.

source/Halibut.Tests/Support/TestCases/NetworkConditionTestCase.cs
```diff
@@ -17,15 +17,18 @@ public class NetworkConditionTestCase
...
         public static NetworkConditionTestCase NetworkConditionPerfect = 
             new (null, 
+                "Perfect", 
                 "Perfect");
 
         public static NetworkConditionTestCase NetworkCondition20MsLatency = 
             new ((i, logger) => PortForwarderBuilder.ForwardingToLocalPort(i, logger).WithSendDelay(TimeSpan.FromMilliseconds(20)).Build(), 
-                "20ms SendDelay");
+                "20ms SendDelay",
+                "20msLatency");
```
sburmanoctopus, line 26: Could we go even shorter with DLY for delay. Or even SD for Send Delay. I'm pretty whatevs, but just a thought 😁

source/Halibut/Transport/PollingClient.cs
```diff
@@ -13,47 +15,62 @@ public class PollingClient : IPollingClient
...
+            {
+                pollingClientLoopTask = Task.Run(async () => await ExecutePollingLoopAsyncCatchingExceptions(workingCancellationTokenSource.Token));
+            }
         }
 
         public void Dispose()
         {
             working = false;
-        }
+            workingCancellationTokenSource.Cancel();
```
sburmanoctopus, line 62: I think this try/catch is the wrong way around.

I.e., you can call 'Dispose' as much as you want, but 'Cancel' will throw if it is already disposed.

source/Halibut/Transport/PollingClient.cs
```diff
@@ -13,47 +15,62 @@ public class PollingClient : IPollingClient
...
+            if (asyncHalibutFeature == AsyncHalibutFeature.Disabled)
+            {
+                pollingClientLoopThread = new Thread(ExecutePollingLoop!);
+                pollingClientLoopThread.Name = "Polling client for " + secureClient.ServiceEndpoint + " for subscription " + subscription;
+                pollingClientLoopThread.IsBackground = true;
+                pollingClientLoopThread.Start();
+            }
+            else
+            {
+                pollingClientLoopTask = Task.Run(async () => await ExecutePollingLoopAsyncCatchingExceptions(workingCancellationTokenSource.Token));
```
sburmanoctopus, line 55: Just wondering, how long running is this thread?

I.e., do we want it taking a thread from the thread pool, and it'll finish with it quickly? Or should we mark it as a long running task, and let it devote a thread to it properly etc.?

Basically, should we use the `TaskCreationOptions.LongRunning`?

LukeButters: Maybe, you can add this to board as something to investigate later.

source/Halibut/Util/Try.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+
+namespace Halibut.Util
+{
+    public static class Try
+    {
+        public static void CatchingError(Action tryThisAction, Action<Exception> onFailure)
```
sburmanoctopus, line 7: Nit: Instead of passing a required lambda, this could be nullable (even with a default of null), so we don't have to pass in a dummy lambda (like we do in `PollingClient.Dispose`)

source/Halibut.Tests/Support/TestAttributes/LatestClientAndPreviousServiceVersionsTestCasesAttribute.cs
LukeButters: You can change this if it turns out to be important.

sburmanoctopus (approved):
LGTM. Some minor nits, but nothing to stop it merging.

source/Halibut.Tests/Support/TestCases/ClientAndServiceTestCasesBuilder.cs
LukeButters: This does make it easier to use the single attribute that gives you all the latest to previous combos. But it does come at this cost. This is similar to existing attributes which let you say pls test this and then they just don't test the thing that can't be tested in that way.

Maybe all of that can be improved.

PR #409, opened 2023-08-09
Increase hang dump time

LukeButters:
# Background

So that we have time to collect a dump, since currently the dump is either not collected or collected and placed into /tmp/ where we can't get it.

# Pre-requisites

### DIFF
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -114,7 +114,10 @@ class Build : NukeBuild
                     .EnableBlameCrash()
                     .SetBlameCrashDumpType("full")
                     .EnableBlameHang()
-                    .SetBlameHangTimeout(TimeSpan.FromMinutes(20).TotalMilliseconds.ToString())
+                    // This is set high since when a hang dump is collected it is saved into /tmp/
+                    // On windows the dump collecting utility appears to be missing and so nothing is collected.
+                    // Setting high means we will have time to get in and collect a dump manually
+                    .SetBlameHangTimeout(TimeSpan.FromMinutes(9999).TotalMilliseconds.ToString())
                     .SetBlameHangDumpType("full"));
 
             });


build/Build.cs
```diff
@@ -114,7 +114,10 @@ ITargetDefinition TestDefinition(ITargetDefinition targetDefinition, Target depe
                     .EnableBlameCrash()
                     .SetBlameCrashDumpType("full")
                     .EnableBlameHang()
-                    .SetBlameHangTimeout(TimeSpan.FromMinutes(20).TotalMilliseconds.ToString())
+                    // This is set high since when a hang dump is collected it is saved into /tmp/
+                    // On windows the dump collecting utility appears to be missing and so nothing is collected.
+                    // Setting high means we will have time to get in and collect a dump manually
+                    .SetBlameHangTimeout(TimeSpan.FromMinutes(9999).TotalMilliseconds.ToString())
```
nathanwoctopusdeploy, line 120: I don't know if we want to tie up build agents for `6.94375` days when something hangs.

I get we need to spot it's hanging (can team city alert us?) and then get someone to access the machine and take a process dump so we need it to remain online for a while. Could we reduce this down a bit so that if for some reason we don't spot a hung build, we at least release the agent after a reasonable period of time.

nathanwoctopusdeploy: Yeh, dial it down a bit and maybe a bit extra to the comment so the reader knows there is another timeout that kicks in before this one.

nathanwoctopusdeploy (approved):

build/Build.cs
nathanwoctopusdeploy: Ahh auto merge!! Lol, never mind!!

LukeButters: https://github.com/OctopusDeploy/Halibut/pull/415

PR #413, opened 2023-08-09
Make AuthenticateAsClientAsync respect timeouts

sburmanoctopus:
[sc-55405]

# Background

The SSL certificate auth was made async. But it did not respect the timeouts set, causing the async versions of `TimeoutsApplyDuringHandShake` to fail.

# Results

## Before
`AuthenticateAsClientAsync` would not time out, despite what was set in the `TcpClient`

## After
We have now made `AuthenticateAsClientAsync` respect the timeouts from the `TcpClient`.

The `TimeoutsApplyDuringHandShake` could now be made async friendly.

We also noticed that `DiscoveryClient` also used `AuthenticateAsClientAsync`, so we also made that respect the timeouts as well.

### .NET 6.0 vs .NET 4.8
Sadly, only .NET 6.0 has the override of `AuthenticateAsClientAsync` that respects the cancellation token.

So for .NET 4.8, we had to use the `NetworkTimeoutStream`, and enhance it to make it work. This involved mapping between the Task Asynchronous Programming (TAP) and older Asynchronous Programming Model (APM) ([demonstrated by Microsoft](https://learn.microsoft.com/en-us/dotnet/standard/asynchronous-programming-patterns/interop-with-other-asynchronous-patterns-and-types))

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Support/TestAttributes/StreamMethodAttribute.cs
+++ b/source/Halibut.Tests/Support/TestAttributes/StreamMethodAttribute.cs
@@ -1,0 +1,34 @@
+﻿using System;
+using System.Collections;
+using System.Threading.Tasks;
+using Halibut.Tests.Transport.Streams;
+using Halibut.Util;
+using NUnit.Framework;
+
+namespace Halibut.Tests.Support.TestAttributes
+{
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = true, Inherited = false)]
+    public class StreamMethodAttribute : TestCaseSourceAttribute
+    {
+        public StreamMethodAttribute(bool testSync = true) : base(
+            typeof(TestCases),
+            nameof(TestCases.GetEnumerator),
+            new object?[]{testSync})
+        {
+        }
+
+        static class TestCases
+        {
+            public static IEnumerable GetEnumerator(bool testSync)
+            {
+                if (testSync)
+                {
+                    yield return StreamMethod.Sync;
+                }
+                
+                yield return StreamMethod.Async;
+                yield return StreamMethod.LegacyAsync;
+            }
+        }
+    }
+}
--- a/source/Halibut.Tests/Support/TestAttributes/StreamMethodAttribute.cs
+++ b/source/Halibut.Tests/Support/TestAttributes/StreamMethodAttribute.cs
@@ -27,7 +27,8 @@ namespace Halibut.Tests.Support.TestAttributes
                 }
                 
                 yield return StreamMethod.Async;
-                yield return StreamMethod.LegacyAsync;
+                yield return StreamMethod.LegacyAsyncCallEndWithinCallback;
+                yield return StreamMethod.LegacyAsyncCallEndOutsideCallback;
             }
         }
     }
--- a/source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
+++ b/source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
@@ -144,7 +144,7 @@ namespace Halibut.Transport.Streams
 
         public override IAsyncResult BeginRead(byte[] buffer, int offset, int count, AsyncCallback? callback, object? state)
         {
-            // BeginRead does not respect timeouts. So force it to use ReadAsync which does.
+            // BeginRead does not respect timeouts. So force it to use ReadAsync, which does.
             return ReadAsync(buffer, offset, count, CancellationToken.None).AsAsynchronousProgrammingModel(callback, state);
         }
 
--- a/source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
+++ b/source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
@@ -139,16 +139,23 @@ namespace Halibut.Transport.Streams
         public override void Close() => inner.Close();
 
         public override async Task CopyToAsync(Stream destination, int bufferSize, CancellationToken cancellationToken) => await inner.CopyToAsync(destination, bufferSize, cancellationToken);
-
-        public override int EndRead(IAsyncResult asyncResult) => inner.EndRead(asyncResult);
-
-        public override void EndWrite(IAsyncResult asyncResult) => inner.EndWrite(asyncResult);
-
+        
         public override int ReadByte() => inner.ReadByte();
 
-        public override IAsyncResult BeginRead(byte[] buffer, int offset, int count, AsyncCallback? callback, object? state) => inner.BeginRead(buffer, offset, count, callback, state);
+        public override IAsyncResult BeginRead(byte[] buffer, int offset, int count, AsyncCallback? callback, object? state)
+        {
+            // BeginRead does not respect timeouts. So force it to use ReadAsync which does.
+            return ReadAsync(buffer, offset, count, CancellationToken.None).AsAsynchronousProgrammingModel(callback, state);
+        }
+
+        public override int EndRead(IAsyncResult asyncResult)
+        {
+            return ((Task<int>)asyncResult).Result;
+        }
 
         public override IAsyncResult BeginWrite(byte[] buffer, int offset, int count, AsyncCallback? callback, object? state) => inner.BeginWrite(buffer, offset, count, callback, state);
+
+        public override void EndWrite(IAsyncResult asyncResult) => inner.EndWrite(asyncResult);
 
         public override void Flush() => inner.Flush();
 
--- a/source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
+++ b/source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
@@ -169,13 +169,9 @@ namespace Halibut.Transport.Streams
         public override void EndWrite(IAsyncResult asyncResult)
         {
             var task = (Task)asyncResult;
-
             try
             {
-                if (task.Exception is not null)
-                {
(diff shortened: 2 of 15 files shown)


source/Halibut/Util/TaskExtensionMethods.cs
```diff
@@ -24,5 +25,34 @@ public static void IgnoreUnobservedExceptions(this Task task)
                 },
                 TaskContinuationOptions.OnlyOnFaulted | TaskContinuationOptions.ExecuteSynchronously);
         }
+
+
+        public static IAsyncResult AsAsynchronousProgrammingModel<T>(
```
sburmanoctopus, line 30: This bit was stolen from [that Microsoft site](https://learn.microsoft.com/en-us/dotnet/standard/asynchronous-programming-patterns/interop-with-other-asynchronous-patterns-and-types)

source/Halibut/Transport/TcpConnectionFactory.cs
```diff
@@ -52,16 +54,22 @@ public async Task<IConnection> EstablishNewConnectionAsync(ExchangeProtocolBuild
             var certificateValidator = new ClientCertificateValidator(serviceEndpoint);
             var client = await CreateConnectedTcpClientAsync(serviceEndpoint, log, cancellationToken);
             log.Write(EventType.Diagnostic, $"Connection established to {client.Client.RemoteEndPoint} for {serviceEndpoint.BaseUri}");
-
-            var stream = client.GetStream();
+            
+            var networkStream = client.GetStream();
```
LukeButters, line 58: Should this networkstream always be wrapped in a networkTimeoutStream? Are we wrapping the ssl stream at a later point?

source/Halibut/Transport/TcpConnectionFactory.cs
```diff
@@ -52,16 +54,22 @@ public async Task<IConnection> EstablishNewConnectionAsync(ExchangeProtocolBuild
...
 
             log.Write(EventType.SecurityNegotiation, "Performing TLS handshake");
-            var ssl = new SslStream(stream, false, certificateValidator.Validate, UserCertificateSelectionCallback);
+
+#if NETFRAMEWORK
```
nathanwoctopusdeploy, line 62: Could `AuthenticateAsClientEnforcingTimeout` contain the `#if ; #else`  stuff as well so it's all wrapped up in one extension method?

source/Halibut/Util/TaskExtensionMethods.cs
LukeButters: Could you note that in the code itself, so others will know what to read when they turn up to this?

source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
```diff
@@ -139,17 +139,24 @@ void ThrowMeaningfulException(Exception? innerException = null)
...
+            // BeginRead does not respect timeouts. So force it to use ReadAsync, which does.
+            return ReadAsync(buffer, offset, count, CancellationToken.None).AsAsynchronousProgrammingModel(callback, state);
+        }
+
+        public override int EndRead(IAsyncResult asyncResult)
+        {
+            return ((Task<int>)asyncResult).Result;
+        }
 
         public override IAsyncResult BeginWrite(byte[] buffer, int offset, int count, AsyncCallback? callback, object? state) => inner.BeginWrite(buffer, offset, count, callback, state);
```
LukeButters, line 156: We should probably do the equivalent on `BeginWrite`

source/Halibut/Transport/TcpConnectionFactory.cs
sburmanoctopus: We could. But the regular 'sync' version respects the timeouts. So there's no need.

Do you think it worth doing anyway for consistency?

source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
sburmanoctopus: On it

source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
```diff
@@ -139,16 +139,49 @@ void ThrowMeaningfulException(Exception? innerException = null)
...
+        }
 
-        public override void EndWrite(IAsyncResult asyncResult) => inner.EndWrite(asyncResult);
+        public override int EndRead(IAsyncResult asyncResult)
+        {
+            try
+            {
+                return ((Task<int>)asyncResult).Result;
+            }
+            catch (AggregateException e) when (e.InnerExceptions.Count == 1 && e.InnerException is not null)
```
sburmanoctopus, line 157: It always comes back as an `AggregateException` (even though it's just 1). So this helps ensure we throw the correct exception, rather than having the caller need to unwrap it.

source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
```diff
@@ -139,16 +139,49 @@ void ThrowMeaningfulException(Exception? innerException = null)
...
 
-        public override IAsyncResult BeginRead(byte[] buffer, int offset, int count, AsyncCallback? callback, object? state) => inner.BeginRead(buffer, offset, count, callback, state);
+        public override void EndWrite(IAsyncResult asyncResult)
+        {
+            var task = (Task)asyncResult;
 
-        public override IAsyncResult BeginWrite(byte[] buffer, int offset, int count, AsyncCallback? callback, object? state) => inner.BeginWrite(buffer, offset, count, callback, state);
+            try
+            {
+                if (task.Exception is not null)
```
sburmanoctopus, line 175: I couldn't find a nicer way of doing this. I'm sure it's out there though.

source/Halibut.Tests/Support/TestAttributes/StreamMethodAttribute.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            public static IEnumerable GetEnumerator(bool testSync)
+            {
+                if (testSync)
+                {
+                    yield return StreamMethod.Sync;
+                }
+                
+                yield return StreamMethod.Async;
+                yield return StreamMethod.LegacyAsync;
```
LukeButters, line 30: nit: It might be worth a comment on this or link to what this is referringto.

sburmanoctopus: Done!

source/Halibut/Transport/TcpConnectionFactory.cs
sburmanoctopus: After a chat, we decided to do it. Mainly because it means that the subsequent SSL methods will also respect the timeouts.

source/Halibut/Transport/TcpConnectionFactory.cs
```diff
@@ -52,16 +54,19 @@ public async Task<IConnection> EstablishNewConnectionAsync(ExchangeProtocolBuild
...
+            
+            var networkStream = client.GetStream();
+            var networkTimeoutStream = new NetworkTimeoutStream(networkStream);
+            var ssl = new SslStream(networkTimeoutStream, false, certificateValidator.Validate, UserCertificateSelectionCallback);
 
             log.Write(EventType.SecurityNegotiation, "Performing TLS handshake");
-            var ssl = new SslStream(stream, false, certificateValidator.Validate, UserCertificateSelectionCallback);
-            
-            // TODO - ASYNC ME UP!
-            // This should take a cancellation token.
```
sburmanoctopus, line 62: Hey @LukeButters, I was just reflecting on this comment.

Technically, we didn't do this for .NET 4.8.

.NET 6.0 will do it, as we are calling the override with the cancellation token.

For .NET 4.8, we now respect the timeouts of the underlying steam. So it's an improvement. But there is no cancellation token being passed into `AuthenticateAsClientAsync` (so nowhere to hook into).

The consequence is that if we cancel this operation, then `AuthenticateAsClientAsync` will ignore that. A cancellation exception will get thrown on the very next line though. So probably not a big deal.

If we were going to handle this, maybe we'd have to pass the cancellation token into the constructor of `NetworkTimeoutStream`, which it can use for all its internal methods. This is a bit yuck, as it means we lock in the cancellation token of the calling method, even if the stream is returned and used outside. But it would work in this case.

I'd lean towards leaving it as is. But wanted to see what you thought.

LukeButters: Just mark it as async me up and add a task to the main convert ticket. Don't forget to add a comment around net48 does not support cancellation tokens here.

LukeButters (approved):
LGTM

source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
LukeButters: Was this caught by the tests?

source/Halibut/Transport/TcpConnectionFactory.cs
sburmanoctopus: Done

source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
sburmanoctopus: Yep!

PR #414, opened 2023-08-10
Make WebSocketStream async

acodrington:
This PR makes `WebSocketStream` async, as part of [sc-53211].

* Implement `CopyToAsync`, `WriteAsync`, `ReadAsync`
* Maintain existing sync behviour (i.e. blocking on async implementations)
* Remove timeout logic, reimplement using timeout and cancellation logic from `NetworkTimeoutStream` which has been extracted and generalised into a reusable helper
* New async implementation of `ReadTextMessage` adds a new timeout argument, so as to not use a static timeout value, as per [sc-45207](https://app.shortcut.com/octopusdeploy/story/45207/improve-halibut-to-support-testing-with-different-tcp-timeouts-allowing-tests-to-control-timeouts). See also [PR comment](https://github.com/OctopusDeploy/Halibut/pull/427#discussion_r1294058105)

### DIFF
--- a/source/Halibut.Tests/Transport/Streams/NetworkTimeoutStreamFixture.cs
+++ b/source/Halibut.Tests/Transport/Streams/NetworkTimeoutStreamFixture.cs
@@ -83,7 +83,7 @@ namespace Halibut.Tests.Transport.Streams
                     stopWatch.Stop();
 
                     actualException.Should().NotBeNull().And.BeOfType<OperationCanceledException>();
-                    actualException!.Message.Should().Be("The ReadAsync operation was cancelled.");
+                    actualException!.Message.Should().Be($"The {nameof(NetworkTimeoutStream)}.ReadAsync operation was cancelled.");
 
                     stopWatch.Elapsed.Should().BeLessThan(TimeSpan.FromSeconds(10));
                 }
@@ -183,7 +183,7 @@ namespace Halibut.Tests.Transport.Streams
                     stopWatch.Stop();
 
                     actualException.Should().NotBeNull().And.BeOfType<OperationCanceledException>();
-                    actualException!.Message.Should().Be("The WriteAsync operation was cancelled.");
+                    actualException!.Message.Should().Be($"The {nameof(NetworkTimeoutStream)}.WriteAsync operation was cancelled.");
 
                     stopWatch.Elapsed.Should().BeLessThan(TimeSpan.FromSeconds(10));
                 }
--- a/source/Halibut.Tests/Transport/Streams/NetworkTimeoutStreamFixture.cs
+++ b/source/Halibut.Tests/Transport/Streams/NetworkTimeoutStreamFixture.cs
@@ -83,7 +83,7 @@ namespace Halibut.Tests.Transport.Streams
                     stopWatch.Stop();
 
                     actualException.Should().NotBeNull().And.BeOfType<OperationCanceledException>();
-                    actualException!.Message.Should().Be($"The {nameof(NetworkTimeoutStream)}.ReadAsync operation was cancelled.");
+                    actualException!.Message.Should().Be("The ReadAsync operation was cancelled.");
 
                     stopWatch.Elapsed.Should().BeLessThan(TimeSpan.FromSeconds(10));
                 }
@@ -183,7 +183,7 @@ namespace Halibut.Tests.Transport.Streams
                     stopWatch.Stop();
 
                     actualException.Should().NotBeNull().And.BeOfType<OperationCanceledException>();
-                    actualException!.Message.Should().Be($"The {nameof(NetworkTimeoutStream)}.WriteAsync operation was cancelled.");
+                    actualException!.Message.Should().Be("The WriteAsync operation was cancelled.");
 
                     stopWatch.Elapsed.Should().BeLessThan(TimeSpan.FromSeconds(10));
                 }
--- a/source/Halibut/Transport/Protocol/WebSocketStream.cs
+++ b/source/Halibut/Transport/Protocol/WebSocketStream.cs
@@ -13,8 +13,8 @@ namespace Halibut.Transport.Protocol
     public class WebSocketStream : Stream
     {
         readonly WebSocket context;
-
         bool isDisposed;
+        readonly CancellationTokenSource cancel = new CancellationTokenSource();
 
         static readonly TimeSpan SendCancelTimeout = TimeSpan.FromSeconds(1);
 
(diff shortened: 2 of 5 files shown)


source/Halibut/Transport/SecureWebSocketListener.cs
```diff
@@ -127,7 +127,7 @@ async Task ExecuteRequest(HttpListenerContext listenerContext)
                 var webSocketContext = await listenerContext.AcceptWebSocketAsync("Octopus").ConfigureAwait(false);
                 webSocketStream = new WebSocketStream(webSocketContext.WebSocket);
 
-                var req = await webSocketStream.ReadTextMessage().ConfigureAwait(false); // Initial message
+                var req = await webSocketStream.ReadTextMessage(CancellationToken.None).ConfigureAwait(false); // Initial message
```
acodrington, line 130: `ReadTestMessage` used to use a token internal to `WebSocketStream`, which only got cancelled when the whole object was disposed. As this could never happen while executing `ReadTestMessage`, it effectively does not use a CancellationToken.

I updated the signature to allow a CancellationToken to be specified in future usages, but to keep the behaviour the same for this usage, I use `CancellationToken.None`.

Let me know if you think this is actually a behaviour change.

source/Halibut/Transport/Protocol/WebSocketStream.cs
```diff
@@ -37,47 +40,74 @@ public override void SetLength(long value)
...
+        public override async Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            AssertCanReadOrWrite();
+            var segment = new ArraySegment<byte>(buffer, offset, count);
+            var receiveResult = await context.ReceiveAsync(segment, cancellationToken);
             return receiveResult.Count;
         }
 
-        public async Task<string> ReadTextMessage()
+        public async Task<string> ReadTextMessage(CancellationToken cancellationToken)
```
nathanwoctopusdeploy, line 55: This method should not be changed. We only want to  change the implementation if Async Halibut is enabled

source/Halibut/Transport/SecureWebSocketListener.cs
nathanwoctopusdeploy: We don't want to change the behaviour if Async Halibut is not enabled

source/Halibut.Tests/Transport/Streams/NetworkTimeoutStreamFixture.cs
```diff
@@ -187,7 +187,7 @@ public async Task WriteAsyncShouldCancel()
                     stopWatch.Stop();
 
                     actualException.Should().NotBeNull().And.BeOfType<OperationCanceledException>();
-                    actualException!.Message.Should().Be("The WriteAsync operation was cancelled.");
+                    actualException!.Message.Should().Be($"The {nameof(NetworkTimeoutStream)}.WriteAsync operation was cancelled.");
```
nathanwoctopusdeploy, line 190: I don't know if we want this change. Out stream is the thing forcing the cancel but it's a RadAsync operation on a different stream that is doing the work. I don't know if it makes sense to name the class which happened to have the timeout logic wrapper.

source/Halibut/Transport/Protocol/WebSocketStream.cs
```diff
@@ -37,47 +40,74 @@ public override void SetLength(long value)
         public override int Read(byte[] buffer, int offset, int count)
         {
             AssertCanReadOrWrite();
-            var segment = new ArraySegment<byte>(buffer, offset, count);
-            var receiveResult = context.ReceiveAsync(segment, CancellationToken.None)
+            return ReadAsync(buffer, offset, count, CancellationToken.None)
```
nathanwoctopusdeploy, line 43: Just avoid changing the existing implementations, but instead add the Async or 'Really'Async methods

source/Halibut/Transport/Protocol/WebSocketStream.cs
nathanwoctopusdeploy: So here I would add a new method `public async Task<string> ReadTextMessageAsync(CancellationToken cancellationToken)` that has your new implementation and leave the old one alone

source/Halibut/Transport/Streams/CancellationAndTimeoutTaskWrapper.cs
```diff
@@ -0,0 +1,63 @@
...
+            void ThrowMeaningfulException(Exception? innerException = null)
+            {
+                if (timeoutCancellationTokenSource.IsCancellationRequested)
+                {
+                    throw getExceptionOnTimeout();
+                }
+
+                if (cancellationToken.IsCancellationRequested)
+                {
+                    throw new OperationCanceledException($"The {callingTypeName}.{methodName} operation was cancelled.", innerException);
```
nathanwoctopusdeploy, line 58: Consider dropping the `{callingTypeName}.` as does it make sense to name the wrapper stream that happened to implement cancellation

source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
```diff
@@ -2,10 +2,9 @@
 using System;
 using System.IO;
 using System.Net.Sockets;
-using System.Runtime.Remoting;
+using System.Runtime.Remoting; // Used in NETFRAMEWORK pragma
```
nathanwoctopusdeploy, line 5: If that // needed?

source/Halibut.Tests/Transport/Streams/NetworkTimeoutStreamFixture.cs
acodrington: The motivation was that I didn't want to lose context - we previously would be able to trace this error back to `NetworkTimeoutStream`, but generalising means the call stack wouldn't be as clear as there are now multiple usages, so I thought the caller class name in the error might help.

I'm happy to remove it though, it was never strictly necessary anyway.

source/Halibut/Transport/Streams/CancellationAndTimeoutTaskWrapper.cs
acodrington: As per above comment, I'll drop this field. 👍🏻

source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
acodrington: Needed? No. I thought it might be helpful as a reminder in the future, as I cleared out this 'unused' `using` statement, and then wondered why it wouldn't compile.

Maybe I'll just be more careful next time I see pre-processor directives, and then I won't need this! 😄

nathanwoctopusdeploy:
[sc-57122]

LukeButters (approved):

PR #422, opened 2023-08-11
Test TryGetAvailableInputBufferSizeShouldReadSize now tests async calls

LukeButters:
# Background

[SC-54458]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Support/TestAttributes/SyncAndAsyncAttribute.cs
+++ b/source/Halibut.Tests/Support/TestAttributes/SyncAndAsyncAttribute.cs
@@ -1,6 +1,9 @@
 ﻿using System;
 using System.Collections;
+using System.IO;
+using System.Threading;
 using System.Threading.Tasks;
+using Halibut.Transport.Streams;
 using Halibut.Util;
 using NUnit.Framework;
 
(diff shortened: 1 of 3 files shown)


source/Halibut.Tests/Support/TestAttributes/SyncAndAsyncAttribute.cs
```diff
@@ -93,6 +96,18 @@ public static AsyncHalibutFeature ToAsyncHalibutFeature(this SyncOrAsync syncOrA
                 _ => throw new ArgumentOutOfRangeException(nameof(syncOrAsync), syncOrAsync, null)
             };
         }
+
+        public static async Task WriteToStream(this SyncOrAsync syncOrAsync, Stream stream, byte[] bytes)
```
LukeButters, line 100: Do these make it worse or better?

source/Halibut/Transport/Streams/StreamExtensionMethods.cs
```diff
@@ -65,6 +65,17 @@ public static async Task WriteByteArrayAsync(this Stream stream, byte[] bytes, C
             await stream.WriteAsync(bytes, 0, bytes.Length, cancellationToken);
         }
         
+        /// <summary>
+        /// Net48 does not have a method which just writes all bytes of an array to a stream.
+        /// </summary>
+        /// <param name="stream"></param>
+        /// <param name="bytes"></param>
+        /// <param name="cancellationToken"></param>
+        public static void WriteByteArray(this Stream stream, byte[] bytes)
```
sburmanoctopus, line 74: As we aren't using this in the Halibut itself, I'd be inclined to move it to the test project for now.

source/Halibut.Tests/Support/TestAttributes/SyncAndAsyncAttribute.cs
sburmanoctopus: If it was used in multiple test classes, I'd say better for sure.

For now, it's only used in 1, so the benefit isn't really there. But if we foresee more usages of this (as streams are common), this can turn into useful easily.

sburmanoctopus (approved):
LGTM.

PR #424, opened 2023-08-14
PoC - Trust all certs issued by trusted root authority - Do Not Merge

pawelpabich:
# Background

Some environments require all traffic to be decrypted and inspected by a centralized proxy.  This is usually done by implementing [MITM](https://en.wikipedia.org/wiki/Man-in-the-middle_attack), where the proxy becomes a trusted Root Authority and the clients trust all certificates issued by it.  Halibut security is based on [certificate pinning](https://security.stackexchange.com/questions/29988/what-is-certificate-pinning), which breaks MITM.

This PoC proposes to support the first leg of MITM (halibut -> proxy) by adding an option to Halibut to allow it to trust any cert issued by a trusted root authority.

### DIFF
--- a/source/Halibut.Tests/TlsInsepctionTests.cs
+++ b/source/Halibut.Tests/TlsInsepctionTests.cs
@@ -1,0 +1,49 @@
+using System;
+using FluentAssertions;
+using Halibut.ServiceModel;
+using Halibut.Tests.Support;
+using Halibut.TestUtils.Contracts;
+using NUnit.Framework;
+
+namespace Halibut.Tests
+{
+    public class TlsInspectionTests : BaseTest
+    {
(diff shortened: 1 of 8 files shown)


jimmyp:
@pawelpabich I found [this](https://octopus.com/docs/security/octopus-tentacle-communication/custom-certificates-with-octopus-server-and-tentacle) in our docs. Is the reason that configuring custom certificates isn't enough to facilitate MITM by configuring tentacle and server to trust the certs presented by the proxy, that the proxies generate new certificates often?

pawelpabich:
Conversation has moved to https://octopusdeploy.slack.com/archives/C075APRMJP5/p1716981406294229

PR #425, opened 2023-08-14
Connection pool async

sburmanoctopus:
[sc-53211]

# Background
While converting Halibut over to async, we found that the connection manager and connection pool disposes connections while taking new ones. These are async operations, and as such, should be made async.

# Results

## Before
Connections were being disposed of synchronously. This is IO that would block any thread while disposable occurred.

## After
All methods that result in disposal of a connection within the `ConnectionPool` now has an async equivalent.

In the `ConnectionManager`, moving from a `lock` to a `SemaphoreSlim` introduced a deadlock. This all revolved around the lock in `OnConnectionDisposed`. This was focused around access to the `activeConnections`, so we rearranged the locks to ensure access to `activeConnections` would not result in a deadlock, and introduced a new lock for general access to the connections in general.

### ConnectionManagerFixture
The `ConnectionManagerFixture` tests had a very convoluted setup, with a mixture of mock and real objects. The real objects did not actually do anything related to the test. This included the `MessageExchangeProtocol` created by `GetProtocol`.  To show how unimportant these 'real' connections were, there was a bug where the async version of `EstablishNewConnection` was never configured, and all async tests were working fine with an NSubstitute version. We also needed to interrogate the connection, so this whole area was simplified.

To help identify the deadlocks mentioned above, extra test cases were added to `ConnectionManagerFixture` that caused the deadlock when using `SemaphoreSlim` as a pure replacement.

## Later
Some more work has to be done in regards to async disposal. But that will be done in the async disposal story.

# How to review this PR

`RequestCancellationTokens.InProgressRequestCancellationToken` was introduced while this was being worked on. I believe this is the correct cancellation token to use (as it is used in the same methods). But please verify that my choice was correct 😁 

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Transport/ConnectionManagerAsync.cs
+++ b/source/Halibut/Transport/ConnectionManagerAsync.cs
@@ -13,21 +13,17 @@ namespace Halibut.Transport
     {
         readonly ConnectionPoolAsync<ServiceEndPoint, IConnection> pool = new();
         readonly Dictionary<ServiceEndPoint, HashSet<IConnection>> activeConnections = new();
-        readonly HalibutTimeoutsAndLimits halibutTimeoutsAndLimits;
 
         // We have separate locks for connections in general (including the pool) vs specifically activeConnections.
         // This is because disposing calls OnConnectionDisposed, which causes deadlocks if we are not careful.
+        // I.e. use connectionsLock when interacting with anything to do with connections (e.g. the pool, or activeConnections),
+        //      but use activeConnectionsLock AND connectionsLock when specifically interacting with activeConnections.
         readonly SemaphoreSlim connectionsLock = new(1, 1);
         readonly SemaphoreSlim activeConnectionsLock = new(1, 1);
         
 
         public bool IsDisposed { get; private set; }
-
-        public ConnectionManagerAsync(HalibutTimeoutsAndLimits halibutTimeoutsAndLimits)
-        {
-            this.halibutTimeoutsAndLimits = halibutTimeoutsAndLimits;
-        }
-
+        
         public void Dispose()
         {
             pool.Dispose();
@@ -161,6 +157,7 @@ namespace Halibut.Transport
             }
         }
 
+        [Obsolete]
         public void ClearPooledConnections(ServiceEndPoint serviceEndPoint, ILog log)
         {
             using (connectionsLock.Lock())
@@ -193,6 +190,7 @@ namespace Halibut.Transport
             return Array.Empty<IConnection>();
         }
 
+        [Obsolete]
         public void Disconnect(ServiceEndPoint serviceEndPoint, ILog log)
         {
             using (connectionsLock.Lock())
@@ -237,6 +235,8 @@ namespace Halibut.Transport
         
         void OnConnectionDisposed(IConnection connection)
         {
+            //TODO: Async version
+
             // If we are not careful, we can introduce a deadlock. Time this out just in case we ever accidentally introduce one.
             using var cts = new CancellationTokenSource(TimeSpan.FromMinutes(1));
             using (activeConnectionsLock.Lock(cts.Token))
--- a/source/Halibut/Transport/ConnectionManagerAsync.cs
+++ b/source/Halibut/Transport/ConnectionManagerAsync.cs
@@ -13,6 +13,7 @@ namespace Halibut.Transport
     {
         readonly ConnectionPoolAsync<ServiceEndPoint, IConnection> pool = new();
         readonly Dictionary<ServiceEndPoint, HashSet<IConnection>> activeConnections = new();
+        readonly HalibutTimeoutsAndLimits halibutTimeoutsAndLimits;
 
         // We have separate locks for connections in general (including the pool) vs specifically activeConnections.
         // This is because disposing calls OnConnectionDisposed, which causes deadlocks if we are not careful.
@@ -21,6 +22,11 @@ namespace Halibut.Transport
         
 
         public bool IsDisposed { get; private set; }
+
+        public ConnectionManagerAsync(HalibutTimeoutsAndLimits halibutTimeoutsAndLimits)
+        {
+            this.halibutTimeoutsAndLimits = halibutTimeoutsAndLimits;
+        }
 
         public void Dispose()
         {
--- a/source/Halibut/Transport/ConnectionManagerAsync.cs
+++ b/source/Halibut/Transport/ConnectionManagerAsync.cs
@@ -25,15 +25,18 @@ namespace Halibut.Transport
         public void Dispose()
         {
             pool.Dispose();
-            IConnection[] connectionsToDispose;
-            using (activeConnectionsLock.Lock())
+            using (connectionsLock.Lock())
             {
-                connectionsToDispose = activeConnections.SelectMany(kv => kv.Value).ToArray();
-            }
+                IConnection[] connectionsToDispose;
+                using (activeConnectionsLock.Lock())
+                {
+                    connectionsToDispose = activeConnections.SelectMany(kv => kv.Value).ToArray();
+                }
 
-            foreach (var connection in connectionsToDispose)
-            {
-                SafelyDisposeConnection(connection, null);
+                foreach (var connection in connectionsToDispose)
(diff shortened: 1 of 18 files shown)


source/Halibut/Transport/ConnectionManagerAsync.cs
```diff
@@ -0,0 +1,297 @@
...
+
+        public async Task ClearPooledConnectionsAsync(ServiceEndPoint serviceEndPoint, ILog log, CancellationToken cancellationToken)
+        {
+            using (await connectionsLock.LockAsync(cancellationToken))
+            {
+                await pool.ClearAsync(serviceEndPoint, log, cancellationToken);
+            }
+        }
+        
+        public IReadOnlyCollection<IConnection> GetActiveConnections(ServiceEndPoint serviceEndPoint)
```
sburmanoctopus, line 171: I didn't make an async version of this, as it's only used in tests.

source/Halibut/Transport/ConnectionPoolAsync.cs
```diff
@@ -0,0 +1,188 @@
...
+
+namespace Halibut.Transport
+{
+    public class ConnectionPoolAsync<TKey, TPooledResource> : IConnectionPool<TKey, TPooledResource> 
+        where TPooledResource : class, IPooledResource
+    {
+        readonly Dictionary<TKey, HashSet<TPooledResource>> pool = new();
+        readonly SemaphoreSlim poolLock = new(1, 1);
+
+        public int GetTotalConnectionCount()
```
sburmanoctopus, line 17: There is no async version of this as it's only used in tests

source/Halibut/Transport/IConnectionPool.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Halibut.Transport
+{
+    public interface IConnectionPool<TKey, TPooledResource> where TPooledResource : class, IPooledResource
+    {
+        int GetTotalConnectionCount();
+        TPooledResource Take(TKey endPoint);
+        Task<TPooledResource> TakeAsync(TKey endPoint, CancellationToken cancellationToken);
+        void Return(TKey endPoint, TPooledResource resource);
+        Task ReturnAsync(TKey endPoint, TPooledResource resource, CancellationToken cancellationToken);
+        void Clear(TKey key, ILog log = null);
```
LukeButters, line 15: Can these sync methods be marked as `[Obsolete]`?

source/Halibut/Transport/IConnectionManager.cs
```diff
@@ -0,0 +1,23 @@
...
+using Halibut.Transport.Protocol;
+
+namespace Halibut.Transport
+{
+    public interface IConnectionManager : IDisposable
+    {
+        bool IsDisposed { get; }
+        IConnection AcquireConnection(ExchangeProtocolBuilder exchangeProtocolBuilder, IConnectionFactory connectionFactory, ServiceEndPoint serviceEndpoint, ILog log, CancellationToken cancellationToken);
+        Task<IConnection> AcquireConnectionAsync(ExchangeProtocolBuilder exchangeProtocolBuilder, IConnectionFactory connectionFactory, ServiceEndPoint serviceEndpoint, ILog log, CancellationToken cancellationToken);
+        void ReleaseConnection(ServiceEndPoint serviceEndpoint, IConnection connection);
```
LukeButters, line 15: Can these sync methods be marked as `[obsolete]`

source/Halibut/Transport/ConnectionManagerAsync.cs
```diff
@@ -0,0 +1,300 @@
...
+
+namespace Halibut.Transport
+{
+    public class ConnectionManagerAsync : IConnectionManager
+    {
+        readonly ConnectionPoolAsync<ServiceEndPoint, IConnection> pool = new();
+        readonly Dictionary<ServiceEndPoint, HashSet<IConnection>> activeConnections = new();
+
+        // We have separate locks for connections in general (including the pool) vs specifically activeConnections.
+        // This is because disposing calls OnConnectionDisposed, which causes deadlocks if we are not careful.
```
LukeButters, line 18: I think it would make sense to describe here how to use these locks, when do I take one out over the other.

source/Halibut/Transport/ConnectionManagerAsync.cs
```diff
@@ -0,0 +1,300 @@
...
+                }
+            }
+
+            IsDisposed = true;
+        }
+
+        [Obsolete]
+        public IConnection AcquireConnection(ExchangeProtocolBuilder exchangeProtocolBuilder, IConnectionFactory connectionFactory, ServiceEndPoint serviceEndpoint, ILog log, CancellationToken cancellationToken)
+        {
+            Tuple<IConnection, Action> openableConnection;
```
LukeButters, line 48: Should the sync paths just throw? I guess perhaps it is safer if they work at least to some extent.

The tests do pass if these were not implemented right?

source/Halibut/Transport/ConnectionManagerAsync.cs
```diff
@@ -0,0 +1,300 @@
...
+        public void ReleaseConnection(ServiceEndPoint serviceEndpoint, IConnection connection)
+        {
+            using (connectionsLock.Lock())
+            {
+                pool.Return(serviceEndpoint, connection);
+                using (activeConnectionsLock.Lock())
+                {
+                    if (activeConnections.TryGetValue(serviceEndpoint, out var connections))
+                    {
+                        connections.Remove(connection);
```
LukeButters, line 137: I wonder if a set of "activeConnection" methods could be made that take locks for the caller that way method only need to worry about taking connectionLocks.

source/Halibut/Transport/ConnectionManagerAsync.cs
```diff
@@ -0,0 +1,300 @@
...
+                }
+
+                foreach (var connection in toDelete)
+                {
+                    SafelyDisposeConnection(connection, log);
+                }
+            }
+        }
+        
+        void OnConnectionDisposed(IConnection connection)
```
LukeButters, line 232: Is an async version of this coming? could it be marked as TODO?

LukeButters (approved):

source/Halibut/Transport/ConnectionPoolAsync.cs
sburmanoctopus: Because we test this in a polymorphic way using `SyncOrAsync`, internal isn't possible.

🤔 I've been eyeing off this, so instead what I did is to just delete this method completely, and changed the test to test the behaviour instead of poking at an internal count 😁

source/Halibut/Transport/IConnectionPool.cs
sburmanoctopus: Sure! Done!

sburmanoctopus: I decorated the actual methods as well

source/Halibut/Transport/IConnectionManager.cs
sburmanoctopus: Done!

source/Halibut/Transport/ConnectionManagerAsync.cs
sburmanoctopus: Done

source/Halibut/Transport/ConnectionManagerAsync.cs
sburmanoctopus: Like you said, I kept it because of the gradual transition to async.

The connection manager tests pass with those Obsolete methods throwing exceptions.

Until we are fully cutover to async, I'd be nervous making sync methods throw.

I've [created a card](https://app.shortcut.com/octopusdeploy/story/56885/make-sync-methods-in-connectionmanagerasync-and-connectionpoolasync-throw-exceptions) to come back to this.

source/Halibut/Transport/ConnectionManagerAsync.cs
sburmanoctopus: Lol, the async version is already written in the next branch 😁  But I'll add a todo to be safe.

PR #426, opened 2023-08-14
Remove unnecessary initialization of ILog from calls to LogProvider.GetLogger("Halibut")

nathanwoctopusdeploy:
# Background

This PR changes the way `ILog` instances are created which are used by the `InMemoryConnectionLog` to improve a performance issue observed in Halibut.

Snippet of `pstacks` from a Server process dump showing the issue

image

Before this change, an ILog would be constructed for every log message that Halibut wanted to write.

After this change, the LogFactory constructs an ILog which is reused for all InMemoryConnectionLogs and so all log messages halibut writes.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/BadCertificatesTests.cs
+++ b/source/Halibut.Tests/BadCertificatesTests.cs
@@ -13,6 +13,7 @@ using Halibut.Tests.Support.TestCases;
 using Halibut.Tests.TestServices.Async;
 using Halibut.Tests.TestServices.SyncClientWithOptions;
 using Halibut.TestUtils.Contracts;
+using Halibut.Transport.Protocol;
 using NUnit.Framework;
 
 namespace Halibut.Tests
@@ -167,7 +168,7 @@ namespace Halibut.Tests
 
                 countingService.GetCurrentValue().Should().Be(0, "With a bad certificate the request never should have been made");
 
-                serviceLoggers[serviceLoggers.Keys.First()].GetLogs().Should()
+                serviceLoggers[serviceLoggers.Keys.First(x => x != nameof(MessageSerializer))].GetLogs().Should()
                     .Contain(log => log.FormattedMessage
                         .Contains("and attempted a message exchange, but it presented a client certificate with the thumbprint " +
                                   "'76225C0717A16C1D0BA4A7FFA76519D286D8A248' which is not in the list of thumbprints that we trust"));
--- a/source/Halibut.Tests/ParallelRequestsFixture.cs
+++ b/source/Halibut.Tests/ParallelRequestsFixture.cs
@@ -17,6 +17,7 @@ namespace Halibut.Tests
 {
     public static class ParallelRequestsFixture
     {
+        [Parallelizable(ParallelScope.None)]
         public class AsyncParallelRequestFixture : BaseTest
         {
             [Test]
@@ -118,7 +119,7 @@ namespace Halibut.Tests
             }
         }
 
-        [Parallelizable(ParallelScope.Children)]
+        [Parallelizable(ParallelScope.None)]
         public class SyncParallelRequestsFixture : BaseTest
         {
             [Test]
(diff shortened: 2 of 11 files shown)


source/Halibut.Tests/ParallelRequestsFixture.cs
```diff
@@ -118,7 +119,7 @@ public async Task SendMessagesToTentacleInParallelAsync(ClientAndServiceTestCase
             }
         }
 
-        [Parallelizable(ParallelScope.Children)]
+        [Parallelizable(ParallelScope.None)]
```
nathanwoctopusdeploy, line 122: This test was failing locally due to the load it creates

source/Halibut.Tests/ParallelRequestsFixture.cs
```diff
@@ -17,6 +17,7 @@ namespace Halibut.Tests
 {
     public static class ParallelRequestsFixture
     {
+        [Parallelizable(ParallelScope.None)]
```
nathanwoctopusdeploy, line 20: These tests was failing locally due to the load they creates

source/Halibut/Diagnostics/InMemoryConnectionLog.cs
```diff
@@ -9,11 +9,19 @@ namespace Halibut.Diagnostics
...
     {
         readonly string endpoint;
-        readonly ConcurrentQueue<LogEvent> events = new ConcurrentQueue<LogEvent>();
+        readonly Logging.ILog logger;
+        readonly ConcurrentQueue<LogEvent> events = new();
 
         public InMemoryConnectionLog(string endpoint)
         {
             this.endpoint = endpoint;
+            this.logger = LogProvider.GetLogger("Halibut");
```
nathanwoctopusdeploy, line 18: Kept for test code

source/Halibut.Tests/BadCertificatesTests.cs
```diff
@@ -167,7 +168,7 @@ public async Task FailWhenClientPresentsWrongCertificateToListeningService(Clien
 
                 countingService.GetCurrentValue().Should().Be(0, "With a bad certificate the request never should have been made");
 
-                serviceLoggers[serviceLoggers.Keys.First()].GetLogs().Should()
+                serviceLoggers[serviceLoggers.Keys.First(x => x != nameof(MessageSerializer))].GetLogs().Should()
```
nathanwoctopusdeploy, line 171: This test assumed the logs it was interested in were going to be first in the list

source/Halibut.Tests/ParallelRequestsFixture.cs
LukeButters: Did other tests also fail?

LukeButters: I am not sure this is the right path, could you look into why they are failing? is it a real problem perhaps?

Is it this test that actually causes a high load? Or is it the number of concurrent tests running. Is the high load itself a problem or is something going wrong because of highload?

source/Halibut/Diagnostics/InMemoryConnectionLog.cs
```diff
@@ -60,7 +68,6 @@ static LogLevel GetLogLevel(LogEvent logEvent)
 
         void SendToTrace(LogEvent logEvent, LogLevel level)
         {
-            var logger = LogProvider.GetLogger("Halibut");
```
LukeButters, line 63: This btw is the performance issue

LukeButters (approved):

source/Halibut.Tests/ParallelRequestsFixture.cs
nathanwoctopusdeploy: Card created to investigate

source/Halibut/Diagnostics/InMemoryConnectionLog.cs
nathanwoctopusdeploy: Ahh yeh, the important change 👍

nathanwoctopusdeploy:
[sc-56658]

PR #427, opened 2023-08-14
Add support for per HalibutRuntime timeouts and limits

LukeButters:
# Background

**Statics make testing hard**

HalibutLimits are statically set, which means all tests are using the same timeout. This is a problem since either we can have:
* High timeouts increasing reliability but with timeout tests taking longer to run.
* Low timeouts reducing the time it takes to run timeout test but with decreased reliability of tests.

This introduces a new type `HalibutTimeoutsAndLimits`, which holds timeouts and limits and is used only in async paths.

The existing `HalibutLimits` class is marked obsolete.

[SC-45207]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Support/LatestClientAndLatestServiceBuilderExtensionMethods.cs
+++ b/source/Halibut.Tests/Support/LatestClientAndLatestServiceBuilderExtensionMethods.cs
@@ -1,4 +1,6 @@
 ﻿using System;
+using Halibut.Tests.Support.TestAttributes;
+using Halibut.Tests.Support.TestCases;
 using Halibut.Tests.TestServices;
 using Halibut.TestUtils.Contracts;
 using Halibut.Util;
@@ -57,6 +59,16 @@ namespace Halibut.Tests.Support
         public static LatestClientAndLatestServiceBuilder WithInstantReconnectPollingRetryPolicy(this LatestClientAndLatestServiceBuilder builder)
         {
             return builder.WithPollingReconnectRetryPolicy(() => new RetryPolicy(1, TimeSpan.Zero, TimeSpan.Zero));
+        }
+
+        public static LatestClientAndLatestServiceBuilder WhenTestingAsyncClient(this LatestClientAndLatestServiceBuilder builder, ClientAndServiceTestCase clientAndServiceTestCase, Action<LatestClientAndLatestServiceBuilder> action)
+        {
+
+            if (clientAndServiceTestCase.SyncOrAsync == SyncOrAsync.Async)
+            {
+                action(builder);
+            }
+            return builder;
         }
     }
 }
--- a/source/Halibut.Tests/Support/TestAttributes/LatestClientAndLatestServiceTestCasesAttribute.cs
+++ b/source/Halibut.Tests/Support/TestAttributes/LatestClientAndLatestServiceTestCasesAttribute.cs
@@ -23,18 +23,19 @@ namespace Halibut.Tests.Support.TestAttributes
             bool testSyncClients = true,
             bool testAsyncClients = true,
             bool testAsyncServicesAsWell = false, // False means only the sync service will be tested.
+            bool testSyncService = true,
             params object[] additionalParameters
             ) :
             base(
                 typeof(LatestClientAndLatestServiceTestCases),
                 nameof(LatestClientAndLatestServiceTestCases.GetEnumerator),
-                new object[] { testWebSocket, testNetworkConditions, testListening, testPolling, testSyncClients, testAsyncClients, testAsyncServicesAsWell, additionalParameters })
+                new object[] { testWebSocket, testNetworkConditions, testListening, testPolling, testSyncClients, testAsyncClients, testAsyncServicesAsWell, testSyncService, additionalParameters })
         {
         }
 
         static class LatestClientAndLatestServiceTestCases
         {
-            public static IEnumerable GetEnumerator(bool testWebSocket, bool testNetworkConditions, bool testListening, bool testPolling, bool testSyncClients, bool testAsyncClients, bool testAsyncServicesAsWell, object[] additionalParameters)
+            public static IEnumerable GetEnumerator(bool testWebSocket, bool testNetworkConditions, bool testListening, bool testPolling, bool testSyncClients, bool testAsyncClients, bool testAsyncServicesAsWell, bool testSyncService, object[] additionalParameters)
             {
                 var serviceConnectionTypes = ServiceConnectionTypes.All.ToList();
 
@@ -69,6 +70,11 @@ namespace Halibut.Tests.Support.TestAttributes
                 if (!testAsyncServicesAsWell)
                 {
                     serviceAsyncHalibutFeatureTestCases.Remove(AsyncHalibutFeature.Enabled);
+                }
+
+                if (!testSyncService)
+                {
+                    serviceAsyncHalibutFeatureTestCases.Remove(AsyncHalibutFeature.Disabled);
                 }
 
                 var builder = new ClientAndServiceTestCasesBuilder(
--- a/source/Halibut.Tests/Timeouts/SendingAndReceivingRequestMessagesTimeoutsFixture.cs
+++ b/source/Halibut.Tests/Timeouts/SendingAndReceivingRequestMessagesTimeoutsFixture.cs
@@ -10,13 +10,58 @@ using Halibut.Tests.Util;
 using Halibut.TestUtils.Contracts;
 using NUnit.Framework;
 using System.Runtime.InteropServices;
+using Halibut.Tests.Support.PortForwarding;
 using Halibut.Tests.Support.TestCases;
 using Halibut.Tests.TestServices.Async;
+using Halibut.Util;
 
 namespace Halibut.Tests.Timeouts
 {
     public class SendingAndReceivingRequestMessagesTimeoutsFixture : BaseTest
     {
+        [Test]
+        [LatestClientAndLatestServiceTestCases(testNetworkConditions: false, testWebSocket: false, testAsyncServicesAsWell: true, testSyncService : false, testSyncClients: false, testAsyncClients: true)]
+        public async Task HalibutTimeoutsAndLimits_AppliesToTcpClientReceiveTimeout(ClientAndServiceTestCase clientAndServiceTestCase)
+        {
+            var expectedTimeout = clientAndServiceTestCase.SyncOrAsync == SyncOrAsync.Sync ? HalibutLimits.TcpClientReceiveTimeout : TimeSpan.FromSeconds(10);
+            using (var clientAndService = await clientAndServiceTestCase.CreateTestCaseBuilder()
+                       .As<LatestClientAndLatestServiceBuilder>()
+                       .WithPortForwarding(port => PortForwarderUtil.ForwardingToLocalPort(port).WithPortForwarderDataLogging(clientAndServiceTestCase.ServiceConnectionType).Build())
+                       .WithPortForwarding(out var portForwarderRef)
+                       .WithEchoService()
+                       .WithDoSomeActionService(() => portForwarderRef.Value.PauseExistingConnections())
+                       .WhenTestingAsyncClient(clientAndServiceTestCase, b =>
+                       {
+                           b.WithHalibutTimeoutsAndLimits(new HalibutTimeoutsAndLimits()
+                               .SetAllTcpTimeoutsTo(TimeSpan.FromSeconds(133))
+                               .WithTcpClientReceiveTimeout(expectedTimeout));
+                       })
+                       .WithInstantReconnectPollingRetryPolicy()
+                       .Build(CancellationToken))
+            {
+                var echo = clientAndService.CreateClient<IEchoService, IAsyncClientEchoService>();
+                await echo.SayHelloAsync("Make a request to make sure the connection is running, and ready. Lets not measure SSL setup cost.");
+
+                var pauseConnections = clientAndService.CreateClient<IDoSomeActionService, IAsyncClientDoSomeActionService>(IncreasePollingQueueTimeout());
+
+                var sw = Stopwatch.StartNew();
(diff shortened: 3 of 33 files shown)


source/Halibut.Tests/Support/LatestClientAndLatestServiceBuilderExtensionMethods.cs
```diff
@@ -58,5 +60,15 @@ public static LatestClientAndLatestServiceBuilder WithInstantReconnectPollingRet
         {
             return builder.WithPollingReconnectRetryPolicy(() => new RetryPolicy(1, TimeSpan.Zero, TimeSpan.Zero));
         }
+
+        public static LatestClientAndLatestServiceBuilder WhenTestingAsyncClient(this LatestClientAndLatestServiceBuilder builder, ClientAndServiceTestCase clientAndServiceTestCase, Action<LatestClientAndLatestServiceBuilder> action)
```
LukeButters, line 64: Some of the existing timeout tests could be made faster by making use of this and lowering timeouts.

source/Halibut.Tests/Support/TestAttributes/LatestClientAndLatestServiceTestCasesAttribute.cs
```diff
@@ -23,18 +23,19 @@ public class LatestClientAndLatestServiceTestCasesAttribute : TestCaseSourceAttr
             bool testSyncClients = true,
             bool testAsyncClients = true,
             bool testAsyncServicesAsWell = false, // False means only the sync service will be tested.
+            bool testSyncService = true,
```
LukeButters, line 26: I wonder if these could be in a `#define` and so defined once but used here and in the below static method. As well as in other builders.

source/Halibut.Tests/Timeouts/SendingAndReceivingRequestMessagesTimeoutsFixture.cs
```diff
@@ -10,13 +10,57 @@
...
+                var sw = Stopwatch.StartNew();
+                var e = (await AssertAsync.Throws<HalibutClientException>(async () => await pauseConnections.ActionAsync())).And;
+                sw.Stop();
+                Logger.Error(e, "Received error");
+                AssertExceptionMessageLooksLikeAReadTimeout(e);
+                sw.Elapsed.Should().BeGreaterThan(expectedTimeout - TimeSpan.FromSeconds(2), "The receive timeout should apply, not the shorter heart beat timeout") // -2s give it a little slack to avoid it timed out slightly too early.
+                    .And
+                    .BeLessThan(expectedTimeout + LowerHalibutLimitsForAllTests.HalfTheTcpReceiveTimeout, "We should be timing out on the tcp receive timeout");
+                
+                // The polling tentacle, will not reconnect in time since it has a 133s receive control message timeout.
```
LukeButters, line 55: The raises other questions like if the tentacle has sent a large response back, will it be able to move into a control message read/write but timeout on that. Not because the next/proceed took a long time itself but because the data ahead of it resulted in the control messages taking some time.

source/Halibut/ServiceEndPoint.cs
```diff
@@ -8,11 +8,15 @@ public class ServiceEndPoint : IEquatable<ServiceEndPoint>
     {
         readonly string baseUriString;
 
+        // TODO - ASYNC ME UP!
```
LukeButters, line 11: In another PR we will need a ServiceEndPoint constructed with the halibuttimeoutsandlimits. Or perhaps the halibut runtime should have a ServiceEndPoint factory with timeouts configured from its limits

source/Halibut/Transport/Protocol/ControlMessageReader.cs
```diff
@@ -151,5 +159,15 @@ static CancellationTokenSource GetCancellationTokenSourceFromStreamReadTimeout(S
 
             return new CancellationTokenSource(HalibutLimits.TcpClientReceiveTimeout); // Just default to a higher timeout, rather than be cancellation token none
         }
+        
+        CancellationTokenSource GetCancellationTokenSourceFromStreamReadTimeoutAsync(Stream stream)
+        {
+            if (stream.CanTimeout)
```
LukeButters, line 165: TODO mark this as TODO to note we should always be given a stream that can timeout.

source/Halibut.Tests/Timeouts/SendingAndReceivingRequestMessagesTimeoutsFixture.cs
```diff
@@ -10,13 +10,57 @@
...
 using Halibut.Tests.Support.TestCases;
 using Halibut.Tests.TestServices.Async;
 
 namespace Halibut.Tests.Timeouts
 {
     public class SendingAndReceivingRequestMessagesTimeoutsFixture : BaseTest
     {
+        [Test]
+        [LatestClientAndLatestServiceTestCases(testNetworkConditions: false, testWebSocket: false, testAsyncServicesAsWell: true, testSyncService : false, testSyncClients: false, testAsyncClients: true)]
+        public async Task HalibutTimeoutsAndLimits_ApplyInAsync(ClientAndServiceTestCase clientAndServiceTestCase)
```
nathanwoctopusdeploy, line 23: This is testing 1 specific timeout, can the test name be changed so it is clear it's not covering all tiemouts

source/Halibut.Tests/Timeouts/SendingAndReceivingRequestMessagesTimeoutsFixture.cs
nathanwoctopusdeploy: Is this something we need to look at? If so should we card it?

source/Halibut.Tests/Transport/DiscoveryClientFixture.cs
```diff
@@ -44,7 +44,7 @@ public async Task DiscoveringNonExistentEndpointThrows(SyncOrAsync syncOrAsync)
 #pragma warning disable CS0612
             await syncOrAsync
                 .WhenSync(() => Assert.Throws<HalibutClientException>(() => client.Discover(fakeEndpoint), "No such host is known")).IgnoreResult()
-                .WhenAsync(async () => await AssertAsync.Throws<HalibutClientException>(() => client.DiscoverAsync(fakeEndpoint, CancellationToken), "No such host is known"));
+                .WhenAsync(async () => await AssertAsync.Throws<HalibutClientException>(() => client.DiscoverAsync(fakeEndpoint, new HalibutTimeoutsAndLimits(), CancellationToken), "No such host is known"));
```
nathanwoctopusdeploy, line 47: Do we want a helper that sets some good defaults instead of `new HalibutTimeoutsAndLimits()` so it's consistent with the statics we set at the moment to keep some consistency amount how timeouts are set. Individual tests can choose to change them, otherwise they all get the same.

nathanwoctopusdeploy: Ahh, I see they are defaulted from the statics so it is doing that 👍

source/Halibut/Diagnostics/HalibutLimits.cs
```diff
@@ -107,4 +108,102 @@ public static TimeSpan SafeTcpClientPooledConnectionTimeout
...
+        /// </summary>
+        public TimeSpan TcpClientReceiveTimeout { get; set; } =  HalibutLimits.TcpClientReceiveTimeout;
+        
+        /// <summary>
+        /// Amount of time a connection can stay in the pool
+        /// </summary>
+        public TimeSpan TcpClientPooledConnectionTimeout { get; set; } =  HalibutLimits.TcpClientPooledConnectionTimeout;
+        
+        public TimeSpan TcpClientHeartbeatSendTimeout { get; set; } =  HalibutLimits.TcpClientHeartbeatSendTimeout;
+        public TimeSpan TcpClientHeartbeatReceiveTimeout { get; set; } =  HalibutLimits.TcpClientHeartbeatReceiveTimeout;
```
nathanwoctopusdeploy, line 167: nit: double space, single spaces, no spaces!!

source/Halibut/Diagnostics/HalibutLimits.cs
```diff
@@ -107,4 +108,102 @@ public static TimeSpan SafeTcpClientPooledConnectionTimeout
...
+                }
+                else
+                {
+                    var timeout = TcpClientReceiveTimeout - TimeSpan.FromSeconds(10);
+                    return timeout > TimeSpan.Zero ? timeout : TcpClientReceiveTimeout;
+                }
+            }
+        }
+
+        public HalibutTimeoutsAndLimits WithTcpClientReceiveTimeout(TimeSpan tcpClientReceiveTimeout)
```
nathanwoctopusdeploy, line 203: Why is this one special? All the properties have public set methods so it this needed? Didi you mean to remove the set methods on properties.

Would this be better suited as an extension method if your trying to create a bit of a fluent API for setting values?

source/Halibut/Diagnostics/HalibutLimits.cs
```diff
@@ -107,4 +108,102 @@ public static TimeSpan SafeTcpClientPooledConnectionTimeout
             }
         }
     }
+#pragma warning disable CS0612
+    public class HalibutTimeoutsAndLimits {
```
nathanwoctopusdeploy, line 112: Worth adding a unit test for this class to make sure we are setting the properties from the correct statics?

source/Halibut/HalibutRuntime.cs
```diff
@@ -91,6 +94,7 @@ public HalibutRuntime(IServiceFactory serviceFactory, X509Certificate2 serverCer
             this.messageSerializer = messageSerializer;
             this.pollingReconnectRetryPolicy = pollingReconnectRetryPolicy;
             invoker = new ServiceInvoker(serviceFactory);
+            this.TimeoutsAndLimits = halibutTimeoutsAndLimits;
```
nathanwoctopusdeploy, line 97: If a sync Halibut runtime is created, would it make sense to enforce that `halibutTimeoutsAndLimits` should be null, so the caller knows these are not going to apply and we don't create confusion?

source/Halibut/Transport/Protocol/ControlMessageReader.cs
nathanwoctopusdeploy: WebSockets?

source/Halibut/Transport/Protocol/WebSocketStream.cs
```diff
@@ -50,7 +50,11 @@ public async Task<string> ReadTextMessage()
             var buffer = new ArraySegment<byte>(new byte[10000]);
             while (true)
             {
+                // TODO - ASYNC ME UP!
+                // What should the timeout be here? thew new code that allows passing in a timeout or the static value?
```
nathanwoctopusdeploy, line 54: https://github.com/OctopusDeploy/Halibut/pull/414 changes this stream to add async methods

source/Halibut/Transport/SecureConnection.cs
```diff
@@ -30,7 +30,11 @@ public void NotifyUsed()
 
         public bool HasExpired()
         {
+            // TODO - ASYNC ME UP!
+            // Use the HalibutRuntimeLimits
```
nathanwoctopusdeploy, line 34: Is there a reason some changes are being split out of this pr?

source/Halibut/Transport/TcpConnectionFactory.cs
```diff
@@ -47,12 +47,12 @@ public IConnection EstablishNewConnection(ExchangeProtocolBuilder exchangeProtoc
             return new SecureConnection(client, ssl, exchangeProtocolBuilder, log);
         }
         
-        public async Task<IConnection> EstablishNewConnectionAsync(ExchangeProtocolBuilder exchangeProtocolBuilder, ServiceEndPoint serviceEndpoint, ILog log, CancellationToken cancellationToken)
+        public async Task<IConnection> EstablishNewConnectionAsync(ExchangeProtocolBuilder exchangeProtocolBuilder, ServiceEndPoint serviceEndpoint, HalibutTimeoutsAndLimits halibutTimeoutsAndLimits, ILog log, CancellationToken cancellationToken)
```
nathanwoctopusdeploy, line 50: Given the limits are set on the runtime, should they be given to the constructor of the factory and then used in the methods so wee don't have to pass them around to all methods

nathanwoctopusdeploy: May mean some methods below would benefit from not being static

nathanwoctopusdeploy (approved):
Nice, a few suggestions. Nice approach marking the old limits as obsolete so you get compile time errors where they are being used 👍

source/Halibut.Tests/Timeouts/SendingAndReceivingRequestMessagesTimeoutsFixture.cs
LukeButters: Yep

source/Halibut/Diagnostics/HalibutLimits.cs
LukeButters: ran code clean up

source/Halibut.Tests/Transport/DiscoveryClientFixture.cs
LukeButters: `HalibutTimeoutsAndLimits` get its defaults from what is set statically which will be the reduced values as part of `LowerHalibutLimits`. In theory those should be "good" defaults.

source/Halibut/Diagnostics/HalibutLimits.cs
LukeButters: Yeah I guess it can be an extension method.

source/Halibut/HalibutRuntime.cs
LukeButters: Yes and done.

source/Halibut/Transport/Protocol/ControlMessageReader.cs
LukeButters: We were not thinking of wrapping that stream in a stream that fakes timing out?

PR #428, opened 2023-08-14
Make ServiceInvoker async

acodrington:
### Description
This PR adds a `InvokeAsync` method to `IServiceInvoker` and `ServiceInvoker`, as part of [sc-53211]. This allows the ServiceInvoker to invoke async implementations of services with sync interfaces.

Upon succesfully acquiring a lease from the `IServiceFactory`, `InvokeAsync` will search for an async method on the registered service which matches the name of requested method, but also follows these conventions:
* Must end with `Async`
* Must have a `CancellationToken` as its last argument
* Must return either `Task` or `Task<T>`, depending on whether the originally requested method returns `void` or a result.

This feature is dependant on the implementation of `IServiceFactory` being used (Octopus Server and Tentacle use their own implementations), but we can now register async implementations of sync services in the `DelegateServiceFactory`:

```
var serviceFactory = new DelegateServiceFactory();
serviceFactory.Register<IMySyncService, IMyAsyncService>(() => new MyAsyncServiceImpl());
```

In the above example, there is no formal relationship between `IMySyncService` and `IMyAsyncService`, but the async interface must replicate the sync interface whilst adhering to the conventions listed above.

The `DelegateServiceFactory` will throw an exception upon registration if the async interface does not follow the conventions, via the new `AsyncServiceVerifier` helper class.

### Main changes:
* Add async code path to `IServiceInvoker`
* Add `AsyncServiceVerifier` for verifying the mapping between sync and async service interfaces.
* Add tests for ServiceInvoker and AsyncServiceVerifier
* Fixed bug in sync code of ServiceInvoker, where method lookup could fail if params array was `null` rather than empty
* Added overload method `DelegateServiceFactory` to allow registering async implementations against a sync service interface, and verify that mapping conforms to conventions.
* Added additional internal method to `DelegateServiceFactory` for purposefully registering non-comforming sync/async service interfaces (used in testing)

### DIFF
--- a/source/Halibut.Tests/AsyncServicesFixture.cs
+++ b/source/Halibut.Tests/AsyncServicesFixture.cs
@@ -1,4 +1,6 @@
-﻿using System.Threading.Tasks;
+﻿using System;
+using System.IO;
+using System.Threading.Tasks;
 using FluentAssertions;
 using Halibut.Tests.Support;
 using Halibut.Tests.Support.TestAttributes;
@@ -14,8 +16,8 @@ namespace Halibut.Tests
     public class AsyncServicesFixture : BaseTest
     {
         [Test]
-        [LatestClientAndLatestServiceTestCases(testAsyncServicesAsWell: true)]
-        public async Task AsyncServicesCanBeRegisteredAndResolved(ClientAndServiceTestCase clientAndServiceTestCase)
+        [LatestClientAndLatestServiceTestCases(testAsyncServicesAsWell: true, testNetworkConditions: false)]
+        public async Task AsyncServiceWithParams_CanBeRegisteredAndResolved(ClientAndServiceTestCase clientAndServiceTestCase)
         {
             var clientAndServiceBuilder = clientAndServiceTestCase.CreateTestCaseBuilder()
                 .AsLatestClientAndLatestServiceBuilder();
@@ -34,18 +36,8 @@ namespace Halibut.Tests
             var clientAndService = await clientAndServiceBuilder.Build(CancellationToken);
 
             var value = Some.RandomAsciiStringOfLength(8);
-            string result = null;
-            await clientAndServiceTestCase.ForceClientProxyType.ToSyncOrAsync()
-                .WhenSync(() =>
-                {
-                    var echoServiceClient = clientAndService.Client.CreateClient<IEchoService>(clientAndService.ServiceEndPoint, CancellationToken);
-                    result = echoServiceClient.SayHello(value);
-                })
-                .WhenAsync(async () =>
-                {
-                    var echoServiceClient = clientAndService.CreateClient<IEchoService, IAsyncClientEchoService>();
-                    result = await echoServiceClient.SayHelloAsync(value);
-                });
+            var echoServiceClient = clientAndService.CreateClient<IEchoService, IAsyncClientEchoService>();
+            var result = await echoServiceClient.SayHelloAsync(value);
 
             if (clientAndServiceTestCase.ServiceAsyncHalibutFeature.IsEnabled())
             {
@@ -55,6 +47,37 @@ namespace Halibut.Tests
             {
                 result.Should().Be($"{value}...");
             }
+        }
+
+        [Test]
+        [LatestClientAndLatestServiceTestCases(testAsyncServicesAsWell: true, testNetworkConditions: false)]
+        public async Task AsyncServiceWithNoParams_CanBeRegisteredAndResolved(ClientAndServiceTestCase clientAndServiceTestCase)
+        {
+            var clientAndServiceBuilder = clientAndServiceTestCase.CreateTestCaseBuilder()
+                .AsLatestClientAndLatestServiceBuilder();
+
+            if (clientAndServiceTestCase.ServiceAsyncHalibutFeature.IsEnabled())
+            {
+                clientAndServiceBuilder = clientAndServiceBuilder
+                    .WithAsyncService<ILockService, IAsyncLockService>(() => new AsyncLockService());
+            }
+            else
+            {
+                clientAndServiceBuilder = clientAndServiceBuilder
+                    .WithService<ILockService>(() => new LockService());
+            }
+
+            var clientAndService = await clientAndServiceBuilder.Build(CancellationToken);
+
+            string fileToWaitFor = Path.GetTempFileName();
+            string fileWhenRequestStarted = Path.GetTempFileName();
+            
+            File.Delete(fileToWaitFor);
+
+            var lockServiceClient = clientAndService.CreateClient<ILockService, IAsyncClientLockService>();
+            await lockServiceClient.WaitForFileToBeDeletedAsync(fileToWaitFor, fileWhenRequestStarted);
+
+            File.Exists(fileWhenRequestStarted).Should().BeTrue();
         }
     }
 }
--- a/source/Halibut.Tests/AsyncServicesFixture.cs
+++ b/source/Halibut.Tests/AsyncServicesFixture.cs
@@ -19,21 +19,10 @@ namespace Halibut.Tests
         [LatestClientAndLatestServiceTestCases(testAsyncServicesAsWell: true, testSyncService:false, testNetworkConditions: false)]
         public async Task AsyncServiceWithReturnType_CanBeRegisteredAndResolved(ClientAndServiceTestCase clientAndServiceTestCase)
         {
-            var clientAndServiceBuilder = clientAndServiceTestCase.CreateTestCaseBuilder()
-                .AsLatestClientAndLatestServiceBuilder();
-
-            if (clientAndServiceTestCase.ServiceAsyncHalibutFeature.IsEnabled())
-            {
-                clientAndServiceBuilder = clientAndServiceBuilder
-                    .WithAsyncService<IEchoService, IAsyncEchoService>(() => new AsyncEchoService());
-            }
-            else
-            {
-                clientAndServiceBuilder = clientAndServiceBuilder
-                    .WithService<IEchoService>(() => new EchoService());
-            }
-
-            var clientAndService = await clientAndServiceBuilder.Build(CancellationToken);
+            var clientAndService = await clientAndServiceTestCase.CreateTestCaseBuilder()
+                .AsLatestClientAndLatestServiceBuilder()
+                .WithAsyncService<IEchoService, IAsyncEchoService>(() => new AsyncEchoService())
+                .Build(CancellationToken);
 
             var value = Some.RandomAsciiStringOfLength(8);
             var echoServiceClient = clientAndService.CreateClient<IEchoService, IAsyncClientEchoService>();
@@ -53,21 +42,10 @@ namespace Halibut.Tests
         [LatestClientAndLatestServiceTestCases(testAsyncServicesAsWell: true, testSyncService: false, testNetworkConditions: false)]
         public async Task AsyncServiceWithNoReturnType_CanBeRegisteredAndResolved(ClientAndServiceTestCase clientAndServiceTestCase)
         {
-            var clientAndServiceBuilder = clientAndServiceTestCase.CreateTestCaseBuilder()
-                .AsLatestClientAndLatestServiceBuilder();
-
-            if (clientAndServiceTestCase.ServiceAsyncHalibutFeature.IsEnabled())
-            {
-                clientAndServiceBuilder = clientAndServiceBuilder
-                    .WithAsyncService<ILockService, IAsyncLockService>(() => new AsyncLockService());
-            }
-            else
-            {
-                clientAndServiceBuilder = clientAndServiceBuilder
-                    .WithService<ILockService>(() => new LockService());
-            }
-
-            using var clientAndService = await clientAndServiceBuilder.Build(CancellationToken);
+            using var clientAndService = await clientAndServiceTestCase.CreateTestCaseBuilder()
+                .AsLatestClientAndLatestServiceBuilder()
+                .WithAsyncService<ILockService, IAsyncLockService>(() => new AsyncLockService())
+                .Build(CancellationToken);
 
             string fileToWaitFor = Path.GetTempFileName();
             string fileWhenRequestStarted = Path.GetTempFileName();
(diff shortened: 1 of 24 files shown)


source/Halibut/ServiceModel/DelegateServiceFactory.cs
```diff
@@ -22,6 +23,33 @@ public DelegateServiceFactory Register<TContract>(Func<TContract> implementation
...
+            lock (serviceTypes)
+            {
+                serviceTypes.Add(serviceType);
+            }
+
+            return this;
+        }
+        
+        // Used for testing badly-registered sync/async contracts
+        internal DelegateServiceFactory RegisterWithNoVerification<TContract, TAsyncContract>(Func<TAsyncContract> implementation)
```
LukeButters, line 41: Would it make sense to make a special _test_ service factory which allows this?

source/Halibut/Util/AsyncServiceVerifier.cs
```diff
@@ -0,0 +1,57 @@
...
+        public static void VerifyAsyncSurfaceAreaFollowsConventions<TSyncService, TAsyncService>()
+        {
+            const string Async = "Async";
+            
+            Type syncService = typeof(TSyncService);
+            Type asyncService = typeof(TAsyncService);
+            
+            var syncServiceMethods = syncService.GetMethods();
+
+            bool ReturnTypesMatch(MethodInfo syncMethod, MethodInfo asyncMethod)
```
LukeButters, line 30: nit: this should probably be a private method outside of this method.

source/Halibut/ServiceModel/DelegateServiceFactory.cs
acodrington: That's a good idea 👍🏻 I'll do that

source/Halibut/Util/AsyncServiceVerifier.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Halibut.Exceptions;
+
+namespace Halibut.Util
+{
+    public class AsyncServiceVerifier
+    {
+        /// <summary>
+        /// This method enforces that all methods on the sync service have a corresponding method on the async service which adheres to the following conventions:
```
LukeButters, line 13: Is it intentional that the async interface can have methods not on the TSyncService?

acodrington: It's intentional that we don't care if there are, yes. I figured that you're always looking for an async version of a requested sync service, and the important part is that the async interface should cover the whole sync interface. If the async interface has some extra methods, then so what?

We could be stricter here, but I'm not sure it gives much benefit.

source/Halibut/ServiceModel/ServiceInvoker.cs
```diff
@@ -36,9 +38,111 @@ public ResponseMessage Invoke(RequestMessage requestMessage)
             }
         }
 
+        public async Task<ResponseMessage> InvokeAsync(RequestMessage requestMessage)
+        {
+            using var lease = factory.CreateService(requestMessage.ServiceName);
+            var methods = lease.Service.GetType().GetHalibutServiceMethods().Where(m => string.Equals(m.Name, requestMessage.MethodName, StringComparison.OrdinalIgnoreCase)).ToList();
```
LukeButters, line 44: Should the lookup be the reverse? specifically prefer an async method over a sync method?

source/Halibut.Tests/Builders/ServiceFactoryBuilder.cs
```diff
@@ -0,0 +1,56 @@
...
+using Halibut.Tests.Util;
+
+namespace Halibut.Tests.Builders
+{
+    public class ServiceFactoryBuilder
+    {
+        bool _conventionVerificationDisabled;
+        DelegateServiceFactory factoryWithConventionVerification = new();
+        NaiveDelegateServiceFactory factoryWithNoConventionVerification = new();
+
```
acodrington, line 12: Why register everything in one factory, when you can register in two!

I started writing something to store the factory Func<T> and the Contract and ClientContract types, then register them in the `Build` method, but it was way easier to just keep a copy of each, and try to register in all of them - `Build` just returns the one you want!

So a pretty dumb implementation, but it's simple and we can improve it later.

source/Halibut/ServiceModel/DelegateServiceFactory.cs
acodrington: Done. Clunky, but done!

source/Halibut/ServiceModel/ServiceInvoker.cs
```diff
@@ -36,9 +38,111 @@ public ResponseMessage Invoke(RequestMessage requestMessage)
...
+            return ResponseMessage.FromResult(requestMessage, asyncResult);
+        }
+
+        static MethodInfo SelectAsyncMethod(IList<MethodInfo> methods, RequestMessage requestMessage)
+        {
+            var argumentTypes = requestMessage.Params?.Select(s => s == null ? null : s.GetType()).ToList() ?? new List<Type>();
+
+            var matches = new List<MethodInfo>();
+
+            foreach (var method in methods)
```
LukeButters, line 73: Could this foreach loop be pulled out into its own function?

source/Halibut/ServiceModel/ServiceInvoker.cs
```diff
@@ -36,9 +38,111 @@ public ResponseMessage Invoke(RequestMessage requestMessage)
...
+                    continue;
+                }
+
+                if (parameters.Length != argumentTypes.Count)
+                {
+                    continue;
+                }
+
+                var isMatch = true;
+                for (var i = 0; i < parameters.Length; i++)
```
LukeButters, line 91: nit: could this for loop be pulled out into its own function?

source/Halibut/ServiceModel/ServiceInvoker.cs
```diff
@@ -103,19 +207,28 @@ static MethodInfo SelectMethod(IList<MethodInfo> methods, RequestMessage request
             throw new AmbiguousMethodMatchHalibutClientException(message.ToString(), new AmbiguousMatchException(message.ToString()));
         }
 
+        async Task<object> InvokeAsyncMethod(MethodInfo method, object obj, params object[] parameters)
+        {
+            var task = (Task)method.Invoke(obj, parameters);
+            await task.ConfigureAwait(false);
+            var resultProperty = task.GetType().GetProperty("Result");
```
LukeButters, line 214: Is it possible to avoid the reflection?

source/Halibut/ServiceModel/ServiceInvoker.cs
```diff
@@ -36,9 +38,111 @@ public ResponseMessage Invoke(RequestMessage requestMessage)
...
+        public async Task<ResponseMessage> InvokeAsync(RequestMessage requestMessage)
+        {
+            using var lease = factory.CreateService(requestMessage.ServiceName);
+            var methods = lease.Service.GetType().GetHalibutServiceMethods().Where(m => string.Equals(m.Name, requestMessage.MethodName, StringComparison.OrdinalIgnoreCase)).ToList();
+            if (methods.Count != 0)
+            {
+                var method = SelectMethod(methods, requestMessage);
+                var args = GetArguments(requestMessage, method);
+                var result = method.Invoke(lease.Service, args);
+                return ResponseMessage.FromResult(requestMessage, result);
```
LukeButters, line 50: nit: Could this be extracted into its own method?

source/Halibut/ServiceModel/ServiceInvoker.cs
```diff
@@ -36,9 +38,111 @@ public ResponseMessage Invoke(RequestMessage requestMessage)
...
+            }
+
+            var asyncMethodName = requestMessage.MethodName + "Async";
```
LukeButters, line 53: Maybe adding "Async" can be pushed into the `AsyncCompatibilityHelper`

LukeButters (approved):

LukeButters (changes requested):
Ah I think this is missing an integration test.

LukeButters:
The description I think should give an example of the behaviour that this PR adds e.g. an example around how an async service can be supplied and called as well as some details around how it has to be done/what the methods need to look like etc.

source/Halibut/ServiceModel/ServiceInvoker.cs
LukeButters: Lets switch the order since its probably more likely that async code calling async code is a better default over async code calling sync code.

As a library maintainer, changing the order later could be hard since we don't know if someone has started to depend on the sync being called over async. Which is possible since we don't enforce either sync or async services.

source/Halibut/ServiceModel/ServiceInvoker.cs
acodrington: No, I don't think so. ServiceInvoker is already primarly based on reflection, I'm not sure how to do it another way.

source/Halibut.Tests/Util/NaiveDelegateServiceFactory.cs
```diff
@@ -0,0 +1,93 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Halibut.Exceptions;
+using Halibut.ServiceModel;
+
+namespace Halibut.Tests.Util
+{
+    // Just like the DelegateServiceFactory, but without any sanity checking
+    // when registering async implementations of sync services.
+    public class NaiveDelegateServiceFactory : IServiceFactory
```
LukeButters, line 11: nit: Could call it `NoSanityCheckingDelegateServiceFactory`

source/Halibut.Tests/AsyncServicesFixture.cs
```diff
@@ -0,0 +1,83 @@
...
+                clientAndServiceBuilder = clientAndServiceBuilder
+                    .WithAsyncService<IEchoService, IAsyncEchoService>(() => new AsyncEchoService());
+            }
+            else
+            {
+                clientAndServiceBuilder = clientAndServiceBuilder
+                    .WithService<IEchoService>(() => new EchoService());
+            }
+
+            var clientAndService = await clientAndServiceBuilder.Build(CancellationToken);
```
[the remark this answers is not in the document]
acodrington: Yes! Good catch, thanks 👍🏻

source/Halibut.Tests/AsyncServicesFixture.cs
```diff
@@ -0,0 +1,83 @@
...
+                .AsLatestClientAndLatestServiceBuilder();
+
+            if (clientAndServiceTestCase.ServiceAsyncHalibutFeature.IsEnabled())
+            {
+                clientAndServiceBuilder = clientAndServiceBuilder
+                    .WithAsyncService<IEchoService, IAsyncEchoService>(() => new AsyncEchoService());
+            }
+            else
+            {
+                clientAndServiceBuilder = clientAndServiceBuilder
```
LukeButters, line 32: Is it valuable to have this if? would it make sense to only test async and so have only the `.WithAsyncService<IEchoService, IAsyncEchoService>` line?

acodrington: The main value in having the sync version is that it verifies that the behaviour still works like it did before in that scenario, which I think is worthwhile here as the feature is still behind a feature toggle.

LukeButters: is the existing behaviour tested over in the UsageFixture?

acodrington: Yeah that's true.

Is there a way we can test _only_ the async service code path, or should I update the test attribute i.e. `testAsyncServicesAsWell`

LukeButters: yes, `, testSyncService: false`

source/Halibut/ServiceModel/ServiceInvoker.cs
```diff
@@ -103,19 +209,46 @@ static MethodInfo SelectMethod(IList<MethodInfo> methods, RequestMessage request
...
         }
 
+        async Task<object> InvokeAsyncMethod(MethodInfo method, object obj, params object[] parameters)
+        {
+            var returnType = method.ReturnType;
+
+            // If the return type is not Task or Task<T>, then shrug and explode!
+            if (!typeof(Task).IsAssignableFrom(returnType))
+            {
+                throw new Exception();
```
LukeButters, line 219: Throw a more meaningful message here.

acodrington: Ah damn, I was going to change that, but I got distracted by someone asking me if the PR was ready ;)

Will do it now 👍🏻

source/Halibut.Tests/AsyncServicesFixture.cs
```diff
@@ -0,0 +1,83 @@
...
+            }
+            else
+            {
+                result.Should().Be($"{value}...");
+            }
+        }
+
+        [Test]
+        [LatestClientAndLatestServiceTestCases(testAsyncServicesAsWell: true, testNetworkConditions: false)]
+        public async Task AsyncServiceWithNoParams_CanBeRegisteredAndResolved(ClientAndServiceTestCase clientAndServiceTestCase)
```
LukeButters, line 54: Should be this be calling a service that has a method that takes no arguments?

acodrington: Ah it's poorly named - it's for testing that we can invoke an async method with `Task` for return type (rather than `Task<T>`)

I can rename these tests, but then that makes me realise I should probably have tests for the _actual_ params/no params cases. Should I add those tests in a separate PR after this one, to prevent this one dragging on longer?

LukeButters: Best do it now in case it digs out some bugs

source/Halibut/ServiceModel/ServiceInvoker.cs
acodrington: Makes sense, done.

LukeButters (approved):

PR #429, opened 2023-08-15
Fix Pending Request Queue Async Memory Usage Issues

nathanwoctopusdeploy:
# Background

The `PendingRequestQueueAsync` had to significantly change the way locking worked from the `PendingRequestQueue` when converted to be async.

This PR fixes some memory leaks identified when testing async Halibut in server.

- AsyncManualResetEvent.WaitAsync should be cancelled if used as part of a Task.WhenAny if passed a cancellation token or no cancellation token should be passed to avoid a memory leak. Not passing a cancellation token if not required is the best scenario as it avoids the creation of a cancellation token source and a Task.WhenAny
- Task.Delay should be cancelled when used in a Task.WhenAny to avoid leaking timers
- Ensure CancellationTokenSource.CreateLinkedCancellationToken is disposed
- PendingRequest used a SemaphoreSlim which was not disposed

This PR also adds a change to delete a PendingRequest from the `queue` when the code that queued the pending request has stopped waiting (cancelled or timed out). This allows the request to be disposed and also removes a known issues where null may be returned for DequeuNextAsync while there are items waiting to be dequeued and removes an un-necessary poll loop from the Polling Service to get the next request.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/ServiceModel/PendingRequestQueueFixture.cs
+++ b/source/Halibut.Tests/ServiceModel/PendingRequestQueueFixture.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
@@ -233,7 +234,7 @@ namespace Halibut.Tests.ServiceModel
         }
 
         [Test]
-        [Ignore("This test takes way too long on Team City")]
+        [IgnoreOnTeamCity("This test takes way too long on Team City")]
         [NonParallelizable]
         public async Task QueueAndWait_Can_Queue_Dequeue_Apply_VeryLargeNumberOfRequests()
         {
@@ -262,6 +263,83 @@ namespace Halibut.Tests.ServiceModel
             
             // Assert
             await ApplyResponsesConcurrentlyAndEnsureAllQueueResponsesMatch(sut, requestsInOrder, queueAndWaitTasksInOrder);
+        }
+
+        [Test]
+        [NonParallelizable]
+        public async Task QueueAndWait_Can_Queue_Dequeue_WhenRequestsAreBeingCancelled()
+        {
+            // Arrange
+            const string endpoint = "poll://endpoint001";
+            const int totalRequest = 500;
+            const int minimumCancelledRequest = 100;
+
+            var sut = new PendingRequestQueueBuilder().WithSyncOrAsync(SyncOrAsync.Async).WithEndpoint(endpoint).Build();
+            
+            var requestsInOrder = Enumerable.Range(0, totalRequest)
+                .Select(_ => new RequestMessageBuilder(endpoint).Build())
+                .ToList();
+
+            // Act
+            var queueAndWaitTasksInOrder = requestsInOrder
+                .Select(request =>
+                {
+                    var requestCancellationTokenSource = new CancellationTokenSource();
+                    return new Tuple<Task<ResponseMessage>, CancellationTokenSource>(
+                        StartQueueAndWait(sut, request, SyncOrAsync.Async, requestCancellationTokenSource.Token), 
+                        requestCancellationTokenSource);
+                })
+                .ToList();
+
+            await WaitForQueueCountToBecome(sut, requestsInOrder.Count);
+            
+            var index = 0;
+            var cancelled = 0;
+            var dequeueTasks = new ConcurrentBag<Task<RequestMessage>>();
+
+            var cancelSomeTask = Task.Run(() =>
+            {
+                foreach (var _ in requestsInOrder)
+                {
+                    var currentIndex = Interlocked.Increment(ref index);
+
+                    if(currentIndex % 2 == 0)
+                    {
+                        Interlocked.Increment(ref cancelled);
+                        queueAndWaitTasksInOrder.ElementAt(index-1).Item2.Cancel();
+                    }
+                }
+            });
+
+            // Cancellation is slow so give it a head start
+            while (cancelled < minimumCancelledRequest)
+            {
+                await Task.Delay(10);
+            }
+
+            var dequeueAllTask = Task.Run(() =>
+            {
+                foreach (var _ in requestsInOrder)
+                {
+                    var task = sut.DequeueAsync(CancellationToken);
+                    dequeueTasks.Add(task);
+                }
+            });
+
+            await Task.WhenAll(cancelSomeTask, dequeueAllTask);
+
+            var completedTask = await Task.WhenAll(dequeueTasks);
+
+            foreach (var queueAndWait in queueAndWaitTasksInOrder)
+            {
+                queueAndWait.Item2.Dispose();
+            }
+
+            // Assert
+            // This test is pretty non-deterministic. It attempts to create some cancellation chaos
+            // and we just want to ensure some things did cancel and some things did complete and nothing
+            // threw an exception
+            completedTask.Count(x => x != null).Should().BeLessThan(totalRequest - minimumCancelledRequest).And.BeGreaterThanOrEqualTo(totalRequest / 2);
         }
 
         [Test]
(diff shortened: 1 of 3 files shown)


source/Halibut/ServiceModel/PendingRequestQueueAsync.cs
```diff
@@ -83,27 +83,83 @@ async Task<PendingRequest> DequeueNextAsync(CancellationToken cancellationToken)
...
+
+                    if (queue.IsEmpty)
+                    {
+                        itemAddedToQueue.Reset();
+                    }
                 }
-                
-                if (queue.IsEmpty)
+
+                if (first.IsDisposed)
```
nathanwoctopusdeploy, line 156: Disposal of the PendingRequest is probably the biggest potential for bugs here and may make some of our tests that depend on a "dead" PendingRequest resulting in an instant null being returned to start failing as they will now wait the full `HalibutLimits.PollingQueueWaitTimeout` before null is returned

nathanwoctopusdeploy:
[sc-56801]

source/Halibut/ServiceModel/PendingRequestQueueAsync.cs
```diff
@@ -54,50 +54,97 @@ public async Task<ResponseMessage> QueueAndWaitAsync(RequestMessage request, Req
...
+                }
+
+            }
+        }
 
         public async Task<RequestMessage> DequeueAsync(CancellationToken cancellationToken)
         {
-            var pending = await DequeueNextAsync(cancellationToken);
-            if (pending == null) return null;
+            var timer = Stopwatch.StartNew();
```
sburmanoctopus, line 93: I wonder if this loop is having the desired outcome.

The aim is to get rid of queue items that are invalid.

But here, we start `DequeueNextAsync`, which can delay for the rest of the timeout. If that happens, and we find `BeginTransfer` returns a dud request, when we go around again, `DequeueNextAsync` will return null (as we reached the end of our time).

I wonder if instead of having a loop around the whole thing, we instead put the loop around `BeginTransfer`.

I.e., if we get something to transfer, and it becomes a dud, then keep pulling from the queue until it is either null or not a dud.

We don't really need to worry about the delay then, because the delay is really just to hang around until something appears on the list. This just becomes a tight "cleanup" loop. Plus it will cope when there are multiple duds on the queue.

source/Halibut/ServiceModel/PendingRequestQueueAsync.cs
```diff
@@ -54,50 +54,97 @@ public async Task<ResponseMessage> QueueAndWaitAsync(RequestMessage request, Req
...
             {
                 return first;
             }
+            
+            using var cleanupCancellationTokenSource = new CancellationTokenSource();
+            using var linkedCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, cleanupCancellationTokenSource.Token);
+
+            await Task.WhenAny(
+                // ReSharper disable once MethodSupportsCancellation
+                itemAddedToQueue.WaitAsync(/*Do not pass a cancellation token as it will increase memory usage and may cause memory leaks*/),
```
sburmanoctopus, line 125: It might be worth elaborating on how this happens?

Not in the code, no one will read that. But if we elaborate against this PR (this comment even), then it is somewhat discoverable?

source/Halibut/ServiceModel/PendingRequestQueueAsync.cs
```diff
@@ -248,15 +295,22 @@ async Task<bool> WaitForResponseToBeSet(TimeSpan timeout, CancellationToken canc
 
             public async Task<bool> BeginTransfer()
             {
-                using (await transferLock.LockAsync())
+                try
```
sburmanoctopus, line 298: The reason we do this is a little obscure. Maybe comment why this is here?

sburmanoctopus (approved):
LGTM. Just some nits, but nothing that would stop this merging.

source/Halibut/ServiceModel/PendingRequestQueueAsync.cs
LukeButters: Example of a leak caused by `WaitAsync` on the `AsyncManualResetEvent` where the given token is not cancelled.
```
[Test]
        public async Task leak()
        {
            var cts = new CancellationTokenSource();
            var foo = new AsyncManualResetEvent();
            foo.Reset();
            for (int i = 0;i < 6520000;i++)
            {
                if(i % 10000 == 0) TestContext.Progress.WriteLine($"Up to: {i}");
                await Task.WhenAny(foo.WaitAsync(cts.Token), NoOp());
            }
        }
```

LukeButters: When `WaitAsync` with a `CancelletationToken` is called it ends up taking the task returned by `AsyncManualResetEvent.WaitAsync()` and passes it into:

```
        /// <summary>
        /// Asynchronously waits for the task to complete, or for the cancellation token to be canceled.
        /// </summary>
        /// <param name="this">The task to wait for. May not be <c>null</c>.</param>
        /// <param name="cancellationToken">The cancellation token that cancels the wait.</param>
        public static Task WaitAsync(this Task @this, CancellationToken cancellationToken)
        {
            if (@this == null)
                throw new ArgumentNullException(nameof(@this));

            if (!cancellationToken.CanBeCanceled)
                return @this;
            if (cancellationToken.IsCancellationRequested)
                return Task.FromCanceled(cancellationToken);
            return DoWaitAsync(@this, cancellationToken);
        }

        private static async Task DoWaitAsync(Task task, CancellationToken cancellationToken)
        {
            using (var cancelTaskSource = new CancellationTokenTaskSource<object>(cancellationToken))
                await (await Task.WhenAny(task, cancelTaskSource.Task).ConfigureAwait(false)).ConfigureAwait(false);
        }
```

Which is just extra wrapping, which consumes more memory.

LukeButters (approved):

LukeButters:
The bug in the queue can be reproduced with:
```
[Test]
        public async Task leakTheQueue()
        {
            var queue = new PendingRequestQueueBuilder().WithSyncOrAsync(SyncOrAsync.Async).WithEndpoint("poll://foo")
                .WithPollingQueueWaitTimeout(TimeSpan.FromMilliseconds(1))
                .Build();

            var cts = new CancellationTokenSource();
            var tasks = new List<Task>();
            for (int i = 0; i < 2000; i++)
            {
                tasks.Add(TryDequeue(queue, cts.Token));
            }

            await Task.WhenAll(tasks.ToArray());
            
        }
        
        static async Task TryDequeue(IPendingRequestQueue queue, CancellationToken cancellationToken)
        {
            for (int i = 0; i < 6520000; i++)
            {
                await queue.DequeueAsync(cancellationToken);
            }
        }
```

source/Halibut/ServiceModel/PendingRequestQueueAsync.cs
nathanwoctopusdeploy: I've revisited this. Can you 👀 @sburmanoctopus

source/Halibut.Tests/ServiceModel/PendingRequestQueueFixture.cs
```diff
@@ -264,6 +265,81 @@ public async Task QueueAndWait_Can_Queue_Dequeue_Apply_VeryLargeNumberOfRequests
...
+                    if(currentIndex % 2 == 0)
+                    {
+                        Interlocked.Increment(ref cancelled);
+                        queueAndWaitTasksInOrder.ElementAt(index-1).Item2.Cancel();
+                    }
+                }
+            });
+
+            // Cancellation is slow so give it a head start
+            if (cancelled < 100)
```
sburmanoctopus, line 313: Should this be a loop?

nathanwoctopusdeploy: Haha, it should!! i wondered why it was a bit crap at cancelling 😆

sburmanoctopus (approved):
LGTM

PR #430, opened 2023-08-15
Support HalibutTimeoutsAndLimits in SecureConnection

LukeButters:
# Background

Fixes one of the TODOs left in https://github.com/OctopusDeploy/Halibut/pull/427

by making SecureConnection support the per runtime timeouts 

[SC-45207]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Transport/ConnectionManagerFixture.cs
+++ b/source/Halibut.Tests/Transport/ConnectionManagerFixture.cs
@@ -10,6 +10,7 @@ using Halibut.Tests.Support;
 using Halibut.Tests.Support.TestAttributes;
 using Halibut.Transport;
 using Halibut.Transport.Protocol;
+using Halibut.Util;
 using NSubstitute;
 using NUnit.Framework;
 
--- a/source/Halibut.Tests/Transport/SecureClientFixture.cs
+++ b/source/Halibut.Tests/Transport/SecureClientFixture.cs
@@ -12,6 +12,7 @@ using Halibut.Tests.Support.TestAttributes;
 using Halibut.TestUtils.Contracts;
 using Halibut.Transport;
 using Halibut.Transport.Protocol;
+using Halibut.Util;
 using NSubstitute;
 using NUnit.Framework;
 using ILog = Halibut.Diagnostics.ILog;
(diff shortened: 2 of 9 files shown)


source/Halibut.Tests/Transport/ConnectionManagerFixture.cs
```diff
@@ -20,7 +21,11 @@ public class ConnectionManagerFixture
         {
             MessageExchangeProtocol ExchangeProtocolBuilder(Stream s, ILog l) => GetProtocol(s, l, syncOrAsync);
 
-            var connection = new SecureConnection(Substitute.For<IDisposable>(), Stream.Null, ExchangeProtocolBuilder, Substitute.For<ILog>());
+            AsyncHalibutFeature asyncHalibutFeature = syncOrAsync == SyncOrAsync.Sync ? AsyncHalibutFeature.Disabled : AsyncHalibutFeature.Enabled;
```
sburmanoctopus, line 24: This will conflict with my PR. I've already dealt with one conflict. Can we please merge this after my PR goes in? 🙏

source/Halibut.Tests/Transport/SecureClientFixture.cs
```diff
@@ -48,6 +49,10 @@ public void Dispose()
         [SyncAndAsync]
         public async Task SecureClientClearsPoolWhenAllConnectionsCorrupt(SyncOrAsync syncOrAsync)
         {
+            AsyncHalibutFeature asyncHalibutFeature = syncOrAsync == SyncOrAsync.Sync ? AsyncHalibutFeature.Disabled : AsyncHalibutFeature.Enabled;
```
sburmanoctopus, line 52: Should we put this in a helper?

source/Halibut/Transport/SecureConnection.cs
```diff
@@ -30,8 +40,10 @@ public void NotifyUsed()
 
         public bool HasExpired()
         {
-            // TODO - ASYNC ME UP!
-            // Use the HalibutRuntimeLimits
+            if (asyncHalibutFeature.IsEnabled())
+            {
+                return lastUsed < DateTimeOffset.UtcNow.Subtract(halibutTimeoutsAndLimits.SafeTcpClientPooledConnectionTimeout);
```
sburmanoctopus, line 45: We can keep them separate, but if we made this method simply be `return lastUsed < DateTimeOffset.UtcNow.Subtract(halibutTimeoutsAndLimits.SafeTcpClientPooledConnectionTimeout);` (without the `IsEnabled` check), would that not be the same?

Mainly because the default limits (which is what we use when we don't give any) use all the values from `HalibutLimits`, right?

sburmanoctopus (approved):
LGTM. Just some nits

source/Halibut/Transport/SecureConnection.cs
LukeButters: amazingly it will be null in that case.

PR #434, opened 2023-08-16
Disposing Asynchronously

sburmanoctopus:
[sc-53211]

# Background
As part of making Halibut async, we need to ensure we dispose of things asynchronously as well.

# Results

Related to https://github.com/OctopusDeploy/Issues/issues/8266

## Before

The [following places](https://docs.google.com/spreadsheets/d/1sxtq-K2c8h2Z0ZUFvlouyNTgfPJT0iQzasOJlcsbJEI/edit#gid=0) were found to have paths that would require async disposal. These were being disposed synchronously.

## After
The green classes in the spreadsheet are not disposed asynchronously.

We did not change `using` statements to `await using' statements in this PR. This touched over 20 files, so will be assessed and implemented in a different PR.

To ensure .NET 4.8 code could compile and use `IAsyncDisposable`, we had to import the `Microsoft.Bcl.AsyncInterfaces` package.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

- [ ] I have read [How we use GitHub Issues](https://github.com/OctopusDeploy/Issues/blob/master/docs/CONTRIBUTING.internal.md) for help deciding when and where it's appropriate to make an issue.
- [ ] I have considered informing or consulting the right people, according to the [ownership map](https://whimsical.com/ownership-map-NzbiD4HJyvhC9jNJNfS6TG).
- [ ] I have considered appropriate testing for my change.

PR #436, opened 2023-08-17
Await using

sburmanoctopus:
[sc-53211]
# Background

Continuing on from [making disposing async](https://github.com/OctopusDeploy/Halibut/pull/434), we need to make disposal with `using` statements become async.

# Results

## Before

Whenever something that supports async operations is disposed, it does the disposal synchronously on the thread. This blocks the thread if disposal involves things like IO.

## After

We now try to use `await using` instead of `using` do that any asynchronous operations done during disposal will no longer block the calling thread.

### .NET 4.8 Complications
Streams in .NET 4.8 do not support `IAsyncDisposable`. This means we cannot add `await` to `using` statements.

This was fixed by wrapping the `await` in a `#if`

### RewindableBufferStream
We also made `RewindableBufferStream` implement `IAsyncDisposable`. This allowed us to using `await using` whenever we interacted with this type of stream.

### Using Our Streams
We made all our streams extend `IAsyncDisposable` so that they would work in .NET 4.8 with `await using`

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


nathanwoctopusdeploy:
[sc-57124]

source/Halibut.Tests/Timeouts/SendingAndReceivingRequestMessagesTimeoutsFixture.cs
```diff
@@ -141,16 +141,25 @@ public async Task WhenThenNetworkIsPaused_WhileSendingARequestMessage_ATcpWriteT
...
 
                 var echoServiceTheErrorWillHappenOn = clientAndService.CreateClient<IEchoService, IAsyncClientEchoService>(IncreasePollingQueueTimeout());
-
-                var stringToSend = Some.RandomAsciiStringOfLength(numberOfBytesBeforePausingAStream * 20);
+                
                 var sw = Stopwatch.StartNew();
-                var e = (await AssertAsync.Throws<HalibutClientException>(() => echoServiceTheErrorWillHappenOn.SayHelloAsync(stringToSend))).And;
+                var e = (await AssertAsync.Throws<HalibutClientException>(() =>
+                {
+                    var stringToSend = Some.RandomAsciiStringOfLength(numberOfBytesBeforePausingAStream * 20);
```
sburmanoctopus, line 148: It was driving me nuts debugging in this area, and having VS take ages to "evaluate local variables" because of this large string. 

So I made sure it was no longer in scope. Solved!

source/Halibut.TestProxy/TcpTunnel.cs
```diff
@@ -23,7 +23,13 @@ public TcpTunnel(TcpClient fromClient, TcpClient toClient, ILogger logger)
 
         public async Task Tunnel(CancellationToken cancellationToken)
         {
+#if !NETFRAMEWORK
```
LukeButters, line 26: If only dotnet had `#define` or better yet virtual threads.

LukeButters (approved):

sburmanoctopus:
[sc-57123]

PR #437, opened 2023-08-17
Create ApmToTapStream

acodrington:
# Background

This PR introduces a new stream `ApmToTapStream`, which can ensure that execution doesn’t unintentionally go down sync code paths when using some streams in .NET Framework 4.8.

Some streams in net48 do not override the modern TAP async methods (e.g. `ReadAsync`, `WriteAsync`) and thus fallback to the default implementation in the `Stream` base class when they are called. This default implementation uses the APM async methods (e.g. `BeginRead`, `EndRead`), which in turn calls the sync methods like `Read` and `Write`. This has the unintended consequence of causing a call to `ReadAsync` on an outer stream eventually calling `Read` on an inner stream, therefore making the execution path go from async to sync.

`DeflateStream` is one of these streams, and the usage of this type in Halibut’s `MessageSerializer` class means that many of our async calls in net48 were not truly async. By wrapping `DeflateStream`'s inner stream in a `ApmToTapStream`, this problem can be mitigated.

# Results

Fixes [sc-57049]

## Before

```
async Task<T> ReadCompressedMessageAsync<T>(Stream stream, IRewindableBuffer rewindableBuffer, CancellationToken cancellationToken)
{
    // Do some stuff

    try
    {
        using var compressedByteCountingStream = new ByteCountingStream(stream, OnDispose.LeaveInputStreamOpen);
        using var zip = new DeflateStream(compressedByteCountingStream, CompressionMode.Decompress, true);
        using var decompressedByteCountingStream = new ByteCountingStream(zip, OnDispose.LeaveInputStreamOpen);
        using var deflatedInMemoryStream = new ReadIntoMemoryBufferStream(decompressedByteCountingStream, readIntoMemoryLimitBytes, OnDispose.LeaveInputStreamOpen);
        
        // Read the stream(s) here
        await deflatedInMemoryStream.BufferIntoMemoryFromSourceStreamUntilLimitReached(cancellationToken);

        // Do some more stuff
    }
}

```
ApmToTapStream Before

## After

```
async Task<T> ReadCompressedMessageAsync<T>(Stream stream, IRewindableBuffer rewindableBuffer, CancellationToken cancellationToken)
{
    // Do some stuff

    try
    {
        using var compressedByteCountingStream = new ByteCountingStream(stream, OnDispose.LeaveInputStreamOpen);
        using var apmToTapStream = new ApmToTapStream(compressedByteCountingStream);
        using var zip = new DeflateStream(apmToTapStream, CompressionMode.Decompress, true);
        using var decompressedByteCountingStream = new ByteCountingStream(zip, OnDispose.LeaveInputStreamOpen);
        using var deflatedInMemoryStream = new ReadIntoMemoryBufferStream(decompressedByteCountingStream, readIntoMemoryLimitBytes, OnDispose.LeaveInputStreamOpen);

        // Read the stream(s) here
        await deflatedInMemoryStream.BufferIntoMemoryFromSourceStreamUntilLimitReached(cancellationToken);

        // Do some more stuff
    }
}
```
ApmToTapStream After

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Transport/Protocol/MessageSerializer.cs
+++ b/source/Halibut/Transport/Protocol/MessageSerializer.cs
@@ -75,8 +75,9 @@ namespace Halibut.Transport.Protocol
         {
             using var compressedByteCountingStream = new ByteCountingStream(stream, OnDispose.LeaveInputStreamOpen);
             using var compressedInMemoryBuffer = new WriteIntoMemoryBufferStream(compressedByteCountingStream, writeIntoMemoryLimitBytes, OnDispose.LeaveInputStreamOpen);
+            using var apmToTapStream = new ApmToTapStream(compressedInMemoryBuffer);
 
-            using (var zip = new DeflateStream(compressedInMemoryBuffer, CompressionMode.Compress, true))
+            using (var zip = new DeflateStream(apmToTapStream, CompressionMode.Compress, true))
             using (var bson = new BsonDataWriter(zip) { CloseOutput = false })
             {
                 // for the moment this MUST be object so that the $type property is included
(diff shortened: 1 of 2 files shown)


source/Halibut/Transport/Protocol/MessageSerializer.cs
```diff
@@ -75,8 +75,9 @@ public async Task WriteMessageAsync<T>(Stream stream, T message, CancellationTok
         {
             using var compressedByteCountingStream = new ByteCountingStream(stream, OnDispose.LeaveInputStreamOpen);
             using var compressedInMemoryBuffer = new WriteIntoMemoryBufferStream(compressedByteCountingStream, writeIntoMemoryLimitBytes, OnDispose.LeaveInputStreamOpen);
+            using var apmToTapStream = new ApmToTapStream(compressedInMemoryBuffer);
```
LukeButters, line 78: This stream will have dispose call dispose on the inner stream, which is not inline with what is happening on the other streams but will be fine since the other streams I think will prevent that dispose call from making it all the way down.

LukeButters (approved):

PR #439, opened 2023-08-17
Write a test to find sync IO

LukeButters:
# Background

[SC-55504]

Adds a test which finds sync IO.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/AsyncClientAndServiceMustNotUseSyncNetworkIO.cs
+++ b/source/Halibut.Tests/AsyncClientAndServiceMustNotUseSyncNetworkIO.cs
@@ -1,0 +1,83 @@
+﻿using System;
+using System.Linq;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Halibut.Tests.Support;
+using Halibut.Tests.Support.Streams;
+using Halibut.Tests.Support.TestAttributes;
+using Halibut.Tests.Support.TestCases;
+using Halibut.Tests.TestServices.Async;
+using Halibut.TestUtils.Contracts;
+using Halibut.Util;
+using NUnit.Framework;
+
+namespace Halibut.Tests
+{
+    public class AsyncClientAndServiceMustNotUseSyncNetworkIO : BaseTest
+    {
+        [Test]
+        // We are testing we only have async IO so only test async client with an async service.
+        // We already know a sync service/client does sync IO.
+        [LatestClientAndLatestServiceTestCases(testSyncService: false, testAsyncServicesAsWell: true,
+            testSyncClients: false, testAsyncClients: true,
+            // TODO: ASYNC ME UP!
+            // WebSockets are not yet supported since they will need a special implementation of WebSocketStream for this to work.
+            testWebSocket: false)]
+        public async Task AsyncClientAndServiceMustNotUseSyncNetworkIOTest(ClientAndServiceTestCase clientAndServiceTestCase)
+        {
+            var syncIoRecordingStreamFactory = new SyncIoRecordingStreamFactory(AsyncHalibutFeature.Enabled);
+
+            await using (var clientAndService = await clientAndServiceTestCase.CreateTestCaseBuilder()
+                       .WithStandardServices()
+                       .AsLatestClientAndLatestServiceBuilder()
+                       .WithStreamFactory(syncIoRecordingStreamFactory)
+                       .Build(CancellationToken))
+            {
+                var echo = clientAndService.CreateClient<IEchoService, IAsyncClientEchoService>();
+                (await echo.SayHelloAsync("Deploy package A")).Should().Be("Deploy package A...");
+
+                for (var i = 0; i < clientAndServiceTestCase.RecommendedIterations; i++)
+                {
+                    (await echo.SayHelloAsync($"Deploy package A {i}")).Should().Be($"Deploy package A {i}...");
+                }
+            }
+
+            var distinctPlaces = syncIoRecordingStreamFactory.PlacesSyncIoWasUsed()
+                .Select(n =>
+                {
+                    var lines = n.ToString().Split('\n')
+                        .Where(s => !s.ToString().Contains("System.Threading.Tasks"))
+                        .Where(s => !s.ToString().Contains("AsyncTaskMethodBuilder"))
+                        .Where(s => !s.ToString().Contains("System.Runtime.CompilerServices."))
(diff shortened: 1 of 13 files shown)


source/Halibut.Tests/AsyncClientAndServiceMustNotUseSyncNetworkIO.cs
```diff
@@ -0,0 +1,80 @@
...
+using NUnit.Framework;
+
+namespace Halibut.Tests
+{
+    public class AsyncClientAndServiceMustNotUseSyncNetworkIO : BaseTest
+    {
+        [Test]
+        [LatestClientAndLatestServiceTestCases(testSyncService: false, testAsyncServicesAsWell: true,
+            testSyncClients: false, testAsyncClients: true,
+            // TODO: ASYNC ME UP!
```
nathanwoctopusdeploy, line 20: ?

nathanwoctopusdeploy: testAsyncServicesAsWell -> testAsyncServices. Do we need the As Well?

source/Halibut.Tests/Support/Streams/NoSyncIOStream.cs
```diff
@@ -0,0 +1,103 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Threading.Tasks;
+using Halibut.Transport.Streams;
+
+namespace Halibut.Tests.Support.Streams
+{
+    public class NoSyncIoStream : DelegateStreamBase
```
nathanwoctopusdeploy, line 10: Can this describe what it actually does? It doesn't do 'NoSyncIo', it just logs when sync IO occurs

source/Halibut/HalibutRuntimeBuilder.cs
```diff
@@ -21,7 +22,19 @@ public class HalibutRuntimeBuilder
         AsyncHalibutFeature asyncHalibutFeature = AsyncHalibutFeature.Disabled;
         Func<string, string, UnauthorizedClientConnectResponse> onUnauthorizedClientConnect;
         HalibutTimeoutsAndLimits halibutTimeoutsAndLimits;
-
+        IBaseStreamWrapper baseStreamWrapper = new NoOpBaseStreamWrapper();
```
nathanwoctopusdeploy, line 25: I'm not too fond of this. Can this not be null if it's not set and when null it does not change the call chains

source/Halibut/HalibutRuntimeBuilder.cs
```diff
@@ -21,7 +22,19 @@ public class HalibutRuntimeBuilder
...
+
+        /// <summary>
+        /// For testing only.
+        /// </summary>
+        /// <param name="baseStreamWrapper"></param>
+        /// <returns></returns>
+        internal HalibutRuntimeBuilder WithBaseStreamWrapper(IBaseStreamWrapper baseStreamWrapper)
```
nathanwoctopusdeploy, line 32: I don't know if BaseStreamWrapper really helps describe what it does. I think a good comment is needed here

nathanwoctopusdeploy (changes requested):
I see the value in this test but I really don't like how invasive it is in Halibut to do this.

Are there other options? Even some nasty reflection to manipulate the Halibut Runtime / manipulate something that is exposed lower down in Halibut. Having to support this from the top down just for a test seems like a lot of noise for future readers to understand the reasoning behind.

nathanwoctopusdeploy:
> I see the value in this test but I really don't like how invasive it is in Halibut to do this.
> 
> Are there other options? Even some nasty reflection to manipulate the Halibut Runtime / manipulate something that is exposed lower down in Halibut. Having to support this from the top down just for a test seems like a lot of noise for future readers to understand the reasoning behind.

Could we do it with a conditional compilation directive? It's invasive and could change behaviour, so we would not want all tests to run like this, but a single test could be run with a specific compile time directive, separate nuke test -> build to compile with the directive etc. When the compilation directive is defined we wrap the base streams #if #endif for all related code and classes so it's guaranteed to be off by default. Constructor could add the IO logging stream to a static collection so the test can access it etc.

This would produce a lot less changes and also mean we only have those changes for one specific test.

source/Halibut/Transport/WebSocketConnectionFactory.cs
```diff
@@ -9,22 +9,25 @@
...
 using Halibut.Transport.Proxy.Exceptions;
+using Halibut.Transport.Streams;
 using Halibut.Util;
 
 namespace Halibut.Transport
 {
     public class WebSocketConnectionFactory : IConnectionFactory
     {
         readonly X509Certificate2 clientCertificate;
+        IStreamFactory streamFactory;
```
LukeButters, line 20: The test caught that we didn't use the factory in all places.

source/Halibut/Transport/WebSocketConnectionFactory.cs
```diff
@@ -55,7 +58,7 @@ public async Task<IConnection> EstablishNewConnectionAsync(ExchangeProtocolBuild
 
             log.Write(EventType.Diagnostic, "Connection established");
 
-            var stream = new WebSocketStream(client);
+            var stream = streamFactory.CreateStream(client);
```
LukeButters, line 61: This is in `EstablishNewConnectionAsync` a new async method.

source/Halibut.Tests/AsyncClientAndServiceMustNotUseSyncNetworkIO.cs
```diff
@@ -0,0 +1,122 @@
...
+            syncIoRecordingStreamFactory.streams.Count.Should().BeGreaterOrEqualTo(2, "Since we should wrap a stream on the client and the service.");
+        }
+
+        internal static string[] DistinctStackTracesOfSyncIo(SyncIoRecordingStreamFactory syncIoRecordingStreamFactory)
+        {
+            return syncIoRecordingStreamFactory.PlacesSyncIoWasUsed()
+                .Select(n =>
+                {
+                    var lines = n.ToString().Split('\n')
+                        .Where(s => !s.ToString().Contains("System.Threading.Tasks"))
```
nathanwoctopusdeploy, line 76: Is this just filtering a noisy stack trace?

nathanwoctopusdeploy (approved):
Nice, LGTM

PR #440, opened 2023-08-18
NetworkTimeoutStream should Close sync calls on Timeout

nathanwoctopusdeploy:
# Background

Make the NetworkTimeoutStream behave the same for sync calls as it does for async calls by closing the stream on timeout.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Transport/Streams/CallCountingStream.cs
+++ b/source/Halibut.Tests/Transport/Streams/CallCountingStream.cs
@@ -1,0 +1,185 @@
+﻿#nullable enable
+using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Halibut.Transport.Streams;
+#if NETFRAMEWORK
+using System.Runtime.Remoting;
+#endif
+
+namespace Halibut.Tests.Transport.Streams
+{
+    class CallCountingStream : AsyncDisposableStream
+    {
+        public volatile int CloseCallCount = 0;
+
+        public volatile int DisposeBoolCallCount = 0;
+        public volatile int DisposeAsyncCallCount = 0;
+
+        readonly Stream inner;
+
+        public CallCountingStream(Stream inner)
+        {
+            this.inner = inner;
+        }
+
+        public override async ValueTask DisposeAsync()
+        {
+            Interlocked.Increment(ref DisposeAsyncCallCount);
+
+            await inner.DisposeAsync();
+        }
+
+        protected override void Dispose(bool disposing)
+        {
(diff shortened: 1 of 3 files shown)


source/Halibut.Tests/Transport/Streams/CallCountingStream.cs
```diff
@@ -0,0 +1,185 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Halibut.Transport.Streams;
+#if NETFRAMEWORK
+using System.Runtime.Remoting;
+#endif
+
+namespace Halibut.Tests.Transport.Streams
+{
+    class CallCountingStream : AsyncDisposableStream
```
nathanwoctopusdeploy, line 13: Override all the methods! Again!

Another candidate to slim down when we have just the things you need to override as part of https://app.shortcut.com/octopusdeploy/story/57119/make-all-streams-async-friendly-you-can-t-mess-it-up

nathanwoctopusdeploy:
[sc-57119]

source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
```diff
@@ -186,22 +138,53 @@ public override void EndWrite(IAsyncResult asyncResult)
...
+            try
+            {
+                return inner.Read(buffer);
+            }
+            catch (Exception ex) when (IsTimeoutException(ex))
+            {
+                inner.Close();
+                throw;
+            }
+        }
```
acodrington, line 174: You should use that nice `CloseOnTimeout` overload you just wrote! Same for the `Write` override below

acodrington (approved):

source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
nathanwoctopusdeploy: You can't as you can't use the byref-like structure (Span and ReadOnlySpan) in a lambda

acodrington: 😞

source/Halibut.Tests/Transport/Streams/CallCountingStream.cs
LukeButters: This can be done if we either don't count calls to the old style async BegineRead/end methods.

or if the async stream no longer marks those streams as sealed:
image

They are marked sealed on the assumption no one will ever need to care about them

PR #445, opened 2023-08-18
Adds a async base stream

LukeButters:
# Background

Creating streams that are actually async can be difficult, since it is not clear what methods need to be implemented. This PR adds a new `AsyncStream` which helps with that. The `AsyncStream` will force the required async methods to be implemented as well as take care of the old style async programming APM to also be async.

This PRs changes some classes (Not including NetworkTimeoutStream) to implement `AsyncStream`.

This PR makes use of the changes added in https://github.com/OctopusDeploy/Halibut/pull/437

[SC-57119]

# Results

Appears to be no change since the APM stream was doing this work at just the right place. This change removes the need to wrap again and instead defaults our streams to be ones that work with async.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Halibut/Transport/Streams/WriteIntoMemoryBufferStream.cs
```diff
@@ -94,7 +99,12 @@ public override int WriteTimeout
         public override void SetLength(long value) => throw new NotSupportedException();
         public override int Read(byte[] buffer, int offset, int count) => throw new NotSupportedException();
 
-        public override async Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        protected override Task<int> _ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            return innerStream.ReadAsync(buffer, offset, count, cancellationToken);
```
sburmanoctopus, line 104: Do we actually read on this stream? Or should it do what `Read` does, and `throw new NotSupportedException();`?

source/Halibut.Tests/Transport/Streams/AsyncStreamFixture.cs
```diff
@@ -0,0 +1,189 @@
...
+            return stream.FlushAsync(cancellationToken);
+        }
+
+        protected override ValueTask _DisposeAsync()
+        {
+            return stream.DisposeAsync();
+        }
+    }
+
+    public abstract class StreamWrapperSupportsAsyncIOFixture : BaseTest
```
sburmanoctopus, line 88: Nit: I think this would be cleaner in a new file

source/Halibut.Tests/Transport/Streams/AsyncStreamFixture.cs
```diff
@@ -0,0 +1,189 @@
...
+using NUnit.Framework;
+
+namespace Halibut.Tests.Transport.Streams
+{
+    public class AsyncStreamFixture : StreamWrapperSupportsAsyncIOFixture
+    {
+        protected override AsyncStream WrapStream(Stream stream) => new NoOpAsyncStream(stream);
+    }
+
+    public class NoOpAsyncStream : AsyncStream
```
sburmanoctopus, line 19: Nit: I think this would be cleaner in a new file

source/Halibut.Tests/Support/Streams/DelegateStreamBase.cs
```diff
@@ -0,0 +1,106 @@
+﻿using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Halibut.Tests.Support.Streams
+{
+    public abstract class DelegateStreamBase :  Stream
+    {
+        public abstract Stream Inner { get; }
```
sburmanoctopus, line 10: Nit: Could make this an actual property (protected even?), so we don't have to implement it

source/Halibut/Transport/Streams/AsyncStream.cs
```diff
@@ -0,0 +1,93 @@
...
+    ///     net48 Stream does not implement IAsyncDisposable.
+    /// </summary>
+    public abstract class AsyncStream : AsyncDisposableStream
+    {
+        public sealed override async Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            return await _ReadAsync(buffer, offset, count, cancellationToken);
+        }
+
+        protected abstract Task<int> _ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken);
```
sburmanoctopus, line 24: Nit: I love starting with a good underscore, but generally that's for member variables in .NET

Secondly, this goes against our style guide. As much as we would like to change it, I think it would be better to give these conforming names.

🤔 `ReadAsyncStream`? 🤷

sburmanoctopus (approved):
LGTM. I would like the underscore names changed. But otherwise, all good.

source/Halibut/Transport/Streams/AsyncStream.cs
LukeButters: Yeah I tried doing something like that. In the end it resulted in clutter in the child classes. Clutter that no one cares about in the child classes. It isn't helpful to see `ReadAsyncStream`.

source/Halibut/Transport/Streams/WriteIntoMemoryBufferStream.cs
LukeButters: Good point done

source/Halibut/Transport/Streams/AsyncStream.cs
nathanwoctopusdeploy: @LukeButters Can we revisit the naming please as discussed Sync and @sburmanoctopus has also raised it. The _Method does not look as expected to the reader and in my opinion looks confusing and like a mistake in the name.

LukeButters: PR to clean that up https://github.com/OctopusDeploy/Halibut/pull/455

source/Halibut.Tests/Transport/Streams/AsyncStreamFixture.cs
LukeButters: Done here https://github.com/OctopusDeploy/Halibut/pull/457

PR #446, opened 2023-08-18
Improve the NetworkTimeoutStream closing and disposal on timeout

nathanwoctopusdeploy:
# Background

Make the NetworkTimeoutStream behave the same for sync calls as it does for async calls by closing the stream on timeout.
- Due to the way some of the streams work e.g. the `DeflateStream` `Dispose` will cause it to `PurgeBuffers`, it has been necessary to make the `NetworkTimeoutStream` handle the case where it is used after it has been closed due to a timeout. Without this logic Halibut will throw an Already Disposed Exception which will look like there is a bug.
image
- The `NetworkTimeoutStream` now tracks when it has timed out along with recording the exception that was throw. If it is used after timeout it will throw the original exception to the caller.

NetworkTimeoutStream should DisposeAsync of the inner stream instead of Close when an async operation is being called.

## Under net48

image

image

## Under net6.0

image

image

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Timeouts/SendingAndReceivingRequestMessagesTimeoutsFixture.cs
+++ b/source/Halibut.Tests/Timeouts/SendingAndReceivingRequestMessagesTimeoutsFixture.cs
@@ -168,16 +168,33 @@ namespace Halibut.Tests.Timeouts
                     addControlMessageTimeout += HalibutLimits.TcpClientHeartbeatSendTimeout;
                 }
 
-                var expectedTimeOut = HalibutLimits.TcpClientSendTimeout // What we actually expected. 
-                                     + HalibutLimits.TcpClientSendTimeout // an extra timeout because of the dispose method of the zip stream (see below)
-                                     + addControlMessageTimeout;
-                sw.Elapsed.Should().BeGreaterThan( expectedTimeOut- TimeSpan.FromSeconds(2), 
-                        "We 'should' wait the send timeout amount of time NOT the heart beat timeout, however when an error occurs writing to the zip (deflate)" +
-                                "stream we also call dispose which again attempts to write to the stream. Thus we wait 2 times the TcpClientSendTimeout.") // -2s give it a little slack to avoid it timed out slightly too early.
-                    .And
-                    .BeLessThan(expectedTimeOut + LowerHalibutLimitsForAllTests.HalfTheTcpReceiveTimeout, 
-                        "We 'should' wait the send timeout amount of time, however when an error occurs writing to the zip (deflate)" +
+                if (clientAndServiceTestCase.SyncOrAsync == SyncOrAsync.Sync)
+                {
+                    var expectedTimeOut = HalibutLimits.TcpClientSendTimeout // What we actually expected. 
+                                          + HalibutLimits.TcpClientSendTimeout // an extra timeout because of the dispose method of the zip stream (see below)
+                                          + addControlMessageTimeout;
+
+                    sw.Elapsed.Should().BeGreaterThan(expectedTimeOut - TimeSpan.FromSeconds(2),
+                            "We 'should' wait the send timeout amount of time NOT the heart beat timeout, however when an error occurs writing to the zip (deflate)" +
+                            "stream we also call dispose which again attempts to write to the stream. Thus we wait 2 times the TcpClientSendTimeout.") // -2s give it a little slack to avoid it timed out slightly too early.
+                        .And
+                        .BeLessThan(expectedTimeOut + LowerHalibutLimitsForAllTests.HalfTheTcpReceiveTimeout,
+                            "We 'should' wait the send timeout amount of time, however when an error occurs writing to the zip (deflate)" +
                             "stream we also call dispose which again attempts to write to the stream. Thus we wait 2 times the TcpClientSendTimeout.");
+                }
+                else
+                {
+                    
+                    var expectedTimeOut = HalibutLimits.TcpClientSendTimeout;
+
+                    sw.Elapsed.Should().BeGreaterThan(
+                            expectedTimeOut - TimeSpan.FromSeconds(2), 
+                            "Should wait the send timeout amount of time NOT the heart beat timeout") // -2s give it a little slack to avoid it timed out slightly too early.
+                        .And
+                        .BeLessThan(
+                            expectedTimeOut + LowerHalibutLimitsForAllTests.HalfTheTcpReceiveTimeout,
+                            "Should wait the send timeout amount of time");
+                }
 
                 await echo.SayHelloAsync("A new request can be made on a new unpaused TCP connection");
             }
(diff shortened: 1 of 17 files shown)


source/Halibut/Diagnostics/ExceptionReturnedByHalibutProxyExtensionMethod.cs
```diff
@@ -62,9 +62,9 @@ public static HalibutNetworkExceptionType IsNetworkError(this Exception exceptio
...
+                if (exception.Message.Contains("An existing connection was forcibly closed by the remote host.")) return HalibutNetworkExceptionType.IsNetworkError;
                 if (exception.Message.Contains("The I/O operation has been aborted because of either a thread exit or an application request")) return HalibutNetworkExceptionType.IsNetworkError;
-                if (exception.Message.Contains("Unable to read data from the transport connection: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.")) return HalibutNetworkExceptionType.IsNetworkError;
+                if (exception.Message.Contains("A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.")) return HalibutNetworkExceptionType.IsNetworkError;
```
nathanwoctopusdeploy, line 67: The exceptions missed the write counterparts. Making less specific to catch both

nathanwoctopusdeploy:
[sc-56625] [sc-57121]

source/Halibut.Tests/Timeouts/SendingAndReceivingRequestMessagesTimeoutsFixture.cs
```diff
@@ -159,16 +159,33 @@ public async Task WhenThenNetworkIsPaused_WhileSendingARequestMessage_ATcpWriteT
...
-                var expectedTimeOut = HalibutLimits.TcpClientSendTimeout // What we actually expected. 
-                                     + HalibutLimits.TcpClientSendTimeout // an extra timeout because of the dispose method of the zip stream (see below)
-                                     + addControlMessageTimeout;
-                sw.Elapsed.Should().BeGreaterThan( expectedTimeOut- TimeSpan.FromSeconds(2), 
-                        "We 'should' wait the send timeout amount of time NOT the heart beat timeout, however when an error occurs writing to the zip (deflate)" +
-                                "stream we also call dispose which again attempts to write to the stream. Thus we wait 2 times the TcpClientSendTimeout.") // -2s give it a little slack to avoid it timed out slightly too early.
-                    .And
-                    .BeLessThan(expectedTimeOut + LowerHalibutLimitsForAllTests.HalfTheTcpReceiveTimeout, 
-                        "We 'should' wait the send timeout amount of time, however when an error occurs writing to the zip (deflate)" +
+                if (clientAndServiceTestCase.ForceClientProxyType == ForceClientProxyType.SyncClient)
```
sburmanoctopus, line 162: I think the `SyncOrAsync` flag on the test case would be better to use.

FYI, I did a similar change in [my last PR](https://github.com/OctopusDeploy/Halibut/pull/436/files) in this area. It might be worth rebasing and adjusting appropriately.

source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
```diff
@@ -161,30 +150,105 @@ public override void EndWrite(IAsyncResult asyncResult)
...
 
 #if !NETFRAMEWORK
-        public override void CopyTo(Stream destination, int bufferSize) => inner.CopyTo(destination, bufferSize);
-        
-        public override int Read(Span<byte> buffer) => inner.Read(buffer);
+        public override void CopyTo(Stream destination, int bufferSize)
+        {
+            ThrowIfAlreadyTimedOut();
+            
+            inner.CopyTo(destination, bufferSize);
```
LukeButters, line 194: Not sure this is correct.

We probably do need to be doing a `TryCloseOnTimeout` of course we wont know which stream had the timeout. 

I think to get around that the solution would be to do, `base.CopyTo` since that would then make calls to `read` on `this` which would then do the `TryCloseOnTimeout` work.

source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
```diff
@@ -87,42 +94,24 @@ public override async ValueTask WriteAsync(ReadOnlyMemory<byte> buffer, Cancella
...
+            ThrowIfAlreadyTimedOut();
+            
+            await inner.CopyToAsync(destination, bufferSize, cancellationToken);
+        }
 
-        public override async Task CopyToAsync(Stream destination, int bufferSize, CancellationToken cancellationToken) => await inner.CopyToAsync(destination, bufferSize, cancellationToken);
-        
-        public override int ReadByte() => inner.ReadByte();
+        public override int ReadByte()
+        {
```
LukeButters, line 112: I wonder if `TryCloseOnTimeout` could be done here.

LukeButters: nevermind

source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
```diff
@@ -87,42 +94,24 @@ public override async ValueTask WriteAsync(ReadOnlyMemory<byte> buffer, Cancella
...
+            
+            inner.Close();
         }
 
-        public override void Close() => inner.Close();
+        public override async Task CopyToAsync(Stream destination, int bufferSize, CancellationToken cancellationToken)
+        {
+            ThrowIfAlreadyTimedOut();
+            
+            await inner.CopyToAsync(destination, bufferSize, cancellationToken);
```
LukeButters, line 108: This will be a problem since the `inner` wont respect timeouts.

```
await base.CopyToAsync(destination, buffferSize, cancellationToken)
```

Would fix that.

If this is not fixed in this PR at least mark it as TODO before merging.

LukeButters (approved):

source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
```diff
@@ -161,30 +150,105 @@ public override void EndWrite(IAsyncResult asyncResult)
             }
         }
 
-        public override void Flush() => inner.Flush();
+        public override void Flush()
+        {
+            ThrowIfAlreadyTimedOut();
+            
+            inner.Flush();
```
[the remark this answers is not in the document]
nathanwoctopusdeploy: The [Flush on NetworkStream](https://learn.microsoft.com/en-us/dotnet/api/system.net.sockets.networkstream.flush?view=net-6.0) doesn't do anything so I didn't wrap here, but as we don't enforce the inner has to be a NetworkStream it should probably not take that into account and do the wrapping 👍

sburmanoctopus (approved):
LGTM

PR #448, opened 2023-08-18
Make WebSockets Timeout

acodrington:
[sc-57052]

# Background

Web sockets currently do not time out. This is a problem, as it means connections will hang.

# Results

## Before

Web sockets wrap a `WebSocket` context, and interact with that link a `Stream`. But they would never timeout the operations, which allowed them to go on forever.

## After

To ensure timeouts, we wrapped the `WebSocketStream` within a `NetworkTimeoutStream` to ensure that the operations would indeed time out.

### JSON Async vs Sync
JSON serialization is not async. So we read off/write to the stream asynchronously, and fall back to 'sync' reading/writing if the payload is too large.

But when this happened, web sockets never timed out (even with `NetworkTimeoutStream`). This is because it would used the synchronous versions of `Read` and `Write`, which the `NetworkTimeoutStream` cannot timeout with.

To solve this, we introduced the `CallUnderlyingAsyncMethodsStream`, so that when JSON called the synchronous `Read` or `Write` methods, they would use the async versions instead. This would allow the `NetworkTimeoutStream` to actually time out.

### WebSocketStream Special Methods
The `IStreamFactory.CreateStream` method for web sockets returned a `WebSocketStream`. This made it hard to wrap in the other streams, as `WebSocketStream` had some special methods for doing the initial read and write.

To solve this, we simply moved the special methods to `WebSocketExtensionMethods`.

### Tests
We turned on web sockets for all the tests we could (not just the timeout tests).

The only places we did **not** reenable web sockets were
- Tests that were also disabling regular polling (web sockets wouldn't work anyway)
- Tests that explicitly commented that they could not work with web sockets.
- Tests in `TimeoutsApplyDuringHandShake`. There were issues, so we made a [separate card](https://app.shortcut.com/octopusdeploy/story/66523/make-timeoutsapplyduringhandshake-work-with-web-sockets) for that.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Halibut/Transport/Streams/CallUnderlyingAsyncMethodsStream.cs
```diff
@@ -0,0 +1,147 @@
+﻿using System.IO;
+using System.Threading.Tasks;
+using System.Threading;
+using System.Runtime.Remoting;
+using System;
+
+namespace Halibut.Transport.Streams
+{
+    class CallUnderlyingAsyncMethodsStream : AsyncStream
```
sburmanoctopus, line 9: To ensure we covered all the methods that `NetworkTimeoutStream` would call, we copied that class, and altered appropriately.

source/Halibut.Tests/TcpKeepAliveTests.cs
```diff
@@ -20,6 +20,7 @@ public class TcpKeepAliveTests : BaseTest
     {
         [Test]
         [LatestClientAndLatestServiceTestCases(testNetworkConditions: false, testListening:false, testWebSocket:false)]
+        // TCP Keep alives were for TcpClients, not web sockets
```
nathanwoctopusdeploy, line 23: Interesting find - side project: Do keep alives work as expected for WebSockets or did we miss something?

source/Halibut.Tests/Timeouts/SendingAndReceivingRequestMessagesTimeoutsFixture.cs
```diff
@@ -122,7 +123,8 @@ await using (var clientAndService = await clientAndServiceTestCase.CreateTestCas
         }
 
         [Test]
-        [LatestClientAndLatestServiceTestCases(testNetworkConditions: false, testWebSocket: false)]
+        [Timeout(120000)]
```
nathanwoctopusdeploy, line 126: nit: Was there a reason for defining the timeout here? Worth a comment if there is?

nathanwoctopusdeploy (approved):
LGTM, lovely job

source/Halibut.Tests/TcpKeepAliveTests.cs
sburmanoctopus: Good question. I've added a [card to the board](https://app.shortcut.com/octopusdeploy/story/66528/make-keep-alives-work-with-web-sockets-if-possible) to look into this.

source/Halibut.Tests/Timeouts/SendingAndReceivingRequestMessagesTimeoutsFixture.cs
sburmanoctopus: Nicely spotted. That was there when I picked up this PR. Talking to the author, it was likely a temporary thing. Plus, it runs fine on my PC without it. So we decided to deleted it 🙂

PR #450, opened 2023-08-18
Make the CompatBinary exit when the test times out if it is still running

nathanwoctopusdeploy:
# Background

Ensure the CompatBinary does not stay running for longer than the test timeout

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.TestUtils.CompatBinary.Base/BackwardsCompatProgramBase.cs
+++ b/source/Halibut.TestUtils.CompatBinary.Base/BackwardsCompatProgramBase.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Threading;
 using System.Threading.Tasks;
 using Halibut.TestUtils.SampleProgram.Base.LogUtils;
 using Halibut.TestUtils.SampleProgram.Base.Services;
@@ -9,16 +10,19 @@ namespace Halibut.TestUtils.SampleProgram.Base
     {
         public static async Task<int> Main()
         {
+            using var cancellationTokenSource = new CancellationTokenSource();
+            Task actionTask;
+
             var mode = SettingsHelper.GetSetting("mode") ?? string.Empty;
             Console.WriteLine($"Mode is: {mode}");
 
             if (mode.Equals("serviceonly"))
             {
-                await RunExternalService();
+                actionTask = RunExternalService(cancellationTokenSource.Token);
             }
             else if(mode.Equals("proxy"))
             {
-                await ProxyServiceForwardingRequestToClient.Run();
+                actionTask = ProxyServiceForwardingRequestToClient.Run(cancellationTokenSource.Token);
             }
             else
             {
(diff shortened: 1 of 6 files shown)


nathanwoctopusdeploy:
[sc-57128]

LukeButters (commented):
I think it might make sense to re-write this change.

Right now if the CT is cancelled which is passed to the `StayAliveUntilHelper` then that helper will stop detecting if the parent dies. We probably never want that.

The existing `StayAliveUntilHelper` also just calls `Environment.Exit` which kills the process immediately. It might make sense to do the same here rather than return an exit code just terminate the process.

source/Halibut.TestUtils.CompatBinary.Base/BackwardsCompatProgramBase.cs
```diff
@@ -9,27 +10,40 @@ public class BackwardsCompatProgramBase
...
             }
 
+            var timeoutTask = GetTestTimeoutTask(cancellationTokenSource.Token);
+            var completedTask = await Task.WhenAny(actionTask, timeoutTask);
+
+            if (completedTask == timeoutTask)
+            {
+                Console.WriteLine($"Program will exit as the TestTimeout of {SettingsHelper.GetTestTimeout()} has elapsed");
+                cancellationTokenSource.Cancel();
+                return -10060;
```
LukeButters, line 40: Use `Environment.Exit` no need for a graceful termination at timeout.

source/Halibut.TestUtils.CompatBinary.Base/Services/StayAliveUntilHelper.cs
```diff
@@ -1,17 +1,16 @@
...
 using System.IO;
-using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
 
 namespace Halibut.TestUtils.SampleProgram.Base.Services
 {
     public class StayAliveUntilHelper
     {
-        public static async Task WaitUntilSignaledToDie()
```
LukeButters, line 11: We probably never want to cancel this, the sub process should never continue if the parent dies.

nathanwoctopusdeploy:
> I think it might make sense to re-write this change.

Re-write may be a little strong here!! Is the concept completely wrong, or just a few things need tweaking!!

> Right now if the CT is cancelled which is passed to the `StayAliveUntilHelper` then that helper will stop detecting if the parent dies. We probably never want that.

Were cancelling to stop all the things and then exit. What use would the `StayAliveUntilHelper` be here, were on the exit path anyway. 

> The existing `StayAliveUntilHelper` also just calls `Environment.Exit` which kills the process immediately. It might make sense to do the same here rather than return an exit code just terminate the process.

Is there any difference given the implementation? This PR tries to cancel all the things and then exit. It seems to work when I tested it locally - hard coding a very short timeout for the binary and ensuring it had shut down unexpectedly during the test run.

LukeButters:
Yep returning in main will wait for some things to finish up before terminating. So `Environment.Exit` would make sense.

I think the change would be to not pass a CT to the `stayalivehelper` and in the top main method if the `When.Any` returns and it is the timeout then just call `Environment.Exit` since we want it to stop rather than wait around for things to finish.

Making those changes means this change is an additional way of exiting which doesn't degrade the existing termination logic.

LukeButters (approved):

PR #451, opened 2023-08-21
Mark sync operations with an async version Obsolete

nathanwoctopusdeploy:
# Background

Mark sync operations with an async version Obsolete

This PR also moves all our customer streams to the Streams folder

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Halibut/Transport/SecureListener.cs
```diff
@@ -105,6 +103,7 @@ public int Start()
             return ((IPEndPoint)listener.LocalEndpoint).Port;
         }
 
+        // TODO - ASYNC ME UP - Async Version Needed
```
nathanwoctopusdeploy, line 106: Card created for this

nathanwoctopusdeploy:
[sc-55554]

source/Halibut/DataStream.cs
```diff
@@ -133,11 +133,14 @@ public static DataStream FromStreamAsync(Stream source, Func<int, CancellationTo
...
             updateProgress ??= (_, _) => Task.CompletedTask;
 
+#pragma warning disable CS0612
             return FromStream(source, 
                 i => updateProgress(i, CancellationToken.None).GetAwaiter().GetResult(),
                 updateProgress);
+#pragma warning restore CS0612
         }
 
+        [Obsolete]
```
LukeButters, line 143: This might not need to be marked as obsolete since its only an issue if the `updateProgress` implementation is sync. But its probs safer to make everything async.

source/Halibut/DataStream.cs
```diff
@@ -149,12 +152,14 @@ public static DataStream FromStream(Stream source, Action<int> updateProgress)
             });
         }
         
+        [Obsolete]
         public static DataStream FromStream(Stream source, Action<int> updateProgress, Func<int, CancellationToken, Task> updateProgressAsync)
```
LukeButters, line 156: Should this be marked as obsolete? The caller gives it everything that is required and halibut will call the sync or async methods based on what its doing. We might want to mark this as obsolete at the time sync is removed from halibut.

This method would be used now in Octopus since it would give us a better feature toggle, as we can keep the sync implementation of `updateProgress`.

source/Halibut/DataStream.cs
```diff
@@ -149,12 +152,14 @@ public static DataStream FromStream(Stream source, Action<int> updateProgress)
...
         {
             var streamer = new StreamingDataStream(source, updateProgress, updateProgressAsync);
             return new DataStream(source.Length, streamer.CopyAndReportProgress, streamer.CopyAndReportProgressAsync);
         }
 
+        [Obsolete]
         public static DataStream FromStream(Stream source)
```
LukeButters, line 163: This is not `obsolete` streams have async methods that will be used when halibut is in async mode.

source/Halibut/DataStream.cs
nathanwoctopusdeploy: > We might want to mark this as obsolete at the time sync is removed from halibut.

This may be a little late, as it would be deleted!! 

It's to ensure that we are aware of all callers to this operation, as it does have sync bits.

LukeButters: I was thinking the existing code would be:
```
DataStream.FromStream(source, ExistingSyncProgressMethod);
```
Then with this method we could have:
```
DataStream.FromStream(source, ExistingSyncProgressMethod, NewAsyncProgressMessage);
```
So `NewAsyncProgressMessage` will only be called if async is enabled. If sync is enabled then the existing `ExistingSyncProgressMethod` would be called. That is the toggle.

If we don't use this method now then we would end up with:
```
DataStream.FromStream(source, NewAsyncProgressMessage);
```
and the toggle would not let us revert to `ExistingSyncProgressMethod`

LukeButters (approved):

source/Halibut/DataStream.cs
nathanwoctopusdeploy: Ok, I'll remove this obsolete

PR #453, opened 2023-08-21
Ensure the NetworkTimeoutStream always closes the inner Stream on timeout for async operations

nathanwoctopusdeploy:
# Background

Under net6.0, when performing async operations on a NetworkStream, a cancellation token can be used to stop the async IO operation. Under net48 this cancellation token is not fully respected and will only cancel the operation if the cancellation token is cancelled before the method is called. The NetworkTimeoutStream was added to address this inconsistent behaviour as well as enforcing timeouts for async IO operations which do not respect the stream timeouts. A decision was made to always close the stream if cancellation or timeout occurred to ensure consistent behaviour and stop the stream from being used after a request timed out or was cancelled.

This PR fixes a possible race condition in the `NetworkTimeoutStream` that would introduce inconsistent behaviour between net48 and net6.0. In a timeout or cancellation scenario, if the IO operation task completed before the timeout task it is possible that the stream would not be closed (disposed). This differs from the behaviour if the timeout task completes, where the stream will be closed.

As this is a race condition in the implementation it is difficult to add a test for this scenario. It was manually verified to work by changing the implementation to ensure the race condition occurred and verify all the tests under net6.0 pass and some tests under net48 fail (as the NetworkStream doesn't respect cancellation so the race condition can't occur).

image

This PR caters for the race condition. An alternative to this would be to change the implementation so that the net6.0 implementation is different to the net4.8 implementation, and so remove the race condition.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Transport/Protocol/WebSocketStream.cs
+++ b/source/Halibut/Transport/Protocol/WebSocketStream.cs
@@ -14,7 +14,7 @@ namespace Halibut.Transport.Protocol
     {
         readonly WebSocket context;
         bool isDisposed;
-        readonly CancellationTokenSource cancel = new CancellationTokenSource();
+        readonly CancellationTokenSource cancel = new();
 
         static readonly TimeSpan SendCancelTimeout = TimeSpan.FromSeconds(1);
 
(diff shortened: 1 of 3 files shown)


nathanwoctopusdeploy:
[sc-56630]

sburmanoctopus (approved):
LGTM

PR #454, opened 2023-08-21
Changing sync method to be async

sburmanoctopus:
# Background

Async progress on the DataStream was never called!

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/UsageFixture.cs
+++ b/source/Halibut.Tests/UsageFixture.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


source/Halibut.Tests/UsageFixture.cs
```diff
@@ -136,7 +137,8 @@ await using (var clientAndService = await clientAndServiceTestCase.CreateTestCas
         }
 
         [Test]
-        [LatestAndPreviousClientAndServiceVersionsTestCases(testNetworkConditions: false)]
```
LukeButters, line 139: It doesn't make sense to measure progress when testing an old client, since in that case we are actually testing latest to old service. Which is a test we already have.

LukeButters (approved):

PR #456, opened 2023-08-21
WriteIntoMemoryBufferStream should not try to write bytes twice

LukeButters:
# Background

I am seeing exceptions in the dispose method of the `WriteIntoMemoryBufferStream`. This is likely to be because `WriteBufferToUnderlyingStream` fails and then it tries to do the same write again in `dispose`.

We should never attempt to re-write the same bytes on a stream after a stream has thrown an exception.

[SC-57434]
# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Transport/Streams/WriteIntoMemoryBufferStream.cs
+++ b/source/Halibut/Transport/Streams/WriteIntoMemoryBufferStream.cs
@@ -117,8 +117,8 @@ namespace Halibut.Transport.Streams
 
                 // We tried our best, but will no longer fit in memory. Transition over to use the sinkStream.
                 memoryBuffer.Position = 0;
-                await memoryBuffer.CopyToAsync(innerStream, 8192, cancellationToken);
                 usingMemoryBuffer = false;
+                await memoryBuffer.CopyToAsync(innerStream, 8192, cancellationToken);
             }
 
             await innerStream.WriteAsync(buffer, offset, count, cancellationToken);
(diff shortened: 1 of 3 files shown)


nathanwoctopusdeploy (commented):
LGTM. Good to get @sburmanoctopus 👀 over it

source/Halibut/Transport/Streams/WriteIntoMemoryBufferStream.cs
```diff
@@ -150,9 +150,10 @@ public async Task WriteBufferToUnderlyingStream(CancellationToken cancellationTo
             if (usingMemoryBuffer)
             {
                 memoryBuffer.Position = 0;
-                await memoryBuffer.CopyToAsync(innerStream, 8192, cancellationToken);
-
+                // Mark we are not using the memory buffer here, since if the write below fails
```
sburmanoctopus, line 153: This makes sense, and will work in this case 😄 

But some follow-on thoughts:
- I think this issue can happen during a `Write` and `WriteAsync` operation as well. Maybe we need to fix those two also?
- Using `usingMemoryBuffer` works fine, and I'm ok if we want to keep using it. But it feels a little like it's serving 2 purposes:
 - Knowing which stream to write to when writing
 - Choosing to write on dispose
- Do you think introducing a new variable for writing on dispose would make things more clear? Or just be more noise 🤔

LukeButters: Probably more noise I think.

LukeButters: Did you want the other methods order changed?

sburmanoctopus: I think it's worthwhile. You found this issue for this particular code path, and I think it can happen the other way as well.

Since we're going to the effort to fix this one, I'd do the complete set.

You can verify by writing the tests, and then fixing it 😁

sburmanoctopus (approved):
LGTM

PR #460, opened 2023-08-22
Allowing ConnectionManager and ConnectionPool to be disposed of both synchronously and asynchronously

sburmanoctopus:
[sc-53245]

# Background

When we tried using the latest Halibut in Octopus Server, tests would time out.

# Results

## Before
Octopus Server would dispose of the Halibut runtime via Autofac, which would sense that the `IHalibutRuntime` implemented `IAsyncDisposable`.

But this completely ignores whether the feature toggle is turned on.

As a result, it would try to dispose a 'synchronous' Halibut 'asynchronously'. 

This would cause the `ConnectionManager` to be disposed of asynchronously, which throws an exception if that occurs. 

## After

We have now made both `ConnectionManager` and `ConnectionPool` be disposable asynchronously.

This solves the issue, but also, it is probably a good idea to allow both paths of disposal to be allowed.

While here, we also noticed that `CreateNewConnection` is never called (so it was deleted).

We also had to reinstate `OSSupportsWebSockets`, as this is used in Octopus Server.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


nathanwoctopusdeploy (approved):
LGTM

PR #466, opened 2023-08-22
Makes SecureListener dispose connections in async

LukeButters:
# Background

[SC-57238]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Transport/SecureListener.cs
+++ b/source/Halibut/Transport/SecureListener.cs
@@ -287,7 +287,7 @@ namespace Halibut.Transport
                 finally
                 {
                     SafelyRemoveClientFromTcpClientManager(client, clientName);
-                    SafelyCloseStream(stream, clientName);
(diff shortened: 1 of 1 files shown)


acodrington (approved):

source/Halibut/Transport/SecureListener.cs
```diff
@@ -287,7 +287,7 @@ async Task ExecuteRequest(TcpClient client)
                 finally
                 {
                     SafelyRemoveClientFromTcpClientManager(client, clientName);
-                    SafelyCloseStream(stream, clientName);
+                    await SafelyCloseStreamAsync(stream, clientName);
                     SafelyCloseClient(client, clientName);
```
LukeButters, line 291: No async close method exists on the TCPClient.

The test in `sast/write-a-test-to-find-sync-io` suggest that with these change sync IO is not done on the stream. So this is probably fine.

PR #468, opened 2023-08-22
HTTPProxy client no longer requires all headers to arrive together (async only)

LukeButters:
# Background

Fixes a bug in HTTPProxyClient in which it would leave parts of the HTTP header in the stream, which would then be passed to the SSL stream which would throw.

The problem is the existing implementation would assume that it is done getting bytes when the network stream happened to have no bytes available at the time it checked. This doesn't make any sense since bytes of the HTTP header can arrive with any timing. When a delay occurs, the HTTPProxyClient would assume it has no more bytes and proceed to checking what it has and passing the stream to the SslStream.

The fix is to simply keep reading until the end of the HTTP headers is signalled which per the spec (citation needed) is when an empty line is received ie when the last 4 bytes are `\r\n\r\n`.

# Results

The Halibut http proxy test library is updated to have the ability to trigger the exact issue. When proxy tests are run with that enabled, they fail. With the fix they pass. Note that only latest client and latest service both in async has the fix and so this new feature of the HTTP proxy is only used in that case.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Support/LatestClientAndLatestServiceBuilder.cs
+++ b/source/Halibut.Tests/Support/LatestClientAndLatestServiceBuilder.cs
@@ -44,7 +44,7 @@ namespace Halibut.Tests.Support
         Action<PendingRequestQueueFactoryBuilder>? pendingRequestQueueFactoryBuilder;
         Reference<PortForwarder>? portForwarderReference;
         Func<RetryPolicy>? pollingReconnectRetryPolicy;
-        Func<HttpProxyService>? proxyFactory;
+        ProxyFactory? proxyFactory;
         LogLevel halibutLogLevel = LogLevel.Info;
         ConcurrentDictionary<string, ILog>? clientInMemoryLoggers;
         ConcurrentDictionary<string, ILog>? serviceInMemoryLoggers;
@@ -243,15 +243,14 @@ namespace Halibut.Tests.Support
 
         public LatestClientAndLatestServiceBuilder WithProxy()
         {
-            this.proxyFactory = () =>
-            {
-                var options = new HttpProxyOptions();
-                var loggerFactory = new SerilogLoggerFactory(new SerilogLoggerBuilder().Build());
-
-                return new HttpProxyService(options, loggerFactory);
-            };
-
+            this.proxyFactory = new ProxyFactory();
             return this;
+        }
+
+        private bool ProxyDelaySendingSectionsOfHttpHeaders()
+        {
+            // We can only delay sending sections of the HTTP headers in versions that have the fix, which is only in async.
+            return this.serviceAsyncHalibutFeature == AsyncHalibutFeature.Enabled && this.forceClientProxyType == ForceClientProxyType.AsyncClient;
         }
 
         public LatestClientAndLatestServiceBuilder WithPendingRequestQueueFactory(Func<ILogFactory, IPendingRequestQueueFactory> pendingRequestQueueFactory)
@@ -381,7 +380,7 @@ namespace Halibut.Tests.Support
 
             var disposableCollection = new DisposableCollection();
             PortForwarder? portForwarder = null;
-            var httpProxy = proxyFactory?.Invoke();
+            var httpProxy = proxyFactory?.WithDelaySendingSectionsOfHttpHeaders(ProxyDelaySendingSectionsOfHttpHeaders()).Build();
             ProxyDetails? httpProxyDetails = null;
 
             if (httpProxy != null)
--- a/source/Halibut.Tests/Support/ProxyFactory.cs
+++ b/source/Halibut.Tests/Support/ProxyFactory.cs
@@ -18,7 +18,7 @@ namespace Halibut.Tests.Support
 
         public HttpProxyService Build()
         {
-            return new HttpProxyService(options, false, serilogLoggerFactory);
+            return new HttpProxyService(options, delaySendingSectionsOfHttpHeaders, serilogLoggerFactory);
         }
     }
 }
--- a/source/Halibut.Tests/Support/ProxyFactory.cs
+++ b/source/Halibut.Tests/Support/ProxyFactory.cs
@@ -1,0 +1,24 @@
+using Halibut.TestProxy;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+using Serilog.Extensions.Logging;
+
+namespace Halibut.Tests.Support
+{
+    public class ProxyFactory
+    {
+        HttpProxyOptions options = new HttpProxyOptions();
+        SerilogLoggerFactory serilogLoggerFactory = new SerilogLoggerFactory(new SerilogLoggerBuilder().Build());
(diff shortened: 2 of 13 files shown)


source/Halibut/Transport/Proxy/HttpProxyClient.cs
```diff
@@ -347,25 +347,31 @@ async Task SendConnectionCommandAsync(string host, int port, CancellationToken c
...
-
-            // create an byte response array  
-            var response = new byte[TcpClient.ReceiveBufferSize];
+  
+            
             var sbuilder = new StringBuilder();
 
-            do
+            var existingTimeout = stream.ReadTimeout;
+            stream.ReadTimeout = WAIT_FOR_DATA_TIMEOUT;
```
LukeButters, line 361: Proxy doesn't work with websockets so it can require that the stream supports timeouts. Which is what we should always have been doing.

source/Halibut.Tests/Support/LatestClientAndLatestServiceBuilder.cs
```diff
@@ -243,17 +243,16 @@ IClientAndServiceBuilder IClientAndServiceBuilder.WithProxy()
...
-                var loggerFactory = new SerilogLoggerFactory(new SerilogLoggerBuilder().Build());
-
-                return new HttpProxyService(options, loggerFactory);
-            };
-
+            this.proxyFactory = new ProxyFactory();
             return this;
         }
 
+        private bool ProxyDelaySendingSectionsOfHttpHeaders()
```
LukeButters, line 250: This is sort of hidden away, I wonder if the builder should expose some concept of enable setting on the proxy that stress the halibut client.

nathanwoctopusdeploy:
[sc-57553]

source/Halibut.Tests/Support/ProxyFactory.cs
```diff
@@ -0,0 +1,24 @@
...
+using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+using Serilog.Extensions.Logging;
+
+namespace Halibut.Tests.Support
+{
+    public class ProxyFactory
+    {
+        HttpProxyOptions options = new HttpProxyOptions();
+        SerilogLoggerFactory serilogLoggerFactory = new SerilogLoggerFactory(new SerilogLoggerBuilder().Build());
+        bool delaySendingSectionsOfHttpHeaders = true;
```
nathanwoctopusdeploy, line 11: What am I missing here!! `delaySendingSectionsOfHttpHeaders` isn't used!!

LukeButters: :facepalm: ugh failed to refactored!

LukeButters: That does work now:
image

You can see if the proxy builder is told to always use this then tests do fail

source/Halibut/Transport/Proxy/HttpProxyClient.cs
nathanwoctopusdeploy: Have you confirmed it doesn't work, I never managed to get it to work but does something magic happen in Server and it does just work?

nathanwoctopusdeploy: Will this somehow cause Server/Tentacle to blow up if a proxy is configured for WebSockets?

LukeButters: Oh I must have understood not getting it to work to mean it doesn't work in halibut.

Looking at the code it looks like it has support for it:

```
WebSocketConnectionFactory.cs:
async Task<ClientWebSocket> CreateConnectedClientAsync(ServiceEndPoint serviceEndpoint, CancellationToken cancellationToken)
        {
            if (!serviceEndpoint.IsWebSocketEndpoint)
                throw new Exception("Only wss:// endpoints are supported");
            var client = new ClientWebSocket();
            ...
            if (serviceEndpoint.Proxy != null)
                client.Options.Proxy = new WebSocketProxy(serviceEndpoint.Proxy);
```

Note that the proxy options are passed directly to the ClientWebSocket. It doesn't look like [HttpProxyClient.cs](https://github.com/OctopusDeploy/Halibut/pull/468/files#diff-f236e7eb8c2b940a8385632e889fc9145bed41a34b6d6f46be77e4858773be37) is involved at all.

image

LukeButters: I also am finding that http proxies do not work with websockets. Interestingly if I put break points into the methods of `WebSocketProxy` a halibut class which implements a dotnet interface `IWebProxy` those break points are never hit!

The `HttpProxyClient` is not involved in websockets as the code expects `ClientWebSocket` a dotnet class to do the work of connecting via the proxy (which it doesn't appear to do).

nathanwoctopusdeploy: That's the same behaviour as I observed. It does seem broken on Microsoft's side!!

nathanwoctopusdeploy: Should we safeguard anyway and not assume the stream can timeout? E.g. it gets fixed and suddenly we have a bug

nathanwoctopusdeploy: We could wrap it at this point with a SupportsTimeoutStream and then a NetworkTimeoutStream if it doesn't support timeouts as an interim fix while the WebSopcketStream doesn't support timeouts

source/Halibut/Transport/Proxy/HttpProxyClient.cs
```diff
@@ -347,25 +347,30 @@ async Task SendConnectionCommandAsync(string host, int port, CancellationToken c
...
+                var response = new byte[1]; // Read 1 byte at a time to prevent over read. (This is not efficient but this is only done once per stream)
+                while (!sbuilder.ToString().EndsWith("\r\n\r\n"))
+                {
+                    var bytes = await stream.ReadAsync(response, 0, response.Length, cancellationToken);
+                    sbuilder.Append(Encoding.UTF8.GetString(response, 0, bytes));
+                }
+            }
+            finally
+            {
+                stream.ReadTimeout = existingTimeout;
```
nathanwoctopusdeploy, line 372: A decorator to safely apply this timeout change and then revert would be nice ;)

source/Halibut/Transport/Proxy/HttpProxyClient.cs
LukeButters: I think it should be fine since we always have a network stream:
image

source/Halibut/Transport/Proxy/HttpProxyClient.cs
LukeButters: I thought you would never ask.

nathanwoctopusdeploy (approved):

PR #471, opened 2023-08-23
Add IStreamFactory to consolidate preping streams for halibut

LukeButters:
# Background

Adds a `StreamFactory` which creates streams from TcpClient or websockets in a way that halibut will work with.

[SC-55504]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Halibut/Transport/DiscoveryClient.cs
```diff
@@ -27,7 +34,7 @@ public ServiceEndPoint Discover(ServiceEndPoint serviceEndpoint, CancellationTok
             try
             {
                 var log = logs.ForEndpoint(serviceEndpoint.BaseUri);
-                using (var client = TcpConnectionFactory.CreateConnectedTcpClient(serviceEndpoint, log, cancellationToken))
+                using (var client = TcpConnectionFactory.CreateConnectedTcpClient(serviceEndpoint, streamFactory, log, cancellationToken))
```
LukeButters, line 37: This is a static method.

source/Halibut/Transport/DiscoveryClient.cs
```diff
@@ -58,9 +65,9 @@ public async Task<ServiceEndPoint> DiscoverAsync(ServiceEndPoint serviceEndpoint
             try
             {
                 var log = logs.ForEndpoint(serviceEndpoint.BaseUri);
-                using (var client = await TcpConnectionFactory.CreateConnectedTcpClientAsync(serviceEndpoint, halibutTimeoutsAndLimits, log, cancellationToken))
+                using (var client = await TcpConnectionFactory.CreateConnectedTcpClientAsync(serviceEndpoint, halibutTimeoutsAndLimits, streamFactory, log, cancellationToken))
```
LukeButters, line 68: This is a static method call.

source/Halibut/Transport/SecureClient.cs
```diff
@@ -68,7 +72,7 @@ public void ExecuteTransaction(ExchangeAction protocolHandler, CancellationToken
                     IConnection connection = null;
                     try
                     {
-                        connection = connectionManager.AcquireConnection(protocolBuilder, new TcpConnectionFactory(clientCertificate, asyncHalibutFeature, halibutTimeoutsAndLimits), ServiceEndpoint, log, cancellationToken);
+                        connection = connectionManager.AcquireConnection(protocolBuilder, new TcpConnectionFactory(clientCertificate, asyncHalibutFeature, halibutTimeoutsAndLimits, streamFactory), ServiceEndpoint, log, cancellationToken);
```
LukeButters, line 75: Note that the streamFactory is passed to the ctr of the TcpConnectionFactory

source/Halibut/Transport/SecureListeningClient.cs
```diff
@@ -67,7 +71,7 @@ public void ExecuteTransaction(ExchangeAction protocolHandler, CancellationToken
                     IConnection connection = null;
                     try
                     {
-                        connection = connectionManager.AcquireConnection(exchangeProtocolBuilder, new TcpConnectionFactory(clientCertificate, asyncHalibutFeature, halibutTimeoutsAndLimits), ServiceEndpoint, log, cancellationToken);
+                        connection = connectionManager.AcquireConnection(exchangeProtocolBuilder, new TcpConnectionFactory(clientCertificate, asyncHalibutFeature, halibutTimeoutsAndLimits, streamFactory), ServiceEndpoint, log, cancellationToken);
```
LukeButters, line 74: Note that the streamFactory is passed to the ctr of the TcpConnectionFactory

source/Halibut/Transport/TcpConnectionFactory.cs
```diff
@@ -82,7 +85,7 @@ public async Task<IConnection> EstablishNewConnectionAsync(ExchangeProtocolBuild
         }
 
         [Obsolete]
-        internal static TcpClient CreateConnectedTcpClient(ServiceEndPoint endPoint, ILog log, CancellationToken cancellationToken)
+        internal static TcpClient CreateConnectedTcpClient(ServiceEndPoint endPoint, IStreamFactory streamFactory, ILog log, CancellationToken cancellationToken)
```
LukeButters, line 88: This is a static method, and hence `streamFactory` is passed in.

source/Halibut/Transport/TcpConnectionFactory.cs
```diff
@@ -94,15 +97,15 @@ internal static TcpClient CreateConnectedTcpClient(ServiceEndPoint endPoint, ILo
...
+                client = new ProxyClientFactory(streamFactory)
                     .CreateProxyClient(log, endPoint.Proxy)
                     .WithTcpClientFactory(CreateTcpClient)
                     .CreateConnection(endPoint.BaseUri.Host, endPoint.BaseUri.Port, endPoint.TcpClientConnectTimeout, cancellationToken);
             }
             return client;
         }
         
-        internal static async Task<TcpClient> CreateConnectedTcpClientAsync(ServiceEndPoint endPoint, HalibutTimeoutsAndLimits halibutTimeoutsAndLimits, ILog log, CancellationToken cancellationToken)
+        internal static async Task<TcpClient> CreateConnectedTcpClientAsync(ServiceEndPoint endPoint, HalibutTimeoutsAndLimits halibutTimeoutsAndLimits, IStreamFactory streamFactory, ILog log, CancellationToken cancellationToken)
```
LukeButters, line 108: This is a static method, and hence `streamFactory` is passed in.

source/Halibut/Transport/DiscoveryClient.cs
sburmanoctopus: Thought for the future:

🤔 Looking at `TcpConnectionFactory`, it seems to serve 2 purposes. Creating `IConnections` or these static `TcpClients`.

Maybe later we could split this class up appropriately. I.e. move the statics into a `TcpClientFactory` class.

If we do that, then maybe we could even make that something we dependency inject instead.

Again, for later. I just wanted to share my idea 😁

source/Halibut/Transport/Proxy/HttpProxyClient.cs
```diff
@@ -163,6 +164,8 @@ public string ProxyName
         /// This property can be set prior to executing CreateConnection to use an existing TcpClient connection.
         /// </summary>
         public TcpClient TcpClient { get; set; }
+        
+        IStreamFactory streamFactory;
```
sburmanoctopus, line 168: Nit: This can be `readonly`

source/Halibut/Transport/DiscoveryClient.cs
LukeButters: Yeah I think re-thinking how this code is laid out would oneday makes sense.

I would encourage us to avoid DI that isn't compile time checked, or DI that makes anything slower or a DI setup that makes it hard to understand the lifetime of things. I guess don't make the code base more complex or less safe by throwing DI at it.

source/Halibut/Transport/SecureListener.cs
```diff
@@ -43,6 +43,7 @@ public class SecureListener : IDisposable
         readonly ExchangeActionAsync exchangeAction;
         readonly AsyncHalibutFeature asyncHalibutFeature;
         readonly HalibutTimeoutsAndLimits halibutTimeoutsAndLimits;
+        IStreamFactory streamFactory;
```
sburmanoctopus, line 46: Nit: This can be `readonly`

source/Halibut/Transport/SecureListeningClient.cs
```diff
@@ -22,20 +23,23 @@ class SecureListeningClient : ISecureClient
         readonly ExchangeProtocolBuilder exchangeProtocolBuilder;
         readonly AsyncHalibutFeature asyncHalibutFeature;
         readonly HalibutTimeoutsAndLimits halibutTimeoutsAndLimits;
-        
+        IStreamFactory streamFactory;
```
sburmanoctopus, line 26: Nit: This can be `readonly`

LukeButters: Good idea, done.

source/Halibut/Transport/SecureListener.cs
LukeButters: Good idea, done.

source/Halibut/Transport/Proxy/HttpProxyClient.cs
LukeButters: Good idea, done.

source/Halibut/Transport/SecureListeningClient.cs
sburmanoctopus: We added `streamFactory` to this class because we don't have DI, and it is needed for `TcpConnectionFactory`.

I wonder if instead, we should add the `TcpConnectionFactory` as a constructor dependency instead. This is why I think it makes sense:

- You could get rid of all the arguments needed for `TcpConnectionFactory` as arguments for this class
- `TcpConnectionFactory` doesn't appear to hold any state. So I don't think it needs to be recreated each time.
- It would clean this class up

LukeButters: I wasn't doing that to minimise changes in the PR, but that seems to make sense.

source/Halibut/Transport/SecureClient.cs
sburmanoctopus: Can we inject the `TcpConnectionFactory` here, and reduce the parameters given to this class.

Similar justification to the other place I already commented 😁

source/Halibut/Transport/Streams/StreamFactory.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace Halibut.Transport.Streams
+{
+    public class StreamFactory : IStreamFactory
+    {
+        public StreamFactory(AsyncHalibutFeature asyncHalibutFeature)
+        {
+            AsyncHalibutFeature = asyncHalibutFeature;
+        }
+
+        public AsyncHalibutFeature AsyncHalibutFeature { get; }
```
sburmanoctopus, line 16: Nit: My default is to make things private readonly unless needed. The lower the public surface, the simpler things are 😁. I don't think we need this exposed?

source/Halibut/Transport/SecureListeningClient.cs
LukeButters: It looks very nice in the end

sburmanoctopus (approved):
LGTM

source/Halibut/Transport/TcpConnectionFactory.cs
```diff
@@ -57,10 +59,11 @@ public async Task<IConnection> EstablishNewConnectionAsync(ExchangeProtocolBuild
             log.Write(EventType.OpeningNewConnection, $"Opening a new connection to {serviceEndpoint.BaseUri}");
 
             var certificateValidator = new ClientCertificateValidator(serviceEndpoint);
-            var client = await CreateConnectedTcpClientAsync(serviceEndpoint, halibutTimeoutsAndLimits, log, cancellationToken);
+            var client = await CreateConnectedTcpClientAsync(serviceEndpoint, halibutTimeoutsAndLimits, streamFactory, log, cancellationToken);
             log.Write(EventType.Diagnostic, $"Connection established to {client.Client.RemoteEndPoint} for {serviceEndpoint.BaseUri}");
             
-            var networkTimeoutStream = client.GetNetworkTimeoutStream();
+            var networkTimeoutStream = streamFactory.CreateStream(client);
```
sburmanoctopus, line 65: A small change I noticed is that here, we used to create the `NetworkTimeoutStream` always (regardless of async Halibut). Whereas now, we only create it if async Halibut is turned on.

I think it's ok, but wanted to draw your attention to it in case it's not 🙂

LukeButters: I think in practice `EstablishNewConnectionAsync` is only called when async halibut is enabled so it should always be the same.

sburmanoctopus: Ah, of course. Disregard 😁

source/Halibut/Transport/Streams/StreamFactory.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace Halibut.Transport.Streams
+{
+    public class StreamFactory : IStreamFactory
+    {
+        public StreamFactory(AsyncHalibutFeature asyncHalibutFeature)
+        {
+            AsyncHalibutFeature = asyncHalibutFeature;
+        }
+
+        AsyncHalibutFeature AsyncHalibutFeature { get; }
```
nathanwoctopusdeploy, line 16: nit: spacing

source/Halibut/Transport/SecureClient.cs
```diff
@@ -68,7 +66,7 @@ public void ExecuteTransaction(ExchangeAction protocolHandler, CancellationToken
                     IConnection connection = null;
                     try
                     {
-                        connection = connectionManager.AcquireConnection(protocolBuilder, new TcpConnectionFactory(clientCertificate, asyncHalibutFeature, halibutTimeoutsAndLimits), ServiceEndpoint, log, cancellationToken);
+                        connection = connectionManager.AcquireConnection(protocolBuilder, tcpConnectionFactory, ServiceEndpoint, log, cancellationToken);
```
nathanwoctopusdeploy, line 69: We are starting to reduce our "toggle off has no changes to sync halibut" approach. Our we consciously doing this now? Does it start to introduce more risk?

source/Halibut/HalibutRuntime.cs
```diff
@@ -74,6 +78,7 @@ public HalibutRuntime(IServiceFactory serviceFactory, X509Certificate2 serverCer
             invoker = new ServiceInvoker(serviceFactory);
 
             connectionManager = new ConnectionManager();
+            this.tcpConnectionFactory = new TcpConnectionFactory(serverCertificate, AsyncHalibutFeature, TimeoutsAndLimits, streamFactory);
```
nathanwoctopusdeploy, line 81: I would have avoided this change. Yes it makes it cleaner in some areas but it's starting to diverge from our attempt to make as few changes as possible to sync halibut

nathanwoctopusdeploy (commented):
Looks good. Let's just keep an eye on changes to sync halibut that are not essential. These are good improvements but could be left for a later stage to derisk async halibut changes and ensure we have a really good feature toggle story.

source/Halibut/HalibutRuntime.cs
LukeButters: Yeah I was worried about that as well, however its such a straight forward top level change that I become less concerned. The change felt worth it since it just simplified the code by a decent amount.

PR #473, opened 2023-08-23
Fix an issue with tests hanging when waiting for the external binary to exit

nathanwoctopusdeploy:
# Background

Fix an issue with tests hanging when waiting for the external binary to exit

When running all the PollingOverWebSocket tests locally, I was able to replicate hanging tests every run/couple of runs. Killing some of the CompatBinaries would normally make the tests continue with a handful of failures. 

With the fixes in this PR, I am able to run the full suite of PollingOverWebSocket tests over and over. It's still not perfect and often fails after 6+ runs (run until failure) but it is a lot more stable.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Support/BackwardsCompatibility/CompatBinaryStayAlive.cs
+++ b/source/Halibut.Tests/Support/BackwardsCompatibility/CompatBinaryStayAlive.cs
@@ -1,28 +1,30 @@
 using System;
 using System.IO;
+using Serilog;
 
 namespace Halibut.Tests.Support.BackwardsCompatibility
 {
     public class CompatBinaryStayAlive : IDisposable
     {
-
+        readonly ILogger logger;
         public static string StayAliveFilePathEnvVarKey = "CompatBinaryStayAliveFilePath";
         
         readonly TmpDirectory tmpDirectory;
         public string LockFile { get; }
         readonly FileStream fileStreamLock;
-        public CompatBinaryStayAlive()
+        public CompatBinaryStayAlive(ILogger logger)
         {
+            logger = logger.ForContext<CompatBinaryStayAlive>();
+
+            this.logger = logger;
             tmpDirectory = new TmpDirectory();
             LockFile = Path.Combine(tmpDirectory.FullPath, "compat-bin-lock");
             
             fileStreamLock = new FileStream(LockFile, FileMode.Create, FileAccess.ReadWrite, FileShare.None);
-
         }
 
         public void Dispose()
         {
-            var logger = new SerilogLoggerBuilder().Build().ForContext<CompatBinaryStayAlive>();
             try
             {
                 fileStreamLock.Dispose();
--- a/source/Halibut.Tests/Support/BackwardsCompatibility/HalibutTestBinaryRunner.cs
+++ b/source/Halibut.Tests/Support/BackwardsCompatibility/HalibutTestBinaryRunner.cs
@@ -176,9 +176,8 @@ namespace Halibut.Tests.Support.BackwardsCompatibility
 
             public void Dispose()
             {
-                compatBinaryStayAlive.Dispose();
                 cts.Cancel();
-                runningOldHalibutTask.GetAwaiter().GetResult();
+                compatBinaryStayAlive.Dispose();
                 tmpDirectory.Dispose();
             }
         }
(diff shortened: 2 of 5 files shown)


source/Halibut.Tests/Support/BackwardsCompatibility/CompatBinaryStayAlive.cs
```diff
@@ -1,28 +1,30 @@
 using System;
 using System.IO;
+using Serilog;
 
 namespace Halibut.Tests.Support.BackwardsCompatibility
 {
     public class CompatBinaryStayAlive : IDisposable
     {
-
+        readonly ILogger logger;
```
LukeButters, line 9: Wow fancy

LukeButters: Oh sorry I thought this was adding real logging to the external compatibility binary which just does crappy STDOUT logging :p

source/Halibut.Tests/Support/BackwardsCompatibility/HalibutTestBinaryRunner.cs
```diff
@@ -98,6 +125,7 @@ public async Task<RunningOldHalibutBinary> Run()
             catch (Exception)
             {
                 cts.Cancel();
+                cts.Dispose();
```
LukeButters, line 128: ha

source/Halibut.Tests/Support/BackwardsCompatibility/ProxyHalibutTestBinaryRunner.cs
```diff
@@ -184,9 +192,9 @@ public class RoundTripRunningOldHalibutBinary : IDisposable
 
             public void Dispose()
             {
-                compatBinaryStayAlive.Dispose();
                 cts.Cancel();
-                runningOldHalibutTask.GetAwaiter().GetResult();
+                cts.Dispose();
```
LukeButters, line 196: Is it the re-ordering that makes it reliable?

source/Halibut.Tests/Support/BackwardsCompatibility/ProxyHalibutTestBinaryRunner.cs
```diff
@@ -184,9 +192,9 @@ public class RoundTripRunningOldHalibutBinary : IDisposable
 
             public void Dispose()
             {
-                compatBinaryStayAlive.Dispose();
                 cts.Cancel();
-                runningOldHalibutTask.GetAwaiter().GetResult();
```
LukeButters, line 189: Or is it that we don't consume another thread waiting for the task to complete? Should this be `DisposeAsync`?

nathanwoctopusdeploy: Cancellation first I think wins and then there is no need to wait on that task.

It can still hang but not as frequently! I think there may be an issue with the way the binary is started and all the tasks that creates as I can cause the test to hang eventually by preventing the manual reset event from being signaled

source/Halibut.Tests/Support/BackwardsCompatibility/CompatBinaryStayAlive.cs
nathanwoctopusdeploy: I didn't like the construction of loggers in dispose methods so this is why there is this logging plumbing change

source/Halibut.Tests/Support/BackwardsCompatibility/HalibutTestBinaryRunner.cs
```diff
@@ -143,10 +171,13 @@ await Cli.Wrap(new HalibutTestBinaryPath().BinPath(version))
                 }
             }, cancellationToken);
 
-            await Task.WhenAny(runningTentacle, hasTentacleStarted.WaitAsync(cancellationToken), Task.Delay(TimeSpan.FromMinutes(1), cancellationToken));
+            var completedTask = await Task.WhenAny(runningTentacle, hasTentacleStarted.WaitAsync(), Task.Delay(TimeSpan.FromMinutes(1), cancellationToken));
 
             // Will throw.
-            if (runningTentacle.IsCompleted) await runningTentacle;
+            if (completedTask == runningTentacle)
```
nathanwoctopusdeploy, line 177: I think checking Is completed may be inaccurate sometimes due to a race condition. I fixed issues in tentacle client caused by using IsCompleted which reported false while it was the task that caused When any to exit!!

LukeButters (approved):
I am not sure that it is the binary not exiting that is the problem, but if it helps LGTM

PR #475, opened 2023-08-25
Making WebSocketStream support async disposal

sburmanoctopus:
[sc-57814]

# Background

While making Halibut async, we found that a `WebSocketStream` does not support asynchronous disposal.

# Results

## Before

`WebSocketStream` disposed synchronously, resulting in a call to the asynchronous `context.CloseOutputAsync` with a blocking `.GetAwaiter().GetResult()` at the end...

## After

`WebSocketStream` now supports async disposal, and therefore calls `context.CloseOutputAsync` asynchronously.

As a result, when `SecureConnection` wraps a `WebSocketStream`, it can dispose of it asynchronously (which it already did, but would result in sync disposal before).

Using `SecureConnection`, the `SecureWebSocketClient` can now dispose of of the connection asynchronously (on async paths).

Also, `SecureWebSocketListener` can now dispose of the `WebSocketStream` asynchronously also.

# How to review this PR

`SecureWebSocketListener` doesn't have a "sync vs async" path. So making it dispose `WebSocketStream` async is a different behaviour to the synchronous version.

I think it will be fine, but thought it worth raising in case we wish to split that out into a separate story.

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/AsyncClientAndServiceMustNotUseSyncNetworkIO.cs
+++ b/source/Halibut.Tests/AsyncClientAndServiceMustNotUseSyncNetworkIO.cs
@@ -96,22 +96,7 @@ namespace Halibut.Tests
                 // TODO: ASYNC ME UP!
                 // All we see in this stack trace is the single call to this method, we see no parent.
                 .Where(s => !s.Contains("SyncIoRecordingWebSocketStream.Flush()"))
-
-                // TODO: ASYNC ME UP!
-                // We are missing an async dispose on web sockets, and ExecuteRequest in SecureWebSocketListener needs to call the async close.
-                .Where(s => !(s.Contains("SynIoRecording.SyncIoRecordingWebSocketStream.Dispose")
-                              && s.Contains("ExecuteRequest")))
                 
-                // TODO: ASYNC ME UP!
-                // SecureConnection should dispose websockets async
-                .Where(s => !(s.Contains("SynIoRecording.SyncIoRecordingWebSocketStream.Dispose")
-                              && s.Contains("Halibut.Transport.SecureConnection.DisposeAsync()")))
-
-                // TODO: ASYNC ME UP!
-                // We seem to be using a sync dispose on DisposableNotifierConnection which results in a sync write or close
-                .Where(s => !(s.Contains("Halibut.Transport.ConnectionManagerAsync.DisposableNotifierConnection.Dispose()")
-                              && s.Contains("SynIoRecording.SyncIoRecordingWebSocketStream")))
-
                 // The follow can not be fixed up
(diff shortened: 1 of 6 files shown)


source/Halibut.Tests/AsyncClientAndServiceMustNotUseSyncNetworkIO.cs
```diff
@@ -96,22 +96,7 @@ internal static string[] FilterOutPlacesWeKnowAreStillSync(string[] distinctPlac
                 // TODO: ASYNC ME UP!
                 // All we see in this stack trace is the single call to this method, we see no parent.
                 .Where(s => !s.Contains("SyncIoRecordingWebSocketStream.Flush()"))
```
sburmanoctopus, line 98: Flush still gets called heaps. So this had to stay in 😞

source/Halibut.Tests/AsyncClientAndServiceMustNotUseSyncNetworkIO.cs
```diff
@@ -96,22 +96,7 @@ internal static string[] FilterOutPlacesWeKnowAreStillSync(string[] distinctPlac
...
-
-                // TODO: ASYNC ME UP!
```
LukeButters, line 100: nice

source/Halibut/Transport/Protocol/WebSocketStream.cs
```diff
@@ -166,14 +166,15 @@ public override long Position
             set => throw new NotImplementedException();
         }
 
-        void SendCloseMessage()
+        async Task SendCloseMessage()
```
LukeButters, line 169: Could you copy past this method to keep the sync behaving just like it used to.

source/Halibut/Transport/Protocol/WebSocketStream.cs
```diff
@@ -189,7 +190,7 @@ protected override void Dispose(bool disposing)
             {
                 try
                 {
-                    SendCloseMessage();
+                    SendCloseMessage().GetAwaiter().GetResult();
```
LukeButters, line 193: And change this back to calling the existing previous SendCloseMEssage

source/Halibut/Transport/Protocol/WebSocketStream.cs
sburmanoctopus: I figured that it was logically the same. But I can put it back to be safe 🙂

source/Halibut/Transport/Protocol/WebSocketStream.cs
sburmanoctopus: Done

LukeButters (approved):
lgtm

PR #476, opened 2023-08-25
Create TcpClient.CloseImmediately extension method

acodrington:
# Background
This PR introduces a `CloseImmediately` extension method for `TcpClient`, which ensures that no blocking I/O is performed upon trying to close the client.

After some investigation, we were able to clearly show via WireShark that while `TcpClient.Close()` returns almost immediately, the OS has not actually closed the socket and is still trying to send data. This can cause stability issues on the build server when running tests, as sockets are being held open at the OS level longer than expected.

The new `CloseImmediately` extension method first closes the internal Socket manually, explicitly passing a `0` timeout, _then_ calls the regular `TcpClient.Close()` method. This ensures that the client _immediately_ kills the connection, regardless of pending operations, freeing up the socket as expected. This behaviour was verified via WireShark, as well as by a new automated test which has been added.

A new integration fixture `TrustFixture` has also been added, which exercises the only non-obsolete code path which will make use of this change.

# Results
Fixes [sc-57237]

## Before
```
TcpClient client = ...

// Setup code

Logger.Information("Writer has blocked, closing...");
var stopWatch = Stopwatch.StartNew();
client.Close();
stopWatch.Stop();
Logger.Information($"Close completed, duration: {stopWatch.Elapsed.TotalMilliseconds}ms");

```
20230825-152618_rider64_0BgbNgwFCJ

20230825-151934_Wireshark_HH3NRZBxP6

## After

```
TcpClient client = ...

// Setup code

Logger.Information("Writer has blocked, closing...");
var stopWatch = Stopwatch.StartNew();
client.CloseImmediately();
stopWatch.Stop();
Logger.Information($"Close completed, duration: {stopWatch.Elapsed.TotalMilliseconds}ms");

```
20230825-163036_rider64_yOp1paJHsW

20230825-162414_Wireshark_bYOYUA2bBd

### DIFF
--- a/source/Halibut.TestProxy/TcpTunnel.cs
+++ b/source/Halibut.TestProxy/TcpTunnel.cs
@@ -50,11 +50,21 @@ namespace Halibut.TestProxy
                 await Task.WhenAll(
                     fromWriter.CompleteAsync().AsTask(),
                     toWriter.CompleteAsync().AsTask());
-                
+
                 Try.CatchingError(() => fromStream.Close(), e => { _logger.LogWarning("Error closing fromStream"); });
                 Try.CatchingError(() => toStream.Close(), e => { _logger.LogWarning("Error closing toStream"); });
-                Try.CatchingError(() => fromClient.Close(), e => { _logger.LogWarning("Error closing fromClient"); });
-                Try.CatchingError(() => toClient.Close(), e => { _logger.LogWarning("Error closing toClient");});
+                Try.CatchingError(() =>
+                {
+                    // Close the socket, with no waiting
+                    fromClient.Client.Close(0);
+                    fromClient.Close();
+                }, e => { _logger.LogWarning("Error closing fromClient"); });
+                Try.CatchingError(() =>
+                {
+                    // Close the socket, with no waiting
+                    toClient.Client.Close(0);
+                    toClient.Close();
+                }, e => { _logger.LogWarning("Error closing toClient"); });
             }
         }
 
--- a/source/Halibut.TestProxy/TcpTunnel.cs
+++ b/source/Halibut.TestProxy/TcpTunnel.cs
@@ -53,18 +53,8 @@ namespace Halibut.TestProxy
 
                 Try.CatchingError(() => fromStream.Close(), e => { _logger.LogWarning("Error closing fromStream"); });
                 Try.CatchingError(() => toStream.Close(), e => { _logger.LogWarning("Error closing toStream"); });
-                Try.CatchingError(() =>
-                {
-                    // Close the socket, with no waiting
-                    fromClient.Client.Close(0);
-                    fromClient.Close();
-                }, e => { _logger.LogWarning("Error closing fromClient"); });
-                Try.CatchingError(() =>
-                {
-                    // Close the socket, with no waiting
-                    toClient.Client.Close(0);
-                    toClient.Close();
-                }, e => { _logger.LogWarning("Error closing toClient"); });
+                fromClient.CloseImmediately(_ => { _logger.LogWarning("Error closing fromClient"); });
+                toClient.CloseImmediately(_ => { _logger.LogWarning("Error closing toClient"); });
             }
         }
 
--- a/source/Halibut.Tests/TcpClientCloseImmediatelyFixture.cs
+++ b/source/Halibut.Tests/TcpClientCloseImmediatelyFixture.cs
@@ -3,10 +3,10 @@ using System.Diagnostics;
 using System.IO;
 using System.Net;
 using System.Net.Sockets;
-using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using FluentAssertions;
+using Halibut.Tests.Support;
 using Halibut.Tests.Transport.Streams;
 using Halibut.Tests.Util;
 using Halibut.Transport;
@@ -17,26 +17,33 @@ namespace Halibut.Tests
     public class TcpClientCloseImmediatelyFixture : BaseTest
     {
         [Test]
-        public async Task DoesNotWait()
+        public async Task DoesNotWaitNew()
         {
-            var client = await GetTcpClientInWritingBlockedState();
+            var (client, clientStream, serverStream) = await BuildTcpClientAndTcpListener(CancellationToken);
+
+            await WriteToStreamUntilBlocked(clientStream);
 
             Logger.Information("Writer has blocked, closing...");
             var stopWatch = Stopwatch.StartNew();
             client.CloseImmediately();
             stopWatch.Stop();
             Logger.Information($"Close completed, duration: {stopWatch.Elapsed.TotalMilliseconds}ms");
-
+            
             stopWatch.ElapsedMilliseconds.Should().BeLessThan(10);
+
+            byte[] received = new byte[65536];
+            await AssertAsync.Throws<IOException>(async () => _ = await serverStream.ReadAsync(received, 0, received.Length, CancellationToken));
         }
 
         [Test]
         public async Task CanBeCalledMultipleTimes()
         {
-            var client = await GetTcpClientInWritingBlockedState();
+            var (client, clientStream, _) = await BuildTcpClientAndTcpListener(CancellationToken);
+
+            await WriteToStreamUntilBlocked(clientStream);
 
             Logger.Information("Writer has blocked, closing...");
-            client.CloseImmediately();
+            client.Close();
             Logger.Information($"Close completed");
 
             // Wait a little, to ensure any behind-the-scenes clean up has completed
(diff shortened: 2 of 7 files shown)


source/Halibut.TestProxy/TcpTunnel.cs
```diff
@@ -50,11 +50,21 @@ public async Task Tunnel(CancellationToken cancellationToken)
...
                     toWriter.CompleteAsync().AsTask());
-                
+
                 Try.CatchingError(() => fromStream.Close(), e => { _logger.LogWarning("Error closing fromStream"); });
                 Try.CatchingError(() => toStream.Close(), e => { _logger.LogWarning("Error closing toStream"); });
-                Try.CatchingError(() => fromClient.Close(), e => { _logger.LogWarning("Error closing fromClient"); });
-                Try.CatchingError(() => toClient.Close(), e => { _logger.LogWarning("Error closing toClient");});
+                Try.CatchingError(() =>
+                {
+                    // Close the socket, with no waiting
```
nathanwoctopusdeploy, line 58: nit: Copy the Entensions into the Halibut.testProxy project

source/Halibut.Tests/TcpClientCloseImmediatelyFixture.cs
```diff
@@ -0,0 +1,189 @@
...
+                            Logger.Information("WRITER: Start");
+                            await sut.WriteToStream(StreamWriteMethod.WriteAsync, data, 0, data.Length, writeTokenSource.Token);
+                            Logger.Information("WRITER: Finished");
+                        },
+                        CancellationToken);
+                });
+
+                Logger.Information("Wait for writer to block...");
+                // Wait for some brute force writing...
+                while (semaphore.Wait(1000))
```
nathanwoctopusdeploy, line 58: I'm not sure I understand the intent of the waiting logic on `semephore`. 

The intent is to only perform the `lient.CloseImmediately();` once the `await sut.WriteToStream(StreamWriteMethod.WriteAsync, data, 0, data.Length, writeTokenSource.Token);` is blocked and has not returned.

Does this achieve it?

source/Halibut.Tests/TcpClientCloseImmediatelyFixture.cs
```diff
@@ -0,0 +1,189 @@
...
+            {
+                // Ensure the timeouts are not used
+                sut.WriteTimeout = (int)TimeSpan.FromSeconds(240).TotalMilliseconds;
+                sut.ReadTimeout = (int)TimeSpan.FromSeconds(240).TotalMilliseconds;
+
+                var data = new byte[655360];
+                var r = new Random();
+                r.NextBytes(data);
+
+                var semaphore = new SemaphoreSlim(1, 1);
```
nathanwoctopusdeploy, line 95: nit: copy paste of all the logic to get the Socket into the correct state. Can this be refactored so it's shared across the 2 tests.

source/Halibut/Transport/TcpClientManager.cs
```diff
@@ -35,7 +35,7 @@ public void Disconnect(string thumbprint)
                 {
                     foreach (var client in tcpClients)
                     {
-                        client.Close();
+                        client.CloseImmediately();
```
nathanwoctopusdeploy, line 38: We could do with some tests that actually ensure this is working as expected when used on the HalibutRuntime. Tests on the extension method are great for all the variation testing but we want to know it really works for Halibut in some way as well.

source/Halibut.Tests/TcpClientCloseImmediatelyFixture.cs
acodrington: It does, but upon review there is still some non-determinism at play here and it's not an intuitive implementation so I'm going to re-implement.

source/Halibut.TestProxy/TcpTunnel.cs
```diff
@@ -50,11 +50,21 @@ public async Task Tunnel(CancellationToken cancellationToken)
...
+                    fromClient.Client.Close(0);
+                    fromClient.Close();
+                }, e => { _logger.LogWarning("Error closing fromClient"); });
+                Try.CatchingError(() =>
+                {
+                    // Close the socket, with no waiting
+                    toClient.Client.Close(0);
```
LukeButters, line 65: can `toClient.Client.Close(0)` throw? if so is it an issue if `toClient.Close();` is not called?

source/Halibut/Transport/TcpClientManager.cs
LukeButters: Would that be to try and test close is an immediate close, or is that to test that the connection is observed as closed?

source/Halibut.Tests/TcpClientCloseImmediatelyFixture.cs
```diff
@@ -0,0 +1,144 @@
...
+using Halibut.Tests.Transport.Streams;
+using Halibut.Tests.Util;
+using Halibut.Transport;
+using NUnit.Framework;
+
+namespace Halibut.Tests
+{
+    public class TcpClientCloseImmediatelyFixture : BaseTest
+    {
+        [Test]
```
LukeButters, line 19: @acodrington and I talked about this.

This test is hard to write since `close()` and `CloseImmediately` appears to be fast even if the send buffer is full (and the last write timed out because the receiver is not reading any bytes).

Looking at ways to test this I notice some differences between `close()` and `CloseImmediatately`, when the read never reads and writes are done until a timeout is thrown:
| After the write times out and the one of the close methods its called: | Close()                | CloseImmediately()                                   | Reason                                                                                                                                                                |
|------------------------------------------------------------------------|------------------------|------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| When reading those writes the number of read bytes is                  | ~64k                   | ~32K                                                 | This probably happens because it reads both the bytes in some recieve buffer and in the case of just close() it also reads the bytes in the send buffer of the writer |
| When reading                                                           | No exception is thrown | An exception is thrown outlining a RST has occurred  |                                                                                                                                                                       |
|                                                                        |                        |                                                      |                                                                                                                                                                       |

acodrington: During our conversation, Luke and I figured out that the test would still pass even if we rolled back the change, so it is not providing any value as an automated test. This happens because in the old code, `TcpClient.Close()` would still return quickly, implying that the client and its socket were closed, but at the OS level, the send buffer was still keeping the socket open trying to send.

I tried reworking this test, but wasn't able to get a workable solution within a reasonable timebox.

I was planning to make the test keep its own copy of all the bytes that were sent by the client _and_ received by the server. Then it's a matter of running the test as before, but allowing the server to finish reading once the connection is closed i.e. finish reading whatever was already sitting in the received buffer. At this point we could compare the bytes sent vs bytes received, and if the connection was killed ASAP, the received bytes should be missing some that were sent, as they would've been dropped from the send buffer as part of the `CloseImmediately` call.

However, I was having to change the test significantly to allow the reading to continue until there was no more data to read from the buffer, and I hit my timebox.

@nathanwoctopusdeploy You mentioned that you were happy to rely on the manual verification for this functionality, so would you prefer I remove these tests? We could also keep them around with an `[ignore]` attribute, as they are still useful for manual verification with WireShark.

source/Halibut.Tests/TcpClientCloseImmediatelyFixture.cs
```diff
@@ -0,0 +1,116 @@
...
+        async Task WriteToStreamUntilBlocked(Stream stream)
+        {
+            var data = new byte[655360];
+            var r = new Random();
+            r.NextBytes(data);
+
+            while (true)
+            {
+                var timeoutTask = Task.Delay(1000, CancellationToken);
+                var writingTask = stream.WriteToStream(StreamWriteMethod.WriteAsync, data, 0, data.Length, CancellationToken);
```
LukeButters, line 103: nit: Can you just call `WriteAsync`

source/Halibut.Tests/TcpClientCloseImmediatelyFixture.cs
```diff
@@ -0,0 +1,116 @@
...
+            client.CloseImmediately();
+            Logger.Information("Close completed");
+        }
+
+        async Task<(TcpClient Client, Stream ClientStream, Stream ServerStream)> BuildTcpClientAndTcpListener(CancellationToken cancellationToken)
+        {
+            var server = new TcpListener(IPAddress.Loopback, 0);
+            server.Start();
+
+            using var semaphore = new SemaphoreSlim(0, 1);
```
LukeButters, line 63: nit: rename this to something a little more meaningful.

source/Halibut.Tests/TcpClientCloseImmediatelyFixture.cs
```diff
@@ -0,0 +1,116 @@
...
+using Halibut.Tests.Util;
+using Halibut.Transport;
+using NUnit.Framework;
+
+namespace Halibut.Tests
+{
+    public class TcpClientCloseImmediatelyFixture : BaseTest
+    {
+        [Test]
+        public async Task DoesNotWaitNew()
```
LukeButters, line 20: It might be worth while giving this test a better name.

source/Halibut.Tests/TcpClientCloseImmediatelyFixture.cs
```diff
@@ -0,0 +1,116 @@
...
+            Logger.Information("Writer has blocked, closing...");
+            var stopWatch = Stopwatch.StartNew();
+            client.CloseImmediately();
+            stopWatch.Stop();
+            Logger.Information($"Close completed, duration: {stopWatch.Elapsed.TotalMilliseconds}ms");
+            
+            stopWatch.ElapsedMilliseconds.Should().BeLessThan(10);
+
+            byte[] received = new byte[65536];
+            await AssertAsync.Throws<IOException>(async () => _ = await serverStream.ReadAsync(received, 0, received.Length, CancellationToken));
```
LukeButters, line 35: It would be worth documenting why the IOException is meaningful to `CloseImmediatly`
Maybe add:
```
// Close immediately will result in the buffers on the senders side being dropped and a TCP RST being sent,
// Its hard to see the data in the senders bufferes being dropped however we can observe a TCP RST is recieved
// by the reciever. So we assert on that.
// An ordinary close will not result in the reciever recieving a TCP RST.
```

It would make sense to assert the exception is one that talks about TCP RST.

Don't forget that the `ReadAsync` doesn't need to fill the buffer given it so `ReadAsync` ought to be in a loop until it returns a value of zero.

source/Halibut.Tests/TrustFixture.cs
```diff
@@ -0,0 +1,34 @@
...
+using Halibut.Tests.TestServices.Async;
+using Halibut.TestUtils.Contracts;
+using NUnit.Framework;
+
+namespace Halibut.Tests
+{
+    public class TrustFixture : BaseTest
+    {
+        [Test]
+        [LatestClientAndLatestServiceTestCases(testListening: true, testPolling: false, testWebSocket: false)]
```
LukeButters, line 16: Should polling and websockets be tested as well?

source/Halibut.Tests/TrustFixture.cs
```diff
@@ -0,0 +1,34 @@
...
+        public async Task TrustOnlyClosesConnectionsToUntrustedClients(ClientAndServiceTestCase clientAndServiceTestCase)
```
LukeButters, line 17: This test doesn't test that connections are closed after a call to trustonly it tests that new requests can't be made to un-trusted services. That is a valuable test.

Is it worth adding a test which checks the connection is clossed? Perhaps with something like `WithCountTcpConnectionsCreated` or something similar (maybe counting closed connections)

LukeButters: You can override the stream in the stream factory (in async) so you could also see the connection being closed via that. if that works better and you did want that test.

source/Halibut.Tests/TrustFixture.cs
acodrington: The polling and websocket tests didn't exercise the code as they didn't start listeners. The `TrustOnly` call only clears listeners.

Would this effectively be testing polling/websockets anyway because Halibut is always listening on one side of the connection? In listening mode, Tentacle is the listener, and in polling/websocket, Octopus Server is the listener. Either way, the Halibut listening code is being exercised.

source/Halibut.Tests/TrustFixture.cs
acodrington: I think this test covers the relevant functionality - we don't care specifically if the connection is closed, we just care that what was once trusted is no longer trusted (the connection itself is an implementation detail)

LukeButters (approved):

PR #481, opened 2023-08-29
Optimise the usage of CancellationTokens

nathanwoctopusdeploy:
# Background

This PR has a few changes around the way cancellation tokens are used to reduce some undesirable memory growth and object retention observed in Async Halibut when running at scale.

The combination of these fixes seems to have reduced the observed memory growth, with the GC seen to be working more efficiently and has removed some suspect retention of object around Sockets.

Please see the comments for details of the individual changes.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Halibut/ServiceModel/RequestCancellationTokens.cs
```diff
@@ -11,28 +11,37 @@ public RequestCancellationTokens(CancellationToken connectingCancellationToken,
...
             ConnectingCancellationToken = connectingCancellationToken;
             InProgressRequestCancellationToken = inProgressRequestCancellationToken;
-        }
-
-        public CancellationToken ConnectingCancellationToken { get; set; }
-        public CancellationToken InProgressRequestCancellationToken { get; set; }
 
-        public CancellationToken LinkedCancellationToken => LazyLinkedCancellationToken.Value;
-
-        Lazy<CancellationToken> LazyLinkedCancellationToken => new (() =>
```
nathanwoctopusdeploy, line 21: Removed the Lazy as the LinkedCancellationToken is always referenced so the Lazy is always created.

nathanwoctopusdeploy: Reason: Simplification to remove any possibility of retention due to the referenced objects of the lazy

source/Halibut/ServiceModel/RequestCancellationTokens.cs
```diff
@@ -11,28 +11,37 @@ public RequestCancellationTokens(CancellationToken connectingCancellationToken,
...
-        {
             if (ConnectingCancellationToken == CancellationToken.None && InProgressRequestCancellationToken == CancellationToken.None)
             {
-                return CancellationToken.None;
+                LinkedCancellationToken = CancellationToken.None;
+            }
+            else if (InProgressRequestCancellationToken == CancellationToken.None)
```
nathanwoctopusdeploy, line 19: Optimisations to not wrap a single cancellation token in a new cancellation token source

source/Halibut/Transport/SecureListener.cs
```diff
@@ -38,7 +38,8 @@ public class SecureListener : IDisposable
         readonly ILogFactory logFactory;
         readonly Func<string> getFriendlyHtmlPageContent;
         readonly Func<Dictionary<string, string>> getFriendlyHtmlPageHeaders;
-        readonly CancellationTokenSource cts = new();
+        readonly CancellationTokenSource cts;
+        readonly CancellationToken cancellationToken;
```
nathanwoctopusdeploy, line 42: Usage of a single cancellation token to avoid creating new cancellation tokens each time they are needed from the source (it's a simple pass-through object but we create a lot of them)

source/Halibut/Transport/SecureListener.cs
```diff
@@ -246,18 +250,28 @@ async Task ExecuteRequest(TcpClient client)
...
+                                           ((IDisposable)weakSsl.Target).Dispose();
+                                   }))
+                            {
+                                tcpClientManager.AddActiveClient(thumbprint, client);
+                                await ExchangeMessages(ssl).ConfigureAwait(false);
+                            }
+                        }
+                        else
                         {
+                            // The stream is wrapped in a NetworkTimeoutStream which handles closing the inner stream on timeout
```
nathanwoctopusdeploy, line 273: For Async Halibut we don't need this weak ref and disposal as we wrap all Async Halibut calls so they cancel / timeout / close / dispose

source/Halibut/Util/CancellationTokenExtensionMethods.cs
```diff
@@ -1,40 +0,0 @@
-﻿#nullable enable
-using System;
-using System.Threading;
-using System.Threading.Tasks;
-
-namespace Halibut.Util
-{
-    internal static class CancellationTokenExtensionMethods
```
nathanwoctopusdeploy, line 8: No longer needed

source/Halibut/ServiceModel/RequestCancellationTokens.cs
```diff
@@ -11,28 +11,37 @@ public RequestCancellationTokens(CancellationToken connectingCancellationToken,
...
+            {
+                LinkedCancellationToken = ConnectingCancellationToken;
             }
+            else if (ConnectingCancellationToken == CancellationToken.None)
+            {
+                LinkedCancellationToken = InProgressRequestCancellationToken;
+            }
+            else
```
LukeButters, line 27: Does another optimisation exists where if the `ConnectingCancellationToken` and `InProgressRequestCancellationToken` are reference equal then use either as the linked token

source/Halibut/Transport/SecureListener.cs
LukeButters: Oh so the CTS creates a new token each time it is asked for a token?

nathanwoctopusdeploy: Yes, it's a very lightweight wrapper that calls into the CancellationTokenSource, but it's another object with another reference.

source/Halibut/ServiceModel/RequestCancellationTokens.cs
nathanwoctopusdeploy: Yes good call. It's not a code path that would be triggered in our current usage but good to optimise for future usage 👍

source/Halibut/Transport/SecureListener.cs
nathanwoctopusdeploy: @LukeButters `TcpClientManager` is now Disposable and the `SecureListener` will Dispose of it

LukeButters: Nice thanks! I think the result is simpler to follow code.

LukeButters (approved):

source/Halibut/ServiceModel/RequestCancellationTokens.cs
```diff
@@ -11,28 +11,41 @@ public RequestCancellationTokens(CancellationToken connectingCancellationToken,
...
+            {
+                linkedCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(ConnectingCancellationToken, InProgressRequestCancellationToken);
 
-            linkedCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(ConnectingCancellationToken, InProgressRequestCancellationToken);
+                LinkedCancellationToken = linkedCancellationTokenSource.Token;
+            }
+        }
+
+        public CancellationToken ConnectingCancellationToken { get; set; }
```
sburmanoctopus, line 39: A bit late to the party, but why are these two property setters public? Why not just read only properties?

PR #484, opened 2023-08-30
Fix potential null reference when disposing of the TCP Manager

nathanwoctopusdeploy:
# Background

image

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Transport/TcpClientManager.cs
+++ b/source/Halibut/Transport/TcpClientManager.cs
@@ -77,15 +77,23 @@ namespace Halibut.Transport
 
         public void Dispose()
         {
-            var clients = activeClients.ToArray();
-            activeClients.Clear();
+            var clients = activeClients?.ToArray();
+            activeClients?.Clear();
 
+            if (clients == null || !clients.Any())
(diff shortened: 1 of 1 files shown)


sburmanoctopus (approved):
LGTM. I saw that error, and was about to put a card on the board 😁 Nice one.

nathanwoctopusdeploy:
> LGTM. I saw that error, and was about to put a card on the board 😁 Nice one.

I made the issue so thought I should fix it 😆

source/Halibut/Transport/TcpClientManager.cs
```diff
@@ -77,15 +77,23 @@ public void RemoveClient(TcpClient client)
 
         public void Dispose()
         {
-            var clients = activeClients.ToArray();
-            activeClients.Clear();
+            var clients = activeClients?.ToArray();
```
sburmanoctopus, line 80: I think this is redundant, as `activeClients` is `readonly`, and is set at the top.

But I'm ok with being safe rather than sorry 🙂

PR #485, opened 2023-08-31
Adds support for sending logs to multiple destinations e.g. to both a in memory log as well as to something else which can be selective about what it logs.

LukeButters:
# Background

[SC-55402]

Makes it possible to add a customer `ILogFactory` which logs specific events or records metrics while also still logging to the in memory logger.

Here is a very spelled out example of how to use the new classes:
```
// A thing that creates a new `InMemoryConnectionLog` on each call.
// Create a single one for performance to cache the LogProvider.GetLogger("Halibut")
ICreateNewILog singleInMemoryConntionLogCreator = new InMemoryConnectionLogCreator();


// These log writers only record and can never return logs.
ILogWriter logConnectionErrors = new LogConnectionErrors();
ILogWriter logSecurityEvents = new LogSecurityEvents();

Func<string, ILogWriter[]> logWriterCreator = prefix => new[] {logConnectionErrors, logSecurityEvents};


// Aggregate the logWriters and the in memory connection log creator
ICreateNewILog aggregateLoggerCreator = new AggregateLogWriterLogCreator(singleInMemoryConntionLogCreator, logWriterCreator);

// Now we need it to be a log factory, and for the ILogs to be cached.
ILogFactory logFactory = aggregateLoggerCreator.ToCachingLogFactory();
```

Changes: 
* Adds a new `ICreateNewILog` which makes new `ILog`s.
* Adds a `InMemoryConnectionLogCreator` which just creates new `InMemoryConnectionLog`
* Adds a `ILogWriter` which is just like a `ILog` except that it is write only, you can't read back logs.
* Adds a `CachingLogFactory` which copies the caching logic from `LogFactory`. Doing so means it is easy to add new ways of logging since they don't need to concern themselves with caching.
* Removes `TestContextLogFactory.cs` and replaces it with `TestContextLogCreator` which is much simpler thanks to splitting out the caching.
* Removes the `AggregateILog` from tests since we now have this capability from Halibut.
* Adds a `AggregateLogWriterLog` which sends logs to both a `ILog` as well as a set of `ILogWriter`.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Halibut/Diagnostics/CachingLogFactory.cs
```diff
@@ -0,0 +1,28 @@
+using System;
+using System.Collections.Concurrent;
+using Halibut.Diagnostics.LogCreators;
+
+namespace Halibut.Diagnostics
+{
+    public class CachingLogFactory : ILogFactory
```
LukeButters, line 7: This is copied from `LogFactory` which did two things cache and log to a in memory log.

source/Halibut/Diagnostics/LogCreators/AggregateLogWriterLogCreator.cs
```diff
@@ -0,0 +1,22 @@
...
+using Halibut.Diagnostics.LogWriters;
+
+namespace Halibut.Diagnostics.LogCreators
+{
+    public class AggregateLogWriterLogCreator : ICreateNewILog
+    {
+        readonly ICreateNewILog nonCachingLogFactory;
+        readonly Func<string, ILogWriter[]> logWriterFactoryForPrefix;
+
+        public AggregateLogWriterLogCreator(ICreateNewILog nonCachingLogFactory, Func<string, ILogWriter[]> logWriterFactoryForPrefix)
```
sburmanoctopus, line 11: Nit: I know the func makes it more reusable, but looking at the example in the Background, gosh it'd be nice to have it with `params`.

Maybe we could add a second constructor like this `public AggregateLogWriterLogCreator(ICreateNewILog nonCachingLogFactory, params ILogWriter[] logWriters)`, and all it does is call the other one with a special func that returns just those objects.

I think that Octopus Server would only use the same instances anyway, right?

Then the example in Background would look like this:

```
These log writers only record and can never return logs.
ILogWriter logConnectionErrors = new LogConnectionErrors();
ILogWriter logSecurityEvents = new LogSecurityEvents();

// Aggregate the logWriters and the in memory connection log creator
ICreateNewILog aggregateLoggerCreator = new AggregateLogWriterLogCreator(singleInMemoryConntionLogCreator, logConnectionErrors, logSecurityEvents);
```

source/Halibut/Diagnostics/LogCreators/AggregateLogWriterLogCreator.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+using Halibut.Diagnostics.LogWriters;
+
+namespace Halibut.Diagnostics.LogCreators
+{
+    public class AggregateLogWriterLogCreator : ICreateNewILog
+    {
+        readonly ICreateNewILog nonCachingLogFactory;
```
sburmanoctopus, line 8: Nit: Do we care that this is 'non-caching'?

Could we just call it `createNewILog`? (or whatever name we came up with before in the previous comment)

source/Halibut/Diagnostics/LogCreators/InMemoryConnectionLogCreator.cs
```diff
@@ -0,0 +1,20 @@
+using System;
+using Halibut.Logging;
+
+namespace Halibut.Diagnostics.LogCreators
+{
+    /// <summary>
```
sburmanoctopus, line 6: Nit: I feel this whole summary is unrequired, and a little confusing when it talks about caching.

Personally, I'd delete it, as I think `ICreateNewILog` implies it's going to create a new one with each call. And the caching thing is a little presumptuous on how we're going to use it.

source/Halibut/Diagnostics/LogEndpointNormalizer.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+
+namespace Halibut.Diagnostics
+{
+    public static class LogEndpointNormalizer
```
sburmanoctopus, line 5: Nit: Could make this a touch sexier and make it an extention method. E.g. `public static Uri Normalize(this Uri endpoint)`

source/Halibut.Tests/Support/Logging/InMemoryLog.cs
```diff
@@ -2,11 +2,11 @@
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using Halibut.Diagnostics;
-using Serilog;
+using Halibut.Diagnostics.LogWriters;
 
 namespace Halibut.Tests.Support.Logging
 {
-    public class InMemoryLog : ILog
+    public class InMemoryLog : ILog, ILogWriter
```
sburmanoctopus, line 9: Nit: I think making a separate `InMemoryLogWriter` would make the tests a little clearer to understand. Otherwise, people might look at the tests you are using this class as an `ILogWriter` for, and wonder "why we are using an `ILog`?"

sburmanoctopus: Maybe the same goes for `TestContextConnectionLog` as well?

sburmanoctopus (approved):
LGTM

source/Halibut/Diagnostics/LogCreators/InMemoryConnectionLogCreator.cs
LukeButters: Improved the message here.

source/Halibut/Diagnostics/LogCreators/AggregateLogWriterLogCreator.cs
LukeButters: done

source/Halibut/Diagnostics/CachingLogFactory.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        readonly ConcurrentDictionary<string, ILog> cache = new();
+        readonly ICreateNewILog logFactory;
```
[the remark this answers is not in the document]
LukeButters: done

source/Halibut.Tests/Support/Logging/InMemoryLog.cs
LukeButters: done for InMemoryLog only

source/Halibut/Diagnostics/LogWriters/AggregateLogWriterLog.cs
```diff
@@ -0,0 +1,43 @@
+using System;
+using System.Collections.Generic;
+
+namespace Halibut.Diagnostics.LogWriters
+{
+    /// <summary>
+    /// An ILog which writes the log events to both an inner ILog and to the set of ILogWriters.   
+    /// </summary>
+    public class AggregateLogWriterLog : ILog
```
nathanwoctopusdeploy, line 9: I'm finding the names of things a little confusing. LogWriterLog for example!!

source/Halibut/Diagnostics/LogCreators/ICreateNewILog.cs
```diff
@@ -0,0 +1,9 @@
+using System;
+
+namespace Halibut.Diagnostics.LogCreators
+{
+    public interface ICreateNewILog
```
nathanwoctopusdeploy, line 5: Do we really need this new Creator concept? It seems like were making this area very complicated

source/Halibut/Diagnostics/CachingLogFactory.cs
nathanwoctopusdeploy, line 7: It's not really `Caching` though is it in the way you would think of caching

source/Halibut/Diagnostics/LogCreators/ICreateNewILog.cs
nathanwoctopusdeploy: Is just rebranding a builder as a different name?

LukeButters: I didn't initially have the CreateNewLog concept, but @sburmanoctopus convinced me we really have two things a `ILogFactory` which returns logs which could be retained (cached) and another thing that always makes new things. Without this concept, the result was types that wanted to express the idea that they need to be cached/retained.

Going down this path of separating things that always make new ones from things that get you a ILog seemed to force users down the right path.

source/Halibut/Diagnostics/CachingLogFactory.cs
LukeButters: hmm do you have better name in mind?

`RetainingLogFactory`
`MemoryLeakingLogFactory`
`ThisIsWhyYourMemoryDumpIsMassiveLogFactory`

source/Halibut/Diagnostics/LogWriters/AggregateLogWriterLog.cs
LukeButters: Dropped the `Log` from the end. Maybe that helps.

LukeButters: For what it is worth it is unlikely that anyone needs to actually use this class directly.

PR #486, opened 2023-08-31
Log exceptions before the queue is reached in a way that makes it easy to determine this is before message exchange.

LukeButters:
# Background

We log errors from a connection from the time theconnection is created to close to when the queue (so no more IO is being done).

These log messages now come out with events types that make it possible to determine they are during the connecting phase. These events are:
* ErrorInInitialisation,
* ErrorInIdentify
* ClientDenied

Some existing log event types are now commented to make it easy for us to identify which ones to log.

[SC-55402]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Diagnostics/EventType.cs
+++ b/source/Halibut/Diagnostics/EventType.cs
@@ -9,12 +9,33 @@ namespace Halibut.Diagnostics
         Security,
         MessageExchange,
         Diagnostic,
+        
+        /// <summary>
+        /// Used when the listening server (which may or may not be a service)
+        /// is not presented with a certificate from the client OR
+        /// the certificate presented by the client is not authorized.
+        /// 
+        /// This happens before messages are exchanged.
+        /// </summary>
         ClientDenied,
+        
         Error,
+        
+        /// <summary>
+        /// An error that has occured before message exchange
+        /// </summary>
+        ErrorInInitialisation,
+        
         ListenerStarted,
         ListenerAcceptedClient,
         ListenerStopped,
         SecurityNegotiation,
-        FileTransfer
+        FileTransfer,
+        
+        /// <summary>
+        /// Used when an error occurs identifying the remote or identify
+        /// This happens before messages are exchanged.
+        /// </summary>
+        ErrorInIdentify
     }
 }
--- a/source/Halibut/Diagnostics/EventType.cs
+++ b/source/Halibut/Diagnostics/EventType.cs
@@ -5,6 +5,7 @@ namespace Halibut.Diagnostics
     public enum EventType
     {
         OpeningNewConnection,
+        [Obsolete]
         UsingExistingConnectionFromPool,
         Security,
         MessageExchange,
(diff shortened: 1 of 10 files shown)


source/Halibut/Transport/Protocol/MessageExchangeProtocol.cs
```diff
@@ -195,10 +195,40 @@ public async Task ExchangeAsServerAsync(Func<RequestMessage, Task<ResponseMessag
...
 
+        async Task IdentifyAsServerAsync(RemoteIdentity identityOfRemote, CancellationToken cancellationToken)
+        {
+            try
+            {
+                await stream.IdentifyAsServerAsync(cancellationToken);
+            }
+            catch (Exception e)
+            {
+                log.WriteException(EventType.ErrorInIdentify, $"Failed to identify as server to the previously identified remote {identityOfRemote.SubscriptionId} of type {identityOfRemote.IdentityType}", e);
```
LukeButters, line 211: These two are hard to tests.

source/Halibut/Transport/Protocol/MessageExchangeProtocol.cs
```diff
@@ -183,8 +183,8 @@ public async Task ExchangeAsServerSynchronouslyAsync(Func<RequestMessage, Respon
 
         public async Task ExchangeAsServerAsync(Func<RequestMessage, Task<ResponseMessage>> incomingRequestProcessor, Func<RemoteIdentity, IPendingRequestQueue> pendingRequests, CancellationToken cancellationToken)
```
LukeButters, line 184: Note that I didn't add the logging to the sync paths. Should I?

LukeButters:
Oops forgot about websockets.

source/Halibut/Transport/Protocol/MessageExchangeProtocol.cs
sburmanoctopus: We are really weighing up leaving the existing code alone, vs being able to log with and without the feature toggle on.

🤔 Is this extra info worth the risk of touching existing code.

@nathanwoctopusdeploy, do you have a preference?

source/Halibut/Diagnostics/EventType.cs
```diff
@@ -5,16 +5,39 @@ namespace Halibut.Diagnostics
     public enum EventType
     {
         OpeningNewConnection,
+        [Obsolete]
```
sburmanoctopus, line 8: I flagged these as obsolete because we don't actually use them. We can clean them up when we remove all the other obsolete flags

nathanwoctopusdeploy (approved):
LGTM

source/Halibut/Transport/Protocol/MessageExchangeProtocol.cs
nathanwoctopusdeploy: We are only really interested in async errors. The only thing we would lose is the ability to see if this issue was already happening in sync

nathanwoctopusdeploy: Happy either way though

sburmanoctopus: Let's go with the safe approach then, and leave the sync path alone.

PR #488, opened 2023-09-01
Observing current RPC calls

sburmanoctopus:
[sc-58702]

# Background

We need to observer how many RPC calls are happening right now.

# Results

## After

We now have an optional observer that gets called just before and after an async RPC call is made.

This will then be used in Octopus Server to create a metric.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Support/LatestClientAndLatestServiceBuilder.cs
+++ b/source/Halibut.Tests/Support/LatestClientAndLatestServiceBuilder.cs
@@ -15,6 +15,7 @@ using Halibut.Tests.TestServices;
 using Halibut.Tests.TestServices.AsyncSyncCompat;
 using Halibut.TestUtils.Contracts;
 using Halibut.TestUtils.Contracts.Tentacle.Services;
+using Halibut.Transport.Observability;
 using Halibut.Transport.Proxy;
 using Halibut.Transport.Streams;
 using Halibut.Util;
@@ -39,8 +40,9 @@ namespace Halibut.Tests.Support
         string serviceTrustsThumbprint;
         ForceClientProxyType? forceClientProxyType;
         AsyncHalibutFeature serviceAsyncHalibutFeature = AsyncHalibutFeature.Disabled;
-        
-        
+        IRpcObserver? clientRpcObserver;
+
+
         bool hasService = true;
         Func<int, PortForwarder>? portForwarderFactory;
         Func<ILogFactory, IPendingRequestQueueFactory>? pendingRequestQueueFactory;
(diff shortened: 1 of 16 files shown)


source/Halibut.Tests/Support/LatestClientAndLatestServiceBuilder.cs
```diff
@@ -348,6 +350,12 @@ public LatestClientAndLatestServiceBuilder WithForcingClientProxyType(ForceClien
             return this;
         }
         
+        public LatestClientAndLatestServiceBuilder WithClientRpcObserver(IRpcObserver? clientRpcObserver)
```
sburmanoctopus, line 353: To be able to get the RPC observer into the HalibutRuntime builder, we had to push it in.

I did consider maybe making a more general purpose `With` method that allows someone access to the `HalibutRuntimeBuilder` of the client when it is being built.

I wasn't sure if now was the time.

source/Halibut/Transport/Protocol/MessageExchangeProtocol.cs
```diff
@@ -40,10 +43,19 @@ public ResponseMessage ExchangeAsClient(RequestMessage request)
 
         public async Task<ResponseMessage> ExchangeAsClientAsync(RequestMessage request, CancellationToken cancellationToken)
         {
-            await PrepareExchangeAsClientAsync(cancellationToken);
+            try
```
sburmanoctopus, line 46: This is where it gets called when we are giving a request to a listening tentacle

source/Halibut/Transport/Protocol/MessageExchangeProtocol.cs
```diff
@@ -342,43 +354,52 @@ async Task<bool> ProcessReceiverInternalAsync(IPendingRequestQueue pendingReques
         {
             try
             {
-                await stream.SendAsync(nextRequest, cancellationToken);
-                if (nextRequest != null)
+                rcpObserver.StartCall(nextRequest);
```
sburmanoctopus, line 357: This is where it gets called when we are giving a request to a polling tentacle

source/Halibut/Transport/Protocol/MessageExchangeProtocol.cs
```diff
@@ -40,10 +43,19 @@ public ResponseMessage ExchangeAsClient(RequestMessage request)
...
         public async Task<ResponseMessage> ExchangeAsClientAsync(RequestMessage request, CancellationToken cancellationToken)
         {
-            await PrepareExchangeAsClientAsync(cancellationToken);
+            rcpObserver.StartCall(request);
 
-            await stream.SendAsync(request, cancellationToken);
-            return await stream.ReceiveAsync<ResponseMessage>(cancellationToken);
+            try
+            {
+                await PrepareExchangeAsClientAsync(cancellationToken);
```
sburmanoctopus, line 50: I almost considered using `SendAndReceiveRequest` below here. But didn't because I thought including `PrepareExchangeAsClientAsync` as part of the 'call' made sense.

nathanwoctopusdeploy (approved):
Nice

PR #496, opened 2023-09-14
Distinguishing between a connection that was authorized or unauthorized.

sburmanoctopus:
[sc-59576]

# Background

When observing logs for async Halibut, we found that tentacles with invalid thumbprints were becoming noise in the metrics for connections.

# Results

## Before

All connections were being put into metrics in Octopus Server, as we did not know if it was an initial connection, or an authorized connection.

## After
We now pass a parameter to the observer stating whether we have successfully authorized.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Transport/SecureListener.cs
+++ b/source/Halibut/Transport/SecureListener.cs
@@ -150,7 +150,6 @@ namespace Halibut.Transport
                         }
 
                         var client = listener.AcceptTcpClient();
-                        connectionsObserver.ConnectionAccepted();
 
                         Task.Run(async () => await HandleClient(client).ConfigureAwait(false)).ConfigureAwait(false);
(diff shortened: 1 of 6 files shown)


source/Halibut/Transport/SecureListener.cs
```diff
@@ -399,9 +407,6 @@ static string GetThumbprint(SslStream stream)
 
         bool Authorize(string thumbprint, EndPoint clientName)
         {
-            if (thumbprint == null)
```
sburmanoctopus, line 402: Unnecessary guard, as we do this exact check on the line before we call `Authorize`.

nathanwoctopusdeploy (approved):

PR #497, opened 2023-09-15
Writing test for usecase where a single tentacle can poll multiple Octopus Servers

sburmanoctopus:
[sc-58911]

# Background

We need a test for when a single tentacle can poll multiple Octopus Servers.

# Results

## Before

There was no test that covered this test case.

## After

We now have a test that covers this test case.

I also removed the usage of some of the deprecated `Poll` methods.

### Future Enhancement
At some point in the future, it might be good to split the builder into a client and a server builder. Then a lot of what we did can be made into reusable components.

# How to review this PR
To minimise change, we only tested latest client and latest service. If we want to test further, let me know and I can roll the changes made to 'latest' out to the other variants as well.

Quality :heavy_check_mark:

# Pre-requisites


source/Halibut.Tests/Support/TestPollingClient.cs
```diff
@@ -0,0 +1,19 @@
+using System;
+
+namespace Halibut.Tests.Support
+{
+    public class TestPollingClient : ITestPollingClient
```
nathanwoctopusdeploy, line 5: nit: why do we need `ITestPollingClient`?

source/Halibut.Tests/Support/TestPollingClient.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        public int ListeningPort { get; }
+        public string WebSocketPath { get; }
+
+        TestPollingClient(int listeningPort, string webSocketPath)
```
nathanwoctopusdeploy, line 10: nit: Is this really PollingClientDetails or similar rather than a client that we think of in terms of client and service

source/Halibut.Tests/Support/LatestClientAndLatestServiceBuilder.cs
```diff
@@ -518,7 +558,7 @@ public async Task<ClientAndService> Build(CancellationToken cancellationToken)
                 portForwarderReference.Value = portForwarder;
             }
 
-            return new ClientAndService(client, service, serviceUri, clientTrustsThumbprint, portForwarder, disposableCollection, httpProxy, httpProxyDetails, forceClientProxyType, cancellationTokenSource);
+            return new ClientAndService(pollingClient, client, service, serviceUri, clientTrustsThumbprint, portForwarder, disposableCollection, httpProxy, httpProxyDetails, forceClientProxyType, cancellationTokenSource);
```
nathanwoctopusdeploy, line 561: This feels a bit confusing to start passing a `PollingClient` when we are setting up listing stuff

source/Halibut.Tests/Support/LatestClientAndLatestServiceBuilder.cs
```diff
@@ -601,6 +642,7 @@ public class ClientAndService : IClientAndService
                 ForceClientProxyType? forceClientProxyType,
                 CancellationTokenSource cancellationTokenSource)
             {
+                PollingClient = pollingClient;
```
nathanwoctopusdeploy, line 645: Is there a way we can pass enough generic detail to achieve the same result? It doesn't feel quite right to start passing polling specific objects into this more generic class

nathanwoctopusdeploy: do you really just need `clientUri`

source/Halibut.Tests/Support/TestPollingClient.cs
sburmanoctopus: Agreed. Deleted

source/Halibut.Tests/Support/TestPollingClient.cs
sburmanoctopus: After discussion, we decided to simplify it down to just a URI

source/Halibut.Tests/Support/LatestClientAndLatestServiceBuilder.cs
sburmanoctopus: After discussion, we decided to simplify it down to a nullable `clientUri`

source/Halibut.Tests/Support/LatestClientAndLatestServiceBuilder.cs
sburmanoctopus: Indeed. Done!

nathanwoctopusdeploy (approved):

PR #500, opened 2023-09-20
Reduce Parallel Tests

nathanwoctopusdeploy:
# Background

- Reduced the number of parallel tests in progress
- Reduced how much the min thread pool is increased, as very high increases can cause performance issues 
- Reduced iteration count for previous client due to the test infrastructure this has to set up

## After 

This is likely to increase test run time. There is a tradeoff between stability and speed

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/ParallelRequestsFixture.cs
+++ b/source/Halibut.Tests/ParallelRequestsFixture.cs
@@ -252,17 +252,16 @@ namespace Halibut.Tests
         
         static int NumberOfParallelRequests(ClientAndServiceTestCase clientAndServiceTestCase)
         {
-            int threadCount = 64;
-            if (!clientAndServiceTestCase.ClientAndServiceTestVersion.IsPreviousClient())
+            if (clientAndServiceTestCase.ClientAndServiceTestVersion.IsPreviousClient())
             {
                 // Reduce this down to reduce thread pool exhaustion.
                 // We already test latest => latest with 64 concurrent task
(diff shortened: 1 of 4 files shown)


nathanwoctopusdeploy:
[sc-59772]

source/Halibut.Tests/ParallelRequestsFixture.cs
```diff
@@ -252,17 +252,16 @@ void WaitForAllThreads(List<Thread> threads)
         
         static int NumberOfParallelRequests(ClientAndServiceTestCase clientAndServiceTestCase)
         {
-            int threadCount = 64;
-            if (!clientAndServiceTestCase.ClientAndServiceTestVersion.IsPreviousClient())
+            if (clientAndServiceTestCase.ClientAndServiceTestVersion.IsPreviousClient())
```
sburmanoctopus, line 255: Lol, didn't I just review this change in the other PR? (with the exception of using 32 instead of 64)

sburmanoctopus (approved):
LGTM

source/Halibut.Tests/ParallelRequestsFixture.cs
nathanwoctopusdeploy: Yes, this wouldn't go gree without the associated change :( 

I tried to split changes across 2 PRs but it was 100% successful!!

PR #502, opened 2023-09-22
Fix TcpClientManager Collection was modified during Disposal

nathanwoctopusdeploy:
# Background

Fix

```
System.InvalidOperationException : Collection was modified; enumeration operation may not execute.
   at System.Collections.Generic.HashSet`1.Enumerator.MoveNext()
   at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source)
   at Halibut.Transport.TcpClientManager.Dispose()
   at Halibut.Transport.SecureListener.Dispose()
   at Halibut.HalibutRuntime.Dispose()
```

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Transport/TcpClientManager.cs
+++ b/source/Halibut/Transport/TcpClientManager.cs
@@ -89,7 +89,9 @@ namespace Halibut.Transport
             {
                 if (client.Value?.Any() == true)
                 {
-                    foreach (var tcpClient in client.Value)
+                    var tcpClients = client.Value.ToArray();
+
+                    foreach (var tcpClient in tcpClients)
                     {
                         tcpClient?.CloseImmediately();
                         tcpClient?.Dispose();
--- a/source/Halibut/Transport/TcpClientManager.cs
+++ b/source/Halibut/Transport/TcpClientManager.cs
@@ -7,11 +7,12 @@ namespace Halibut.Transport
 {
     class TcpClientManager : IDisposable
     {
-        readonly Dictionary<string, HashSet<TcpClient>> activeClients = new();
+        Dictionary<string, HashSet<TcpClient>> activeClients = new();
+        readonly object syncLock = new();
 
         public void AddActiveClient(string thumbprint, TcpClient client)
         {
-            lock (activeClients)
+            lock (syncLock)
             {
                 if (activeClients.TryGetValue(thumbprint, out var tcpClients))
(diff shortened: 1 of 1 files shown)


source/Halibut/Transport/TcpClientManager.cs
```diff
@@ -89,7 +89,9 @@ public void Dispose()
             {
                 if (client.Value?.Any() == true)
                 {
-                    foreach (var tcpClient in client.Value)
+                    var tcpClients = client.Value.ToArray();
```
sburmanoctopus, line 92: 🤔 I'm not sure this will fix the crash.

Looking at the stack trace, it's failing on a `System.Linq.Enumerable.Any`. We have two of those (line 83 and line 90).

But it's also on a `System.Collections.Generic.HashSet`, so that makes it line 90.

However, this fix is after that. Also, whatever is modifying the collection can happen during a `ToArray` just as much as it could happen during `Any`.

From what I can see, this class only modifies this set within locks. 

So I wonder if the fix we are looking for is to wrap the whole dispose methods within a 'lock(activeClients)` block?

sburmanoctopus: I suspect that something is interacting with this class on another thread when we decide to dispose of it.

But that in itself is a little concerning.

If we do the lock, and `Dispose` is forced to wait until the other operation finishes, that should be fine.

As long as something else doesn't want to modify this object after we have disposed of it...

nathanwoctopusdeploy: I should learn to read the error message 😆 !!

source/Halibut/Transport/TcpClientManager.cs
```diff
@@ -58,37 +60,47 @@ public IReadOnlyCollection<TcpClient> GetActiveClients(string thumbprint)
...
-            {
-                return;
-            }
-            
-            foreach (var client in clients)
+            lock (syncLock)
             {
-                if (client.Value?.Any() == true)
+                var clients = activeClients?.ToArray();
+                activeClients = null;
```
nathanwoctopusdeploy, line 90: This will help us identify what is modifying this list during disposal / what we are not correctly stopping in Halibut etc.

sburmanoctopus (approved):
LGTM

PR #503, opened 2023-09-27
Add IP address to RPC errors

acodrington:
# Background

We would like to have more information in our logs to help us diagnose potential future issues. This PR adds the relevant IP address to a number of RPC error logs.

# Results
Fixes [sc-60356]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Transport/SecureListener.cs
+++ b/source/Halibut/Transport/SecureListener.cs
@@ -165,7 +165,7 @@ namespace Halibut.Transport
                     catch (Exception ex)
                     {
                         numberOfFailedAttemptsInRow++;
-                        log.WriteException(EventType.ErrorInInitialisation, "Error accepting TCP client: {0}", ex, client?.Client.RemoteEndPoint);
+                        log.WriteException(EventType.ErrorInInitialisation, "Error accepting TCP client: {0}", ex, client.GetRemoteEndpointString());
                         // Slow down the logs in case an exception is immediately encountered after X failed AcceptTcpClient calls
                         if (numberOfFailedAttemptsInRow >= errorThreshold)
                         {
@@ -203,7 +203,7 @@ namespace Halibut.Transport
 #pragma warning restore CS0612
                 }
 
-                log.Write(EventType.ListenerAcceptedClient, "Accepted TCP client: {0}", client.Client.RemoteEndPoint);
+                log.Write(EventType.ListenerAcceptedClient, "Accepted TCP client: {0}", client.GetRemoteEndpointString());
                 await ExecuteRequest(client).ConfigureAwait(false);
             }
             catch (ObjectDisposedException)
@@ -211,7 +211,7 @@ namespace Halibut.Transport
             }
             catch (Exception ex)
             {
-                log.WriteException(EventType.ErrorInInitialisation, "Error initializing TCP client: {0}", ex, client.Client.RemoteEndPoint);
+                log.WriteException(EventType.ErrorInInitialisation, "Error initializing TCP client: {0}", ex, client.GetRemoteEndpointString());
             }
         }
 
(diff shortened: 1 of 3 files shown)


source/Halibut/Transport/SecureListener.cs
```diff
@@ -164,7 +165,7 @@ void WaitForPendingConnectionOrCancellation()
                     catch (Exception ex)
                     {
                         numberOfFailedAttemptsInRow++;
-                        log.WriteException(EventType.ErrorInInitialisation, "Error accepting TCP client", ex);
+                        log.WriteException(EventType.ErrorInInitialisation, "Error accepting TCP client: {0}", ex, client?.Client.RemoteEndPoint);
```
nathanwoctopusdeploy, line 168: Client.RemoteEndPoint can throw exceptions so do you think it would be safer to call an extension method or similar that wraps the call in a try catch so we guarantee we don't error trying to log an error 
image

nathanwoctopusdeploy: I don't know if it would ever throw in the situation we are using it, but just incase

nathanwoctopusdeploy: Kind of a on th lines of Halibut.Transport.Proxy.Utils.GetHost and GetPort

source/Halibut/Transport/TcpClientExtensions.cs
```diff
@@ -93,5 +94,17 @@ public static void CloseImmediately(this TcpClient client, Action<Exception> onE
...
+        public static string GetRemoteEndpointString(this TcpClient client)
+        {
+            try
+            {
+                return client?.Client.RemoteEndPoint?.ToString() ?? "<none>";
+            }
+            catch
+            {
+                return "<error>";
+            }
```
acodrington, line 107: RemoteEndPoint can throw if the socket has already been disposed/closed, but that's a bit wordy for this usage i.e. log message.

I figured `<error>` vs `<none`> would help us differentiate via the logs what happened without being too wordy. These will be easy to find via a quick CTRL + F.

source/Halibut/Transport/SecureListener.cs
```diff
@@ -202,29 +203,29 @@ async Task HandleClient(TcpClient client)
...
+                log.WriteException(EventType.ErrorInInitialisation, "Error initializing TCP client: {0}", ex, client.GetRemoteEndpointString());
             }
         }
 
         async Task ExecuteRequest(TcpClient client)
         {
             var connectionAuthorizedAndObserved = false;
 
-            var clientName = client.Client.RemoteEndPoint;
+            string clientName = client.GetRemoteEndpointString();
```
nathanwoctopusdeploy, line 222: var -> string?

acodrington: Unintentional, will revert

source/Halibut/Transport/SecureWebSocketListener.cs
```diff
@@ -114,7 +115,7 @@ async Task Accept()
                     {
                         if (!cts.IsCancellationRequested)
                         {
-                            log.WriteException(EventType.ErrorInInitialisation, "Error accepting Web Socket client", ex);
+                            log.WriteException(EventType.ErrorInInitialisation, "Error accepting Web Socket client: {0}", ex, context?.Request.RemoteEndPoint);
```
nathanwoctopusdeploy, line 118: Should these use the extension method or is it a different APi that doesn't throw

nathanwoctopusdeploy (approved):

source/Halibut/Transport/SecureWebSocketListener.cs
acodrington: The latter - it doesn't throw but I just checked and it _can_ be null. I'm not sure what the log method will do with a null object, so I'll check that behaviour and update the code if necessary.

acodrington:
[sc-60356]

PR #504, opened 2023-10-02
Add xUnit

acodrington:
# Background

This PR adds xUnit to Halibut, for the following reasons:
* We believe consistency between Octopus Server (which uses xUnit) and other projects is beneficial for developer productivity.
* In order to complete [sc-56164], we must add new mechanisms for accessing the test logs and current test status, and we don't want to implement this feature twice - once now, and once when we eventually migrate to xUnit.

As well as adding the new dependencies, this PR migrates some very simple nUnit tests (`CachingLogFactoryFixture`) to xUnit as a proof of concept.

# Results

There should be no noticeable differences between test runs before and after this migration. The same tests should be run for each build, and test history in TeamCity should continue uninterrupted when a test is converted to xUnit. 

The below screenshot shows the test history for a converted test, showing that the test history is unaffected.

20231002-145738_firefox_rMfy7juGVG

## Before

20231002-145819_firefox_BdpEA6zDcs

## After

20231002-145941_firefox_ycybKiHyXT

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Diagnostics/CachingLogFactoryFixture.cs
+++ b/source/Halibut.Tests/Diagnostics/CachingLogFactoryFixture.cs
@@ -4,13 +4,13 @@ using FluentAssertions;
 using Halibut.Diagnostics;
 using Halibut.Diagnostics.LogCreators;
 using Halibut.Tests.Support.Logging;
-using NUnit.Framework;
+using Xunit;
 
 namespace Halibut.Tests.Diagnostics
 {
     public class CachingLogFactoryFixture
     {
-        [Test]
+        [Fact]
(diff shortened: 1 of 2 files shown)


acodrington:
We have decided to not migrate to xUnit.
* There is not enough benefit to justify the cost (time and effort) of migration
* xUnit is an opinionated testing framework, and does natively support some behaviour we require (i.e. accessing the test context from within the test). We would have to add in some 3rd party plugins to restore the behaviour that nUnit already supports.

While test framework consistency across Octopus projects would be nice to have, this reason alone doesn't justify the investment.

PR #508, opened 2023-10-13
Keep trace logs for failed tests

acodrington:
# Background

Trace logs can provide us with invaluable information about Halibut when debugging problems, but unfortunately they are too verbose to be practical on the build server, as the extra logging causes the build logs to be prohibitively large.

This PR attempts to have the best of both worlds (info-level logs when everything is fine, trace-level logs when there are problems) by logging a test's trace logs to a separate log file which can be kept if the test fails.

# Results

Fixes [sc-56164]

During the setup for each individual test (`BaseTest.SetUp`), a `TraceLogFileLogger` is created and passed to the `SerilogLoggerBuilder` when initializing the logger. This new class is responsible for writing logs to a temporary file, and copying that file to the build artifacts directory if needed.

Once the `SerilogLoggerBuilder.Build` method is called, the instance of `TraceLogFileLogger` is stored in memory for the currently running test and is utilised by a custom Serilog sink which will find and write to the appropriate trace logger for the current test each time a log message is sent.

After the test is completed, the teardown method (`BaseTest.TearDown`) will check the status of the test and have the `TraceLogFileLogger` copy the trace log file to the build artifacts if the test failed.

Trace log files are named _{TestHash}_.tracelog, where _TestHash_ is a short Base64 hash based on the test name. This TestHash is also printed in the log output, allowing a developer to easily find the corresponding trace log file for any failed test.

## Before
No trace logs for failed tests 😞 

## After
20231013-111756_firefox_pNkiCzwGe3

20231013-111549_firefox_Fje2qcr1nq

20231013-140855_jq5jSdEWB4

# How to review this PR

Are there are any other tests or test base class which need to modified to log trace-level logs? I've modified the default for the `IClientAndServiceBuilder` implementations, but are there any tests that _don't_ use these builders that need updating?

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Support/Logging/TestContextConnectionLog.cs
+++ b/source/Halibut.Tests/Support/Logging/TestContextConnectionLog.cs
@@ -4,7 +4,9 @@ using System.Threading;
 using Halibut.Diagnostics;
 using Halibut.Diagnostics.LogWriters;
 using Halibut.Logging;
+using Serilog.Events;
 using ILog = Halibut.Diagnostics.ILog;
+using LogEvent = Halibut.Diagnostics.LogEvent;
 
 namespace Halibut.Tests.Support.Logging
 {
@@ -44,7 +46,7 @@ namespace Halibut.Tests.Support.Logging
             {
                 new SerilogLoggerBuilder().Build()
                     .ForContext<TestContextConnectionLog>()
-                    .Information(string.Format("{5, 16}: {0}:{1} {2}  {3} {4}", logEventLogLevel, logEvent.Error, endpoint, Thread.CurrentThread.ManagedThreadId, logEvent.FormattedMessage, name));
+                    .Write(GetSerilogLevel(logEvent), string.Format("{5, 16}: {0}:{1} {2}  {3} {4}", logEventLogLevel, logEvent.Error, endpoint, Thread.CurrentThread.ManagedThreadId, logEvent.FormattedMessage, name));
             }
         }
 
@@ -62,6 +64,23 @@ namespace Halibut.Tests.Support.Logging
                     return LogLevel.Debug;
                 default:
                     return LogLevel.Info;
+            }
+        }
+
+        static LogEventLevel GetSerilogLevel(LogEvent logEvent)
+        {
+            switch (logEvent.Type)
+            {
+                case EventType.Error:
+                    return LogEventLevel.Error;
+                case EventType.Diagnostic:
+                case EventType.SecurityNegotiation:
+                case EventType.MessageExchange:
+                    return LogEventLevel.Verbose;
+                case EventType.OpeningNewConnection:
+                    return LogEventLevel.Debug;
+                default:
+                    return LogEventLevel.Information;
             }
         }
     }
--- a/source/Halibut.Tests/Support/SerilogLoggerBuilder.cs
+++ b/source/Halibut.Tests/Support/SerilogLoggerBuilder.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Collections.Concurrent;
 using System.IO;
-using System.Linq;
 using System.Security.Cryptography;
 using NUnit.Framework;
 using Serilog;
@@ -14,45 +13,58 @@ namespace Halibut.Tests.Support
     public class SerilogLoggerBuilder
     {
         static readonly ILogger Logger;
+        static readonly ConcurrentDictionary<string, TraceLogFileLogger> TraceLoggers = new();
+
+        TraceLogFileLogger? traceFileLogger;
 
         static SerilogLoggerBuilder()
         {
-            var outputTemplate = 
-                "{TestName} "
+            const string teamCityOutputTemplate =
+                "{TestHash} "
                 + "{Timestamp:HH:mm:ss.fff zzz} "
                 + "{ShortContext} "
                 + "{Message}{NewLine}{Exception}";
 
+            const string localOutputTemplate =
+                "{Timestamp:HH:mm:ss.fff zzz} "
+                + "{ShortContext} "
+                + "{Message}{NewLine}{Exception}";
+
+            var nUnitOutputTemplate = TeamCityDetection.IsRunningInTeamCity()
+                ? teamCityOutputTemplate
+                : localOutputTemplate;
+
             Logger = new LoggerConfiguration()
-                .MinimumLevel.Debug()
-                .WriteTo.Sink(new NonProgressNUnitSink(new MessageTemplateTextFormatter(outputTemplate)))
+                .MinimumLevel.Verbose()
+                .WriteTo.Sink(new NonProgressNUnitSink(new MessageTemplateTextFormatter(nUnitOutputTemplate)), LogEventLevel.Debug)
+                .WriteTo.Sink(new TraceLogsForFailedTestsSink(new MessageTemplateTextFormatter(localOutputTemplate)))
                 .CreateLogger();
+        }
+
+        public SerilogLoggerBuilder SetTraceLogFileLogger(TraceLogFileLogger logger)
+        {
+            this.traceFileLogger = logger;
+            return this;
         }
 
         public ILogger Build()
         {
             // In teamcity we need to know what test the log is for, since we can find hung builds and only have a single file containing all log messages.
-            var testName = "";
-            if (TeamCityDetection.IsRunningInTeamCity())
-            {
-                testName = CurrentTestHash();
-            }
+            var testName = TestContext.CurrentContext.Test.Name;
+            var testHash = CurrentTestHash();
+            var logger = Logger.ForContext("TestHash", testHash);
 
-            var logger = Logger.ForContext("TestName", testName);
-            
-            if (TeamCityDetection.IsRunningInTeamCity())
+            if (traceFileLogger != null)
             {
-                if (!HasLoggedTestHash.Contains(TestContext.CurrentContext.Test.Name))
-                {
-                    HasLoggedTestHash.Add(TestContext.CurrentContext.Test.Name);
-                    logger.Information($"Test: {TestContext.CurrentContext.Test.Name} has hash {CurrentTestHash()}");
-                }
+                TraceLoggers.AddOrUpdate(testName, traceFileLogger, (_, _) => throw new Exception("This should never be updated. If it is, it means that a test is being run multiple times in a single test run"));
+                logger.Information($"Test: {TestContext.CurrentContext.Test.Name} has hash {testHash}");
+                traceFileLogger.SetTestHash(testHash);
             }
 
             return logger;
         }
 
-        public static string CurrentTestHash()
+        static string CurrentTestHash()
         {
             using (SHA256 mySHA256 = SHA256.Create())
             {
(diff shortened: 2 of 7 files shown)


source/Halibut.Tests/Support/Logging/TestContextConnectionLog.cs
```diff
@@ -64,5 +66,22 @@ static LogLevel GetLogLevel(LogEvent logEvent)
                     return LogLevel.Info;
             }
         }
+
+        static LogEventLevel GetSerilogLevel(LogEvent logEvent)
```
sburmanoctopus, line 70: I feel like this should exist somewhere already, and we should just use that. But I can't find anywhere that has this 😑 Oh well 🙂

source/Halibut.Tests/Support/SerilogLoggerBuilder.cs
```diff
@@ -105,5 +115,35 @@ public void Emit(LogEvent logEvent)
                 }
             }
         }
+
+        public class TraceLogsForFailedTestsSink : ILogEventSink
+        {
+            private readonly MessageTemplateTextFormatter _formatter;
+
+            public TraceLogsForFailedTestsSink(MessageTemplateTextFormatter formatter) => _formatter = formatter != null ? formatter : throw new ArgumentNullException(nameof(formatter));
```
sburmanoctopus, line 123: Do we need to guard against nulls? We tend not to do that at Octopus, and rely instead of compile time checking.

source/Halibut.Tests/Support/SerilogLoggerBuilder.cs
```diff
@@ -105,5 +115,35 @@ public void Emit(LogEvent logEvent)
                 }
             }
         }
+
+        public class TraceLogsForFailedTestsSink : ILogEventSink
+        {
+            private readonly MessageTemplateTextFormatter _formatter;
```
sburmanoctopus, line 121: Nit: I love a good underscore, and it's my preferred format. But I believe our house style is to not use the underscore.

Also, I think our house style is to drop the 'private' (again, I prefer to keep it, but it's not my choice 😞 )

source/Halibut.Tests/Support/SerilogLoggerBuilder.cs
```diff
@@ -105,5 +115,35 @@ public void Emit(LogEvent logEvent)
...
+
+            public TraceLogsForFailedTestsSink(MessageTemplateTextFormatter formatter) => _formatter = formatter != null ? formatter : throw new ArgumentNullException(nameof(formatter));
+
+            public void Emit(LogEvent logEvent)
+            {
+                if (logEvent == null)
```
sburmanoctopus, line 127: Nit: old formatting habits die hard. I think we tend to either use brackets for new lines, or keep it on the one line with no brackets.

If you disagree, then ignore this 😁

source/Halibut.Tests/Support/SerilogLoggerBuilder.cs
```diff
@@ -14,58 +14,69 @@ namespace Halibut.Tests.Support
...
         {
-            var outputTemplate = 
-                "{TestName} "
+            const string teamCityOutputTemplate =
+                "{TestHash} "
                 + "{Timestamp:HH:mm:ss.fff zzz} "
                 + "{ShortContext} "
                 + "{Message}{NewLine}{Exception}";
 
+            const string localOutputTemplate =
```
sburmanoctopus, line 29: I noticed we have lost the `TestName` for when it is run locally, as we share this with the `TraceLogsForFailedTestsSink`.

Is that what we want?

source/Halibut.Tests/Support/SerilogLoggerBuilder.cs
```diff
@@ -14,58 +14,69 @@ namespace Halibut.Tests.Support
...
-            if (TeamCityDetection.IsRunningInTeamCity())
+            if (traceFileLogger != null)
             {
-                if (!HasLoggedTestHash.Contains(TestContext.CurrentContext.Test.Name))
-                {
-                    HasLoggedTestHash.Add(TestContext.CurrentContext.Test.Name);
-                    logger.Information($"Test: {TestContext.CurrentContext.Test.Name} has hash {CurrentTestHash()}");
-                }
+                TraceLoggers.AddOrUpdate(testName, traceFileLogger, (_, _) => throw new Exception("This should never be updated. If it is, it means that a test is being run multiple times in a single test run"));
+                logger.Information($"Test: {TestContext.CurrentContext.Test.Name} has hash {testHash}");
```
sburmanoctopus, line 61: Before, we would log this always (on team city). Can we get into a situation where `traceFileLogger` is null and we want to log this?

source/Halibut.Tests/Support/SerilogLoggerBuilder.cs
```diff
@@ -105,5 +115,35 @@ public void Emit(LogEvent logEvent)
...
+            {
+                if (logEvent == null)
+                    throw new ArgumentNullException(nameof(logEvent));
+
+                var testName = TestContext.CurrentContext.Test.FullName;
+
+                if (!TraceLoggers.TryGetValue(testName, out var traceLogger))
+                    throw new Exception($"Could not find trace logger for test '{testName}'");
+
+                var output = new StringWriter();
```
sburmanoctopus, line 135: Once we have the traceLogger, I wonder if it would make more sense to move the "Logging" logic into `TraceLogFileLogger.WriteLine`?

source/Halibut.Tests/TraceLogFileLogger.cs
```diff
@@ -0,0 +1,52 @@
...
+        
+        public void WriteLine(string logMessage)
+        {
+            File.AppendAllLines(tempFilePath, new[] { logMessage });
+        }
+
+        public bool CopyLogFileToArtifacts()
+        {
+            var currentDirectory = Directory.GetCurrentDirectory();
+            var rootDirectory = new DirectoryInfo(currentDirectory).Parent.Parent.Parent.Parent.Parent;
```
sburmanoctopus, line 24: The `.Parent` train here isn't great. Is there not a more legit way of finding the artifact directory?

source/Halibut.Tests/TraceLogFileLogger.cs
```diff
@@ -0,0 +1,52 @@
...
+            var traceLogsDirectory = rootDirectory.CreateSubdirectory("artifacts").CreateSubdirectory("trace-logs");
+            var fileName = $"{testHash}.tracelog";
+
+            try
+            {
+                File.Copy(tempFilePath, Path.Combine(traceLogsDirectory.ToString(), fileName), true);
```
sburmanoctopus, line 30: Could we use 'TestContext.AddTestAttachment' here as well/instead?

source/Halibut.Tests/TraceLogFileLogger.cs
sburmanoctopus: Maybe the test context might have something for this?

sburmanoctopus (approved):
Nice one.

Some nits, and a question on a better way to find the artifact directory, but it should be fine.

source/Halibut.Tests/Support/SerilogLoggerBuilder.cs
acodrington: I'm not 100% I'm interpreting your comment correctly, but I assume you mean that we've lost `TeamName` because the `localOutputTemplate` const doesn't contain it?

The `TestName` field was already empty when running locally, so the logger just didn't print anything out for that field.

Having separate template strings was just an attempt to remove that extra unnecessary space from the start of the log message when we know for certain that the name/hash field is empty.

Let me know if I've misinterpreted and we can talk about it further.

source/Halibut.Tests/Support/SerilogLoggerBuilder.cs
acodrington: Hmmmm, yes actually - we won't log this line anymore if:
* Test fixture doesn't derive from `BaseTest`
* `SerilogLoggerBuilder` is called manually to log some lines, but without initializing a trace logger.

This means it might be hard to find non-`BastTest` tests in the TC build log, as there is no hash printed in the test's logs.

The motivation for moving this line into this `if` block was that we only want this line printed once per test, but it would get printed multiples times if we print it everytime `Build` is called.

I'll look into some other options to deal with this scenario.

source/Halibut.Tests/Support/SerilogLoggerBuilder.cs
acodrington: I would prefer to keep `TraceLogFileLogger` ignorant of any other concerns besides "log this line to a file". I think the sink is a better place for the 'trim/reformat the `SourceContext` value'.

source/Halibut.Tests/TraceLogFileLogger.cs
acodrington: Yeah it's definitely yuck, I just don't know if there is a pre-defined field for "what is the 'project root directory' for this test".

I'll have another look and improve it if I can.

source/Halibut.Tests/TraceLogFileLogger.cs
acodrington: I did try that actually, and it worked well locally but TeamCity threw a wobbly i.e. exceptions in the JVM, so I removed it

When running in TeamCity, the file name for the attached file is something long and unfriendly (I think it might've been a URL-encoded string of the full test name, which gets _really_ long and hard to read.

Maybe I could attach the file to local tests only, so at least it's easier for local dev?

source/Halibut.Tests/Support/SerilogLoggerBuilder.cs
sburmanoctopus: Ah, I see. Sorted 😁

source/Halibut.Tests/TraceLogFileLogger.cs
sburmanoctopus: Without looking hard, there is `TestContext.CurrentContext.WorkDirectory`. That might be helpful? (Or there might be something else in the context area?)

source/Halibut.Tests/TraceLogFileLogger.cs
sburmanoctopus: All good. It sounds like this option was well explored (and found wanting 😁 )

source/Halibut.Tests/TraceLogFileLogger.cs
acodrington: `WorkDirectory` was the one I tried, it's basically the current dir and I still had to navigate 5 levels up via the Parent prop to get to the root.

acodrington: Yeah, [WorkDirectory defaults to the current dir if not explicitly configured](https://octopusdeploy.slack.com/archives/C047M75KF6C/p1697250813966669).

We could configure it via nuke or the TC build, but I feel like it's diminishing returns at this point.

acodrington: There's no neat way of getting this that I can see, but I'm going to add a comment explaining the expected directory structure, so if it ever changes in the future (unlikely) we can update the chain of Parents accordingly.

source/Halibut.Tests/Support/SerilogLoggerBuilder.cs
acodrington: I'm going to re-add the `HasLoggedTestHash` concurrent list, as this is the reason it existed - I was trying to consolidate it with the dictionary of trace loggers, but it serves it's own purpose so we need both. 🤷🏻

PR #513, opened 2023-10-20
Disposing an SslStream that leaves the inner stream open causes it to…

sburmanoctopus:
… flush the stream.

But the underlying NetworkTimeoutStream throws an exception when it flushes and is in a faulted/timed out state.byte

That is what we are fixing.

**_Are you a customer of Octopus Deploy? Please contact [our support team](https://octopus.com/support) so we can triage your PR, so that we can make sure it's handled appropriately._**

# Background

# Results

Fixes https://github.com/OctopusDeploy/Issues/issues/... _(optional public issue)_

Fixes https://github.com/OctopusDeploy/ResearchAndDevelopment/issues/... _(optional private issue)_

See [How we use GitHub Issues](https://github.com/OctopusDeploy/Issues/blob/master/docs/CONTRIBUTING.internal.md) (including [this flowchart](https://whimsical.com/r-d-incoming-work-workflow-aug-21-NsDnGQXcwBLwU66a88Zhue) 

## Before

Before

## After

After

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


pawelpabich:
Thanks for a great PR description. It helped me orient myself quickly 👍.

Your solution seems to solve the problem, but it relies on undocumented behaviours of SSLStream and NetworkStream.  MS is free to change those behaviours at any time, and this will make our code invalid.

Could we approach this problem from a different angle? The error is raised when we dispose the SSLStream. Do we care whether the disposal fails or succeeds? If not, we could use an approach similar to [SafelyCloseStreamAsync](https://github.com/OctopusDeploy/Halibut/blob/07cc14751fc7fe995fc85dbda6cdc3dd68cd8148/source/Halibut/Transport/SecureListener.cs#L334-L344) to make sure no errors are propagated when we dispose of the SSLStream. 

Thoughts?

source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
```diff
@@ -30,6 +30,8 @@ public override async ValueTask DisposeAsync()
 
         public override async Task FlushAsync(CancellationToken cancellationToken)
         {
+            ThrowIfAlreadyTimedOut();
```
sburmanoctopus, line 33: I moved this out from `WrapWithCancellationAndTimeout`, because `WrapWithCancellationAndTimeout` never ended up calling `ThrowIfAlreadyTimedOut`.

This was because `CanTimeout` is called outside `WrapWithCancellationAndTimeout`, which also calls `ThrowIfAlreadyTimedOut`. So when you look at the stack trace, it goes via `CanTimeout`. 

Which is just a little odd. 

Really, this type of thing should be at the public entry point level, not within a private.

source/Halibut/Util/SilentStreamDisposer.cs
```diff
@@ -0,0 +1,31 @@
...
+        
+        public async ValueTask DisposeAsync()
+        {
+            try
+            {
+                await streamToDispose.DisposeAsync();
+            }
+            catch (Exception ex)
+            {
+                onFailure?.Invoke(ex);
```
pawelpabich, line 27: Not sure how valuable the exception here is, but if we want to log something here, then we could do this at the Verbose level.

source/Halibut/Util/Try.cs
```diff
@@ -15,6 +16,11 @@ public static void CatchingError(Action tryThisAction, Action<Exception> onFailu
                 onFailure(e);
             }
         }
+
+        public static SilentStreamDisposer CatchingErrorOnDisposal(Stream streamToDispose, Action<Exception> onFailure = null)
```
pawelpabich, line 20: YAGNI, if don't use this paramter then let's remove it :)

PR #514, opened 2023-10-23
Add support for derived types to the default TypeRegistry

APErebus:
# Background

If you supply a complex base type as part of the request message, the `TypeRegistry` doesn't know how to resolve any derived types.

This PR updates the `TypeRegistry` to check the assembly the base type is in as well as any extra assemblies provided during setup for any derived types.

# Results

I chose to implement a new `TypeRegistryBuilder` to allow consumers to configure the built-in type registry without having to create their own (similar to how we can configure the Message Serializer as part of the `HalibutRuntimeBuilder`.

We also only look for `ExportedTypes()` in the assemblies, so types that are `private/protected/internal` are not registered.

I've added some unit tests to cover the new serialization behaviour.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/UsageFixture.cs
+++ b/source/Halibut.Tests/UsageFixture.cs
@@ -278,6 +278,7 @@ namespace Halibut.Tests
                 .Build(CancellationToken);
 
             var service = clientAndService.CreateClient<IComplexObjectService, IAsyncClientComplexObjectService>();
+
             var request = new ComplexObjectWithInheritance
             {
                 Child1 = new ComplexInheritedChild1(childPayload1),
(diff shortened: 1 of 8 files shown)


acodrington (approved):
LGTM 👍🏻 :shipit:

source/Halibut.Tests/UsageFixture.cs
```diff
@@ -257,12 +257,43 @@ public async Task OctopusCanSendAndReceiveComplexObjects_WithMultipleComplexChil
...
+
                 response.Child2.Should().NotBeSameAs(request.Child2);
                 response.Child2.EnumPayload.Should().Be(enumValue);
                 response.Child2.ComplexPayloadSet.Should().NotBeSameAs(request.Child2.ComplexPayloadSet);
                 response.Child2.ComplexPayloadSet.Select(x => new ComplexPair<string>(x.EnumValue, x.Payload.ReadAsString())).ToHashSet().Should().BeEquivalentTo(set);
             }
         }
+
+        [Test]
+        [LatestClientAndLatestServiceTestCases(testNetworkConditions: false)]
```
APErebus, line 269: After discussion with @acodrington, we agreed that it doesn't make sense to do the backwards compatible test cases because the old versions won't have the type registry changes needed, thus they won't work.

PR #524, opened 2023-11-02
Fix handling of errors in async services.

LukeButters:
# Background

Fixes an issue where the exception thrown by an async service would not be wrapped in a `InvocationTargetException`, resulting in the client being unable to determine that the exception occurred within the service itself.

The change is to wrap the exception thrown by the async service in a  `InvocationTargetException` which means the halibut service invoker now always throws a `InvocationTargetException` for exceptions thrown in sync or async services.

An async service is one that looks like: (e.g. `class AsyncEchoService : IAsyncEchoService { public Task Foo() {await Task.Complete;}}) 

# How to review this PR

Review the two commits separately, the first modifies the test infrastructure and the second adds the fix and tests.

### DIFF
--- a/source/Halibut.TestUtils.CompatBinary.Base/DelegateEchoService.cs
+++ b/source/Halibut.TestUtils.CompatBinary.Base/DelegateEchoService.cs
@@ -29,7 +29,16 @@ namespace Halibut.TestUtils.SampleProgram.Base
         public bool Crash()
         {
             Console.WriteLine("Forwarding Crash() call to delegate");
-            return echoService.Crash();
+            try
+            {
+                return echoService.Crash();
+            }
+            catch (Exception e)
+            {
+                throw new Exception("Received Exception: START:\n" +
+                                    e.Message
+                                    + "\nEnd Received Exception");
+            }
         }
 
         public int CountBytes(DataStream dataStream)
--- a/source/Halibut.Tests/FailureModesFixture.cs
+++ b/source/Halibut.Tests/FailureModesFixture.cs
@@ -73,7 +73,6 @@ namespace Halibut.Tests
         public async Task FailsWithInvocationServiceExceptionWhenAsyncServiceCrashes(ClientAndServiceTestCase clientAndServiceTestCase)
         {
             await using (var clientAndService = await clientAndServiceTestCase.CreateTestCaseBuilder()
-                             //.AsPreviousClientVersionAndLatestServiceBuilder()
                              .WithAsyncService<IEchoService, IAsyncEchoService>(() => new AsyncEchoService())
                              .Build(CancellationToken))
             {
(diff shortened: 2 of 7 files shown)


source/Halibut.TestUtils.CompatBinary.Base/DelegateEchoService.cs
```diff
@@ -29,7 +29,16 @@ public string SayHello(string name)
         public bool Crash()
         {
             Console.WriteLine("Forwarding Crash() call to delegate");
-            return echoService.Crash();
+            try
+            {
+                return echoService.Crash();
+            }
+            catch (Exception e)
```
sburmanoctopus, line 36: Just wondering, why did we wrap this?

source/Halibut.Tests/FailureModesFixture.cs
```diff
@@ -63,6 +63,39 @@ await using (var clientAndService = await clientAndServiceTestCase.CreateTestCas
...
+        /// InvocationServiceException is how clients know that the failure occured within the service, so we check this is returned in the ResponseMessage
+        /// </summary>
+        /// <param name="clientAndServiceTestCase"></param>
+        [Test]
+        [LatestClientAndLatestServiceTestCases(testNetworkConditions: false, testSyncClients: false, testAsyncClients: true, testSyncService: false, testAsyncServicesAsWell: true)]
+        [PreviousClientAndLatestServiceVersionsTestCases(testNetworkConditions: false, testSyncClients: false, testAsyncClients: true, testSyncService: false, testAsyncServicesAsWell: true)]
+        public async Task FailsWithInvocationServiceExceptionWhenAsyncServiceCrashes(ClientAndServiceTestCase clientAndServiceTestCase)
+        {
+            await using (var clientAndService = await clientAndServiceTestCase.CreateTestCaseBuilder()
+                             //.AsPreviousClientVersionAndLatestServiceBuilder()
```
sburmanoctopus, line 76: ??

source/Halibut.TestUtils.CompatBinary.Base/DelegateEchoService.cs
LukeButters: Otherwise its hard to determine what the actual service returned, we need to know the error message is a `InvocationTargetException`

sburmanoctopus: Just wondering, we can leave it as a string wrapper. Or would it be better making a custom exception, and throw that instead?

LukeButters: The custom exception gets `toString()` called on it and then placed into the message section of the Exception sent over the wire bringing us back to square one.

sburmanoctopus (approved):
LGTM

PR #525, opened 2023-11-03
Co-operatively cancel all RPC Calls

nathanwoctopusdeploy:
# Background

[sc-58429]

Before Async Halibut was introduced, Halibut only supported cancellation of an RPC call when the request was in the connecting state, queued for a Polling Request and the TCP Connection being established for a Listening Request. Once the RPC was in flight, dequeued for a Polling Request, and the connection established for a Listening Request, cancellation was not supported.

As the cancellation of an in-flight request was not supported, a caller would only have the option to walk away from an RPC call and leave threads and IO running. 

When Async Halibut was introduced, cooperative cancellation was introduced down to the socket for RPC calls to Listening Service and to the queue for Polling Services. Cancellation could be performed if the RPC call was connecting and in-flight. So as not to force clients to change there existing behaviour, two cancellation tokens were introduced, one to cancel when connecting and one to cancel when in-flight.

# Results

This PR extends the previous cancellation support so that an in-flight Polling Request can be cancelled to the Socket. Terminating any in-flight Reads or Writes on the TCP Connection.

The two cancellation tokens have been combined into one to simplify the cancellation logic. The ability to cancel just when connecting is no longer required in TentacleClient. Instead, the desire is to co-operatively cancel an RPC call regardless of it's current state.

Tentacle Client does need to know if the cancelled RPC call was connecting or in flight to make the most efficient decisions about what it has to do next, so cancellation will throw specific errors based on the state of the RPC call when it was cancelled.

net48 does not support full cancellation due to the usage of a deflate stream, which uses APM Begin and End methods and loses the cancellation token. As TentacleClient only supports net6.0, this is not an issue.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Halibut/ServiceModel/PendingRequestQueueAsync.cs
```diff
@@ -270,16 +274,27 @@ public async Task WaitUntilComplete(RequestCancellationTokens requestCancellatio
...
+                    using var linkedTokenSource = CancellationTokenSource.CreateLinkedTokenSource(timeoutCancellationTokenSource.Token, cancellationToken);
                     await responseWaiter.WaitAsync(linkedTokenSource.Token);
                 }
-                catch (OperationCanceledException)
+                catch (OperationCanceledException ex)
                 {
-                    if (cancellationTokenSource.IsCancellationRequested) return false;
-                    throw;
+                    // Cancel the Request to force cancellation to the socket if it is currently being transferred 
+                    CancelRequest();
```
nathanwoctopusdeploy, line 287: This signals that we want to cancel the RPC call to the dequeuing thread.

source/Halibut/Exceptions/RequestCancelledException.cs
```diff
@@ -0,0 +1,47 @@
+using System;
+
+namespace Halibut.Exceptions
```
nathanwoctopusdeploy, line 3: These exceptions are used to determine where cancellation occurred. Needed in TentacleClient

source/Halibut/ServiceModel/HalibutProxyRequestOptions.cs
```diff
@@ -1,34 +1,17 @@
...
-        /// This is a best effort cancellation and is not guaranteed.
-        /// For Sync Halibut, providing this cancellation token is not supported.
-        /// For Async Halibut this will attempt to cancel the RPC call.
-        /// If the call is to a Listening Service, then cancellation is performed all the way down to the Socket operations.
-        /// if the call is to a Polling Service, then cancellation is performed all the way down to the Polling Queue,
-        /// this means the client can cancel the call but the service will still process the request and return a response.
+        /// When cancelled, will cancel a connecting or in-progress/in-flight RPC call.
         /// </summary>
-        public CancellationToken? InProgressRequestCancellationToken { get; }
+        public CancellationToken RequestCancellationToken { get; }
```
nathanwoctopusdeploy, line 10: Simplify to a single cancellation token as we don't need to retain the ability to only cancel connecting requests

source/Halibut/ServiceModel/HalibutProxyWithAsync.cs
```diff
@@ -18,20 +18,17 @@ public class HalibutProxyWithAsync : DispatchProxyAsync
         ServiceEndPoint endPoint;
         long callId;
         bool configured;
-        CancellationToken globalCancellationToken;
```
nathanwoctopusdeploy, line 21: This was a left over from SyncHalibut and not used any more

source/Halibut/Transport/Protocol/MessageExchangeProtocol.cs
```diff
@@ -207,26 +208,36 @@ async Task ProcessSubscriberAsync(IPendingRequestQueue pendingRequests, Cancella
...
-        async Task<bool> ProcessReceiverInternalAsync(IPendingRequestQueue pendingRequests, RequestMessage nextRequest, CancellationToken cancellationToken)
+        async Task<bool> ProcessReceiverInternalAsync(IPendingRequestQueue pendingRequests, RequestMessageWithCancellationToken nextRequest, CancellationToken cancellationToken)
         {
             try
             {
                 if (nextRequest != null)
                 {
-                    var response = await SendAndReceiveRequest(nextRequest, cancellationToken);
-                    await pendingRequests.ApplyResponse(response, nextRequest.Destination);
+                    using var linkedTokenSource = CancellationTokenSource.CreateLinkedTokenSource(nextRequest.CancellationToken, cancellationToken);
```
nathanwoctopusdeploy, line 217: The cancellation token on the Request is now observed so we can cancel the thread that is dequeuing the pending request

source/Halibut/Transport/Protocol/RequestMessageWithCancellationToken.cs
```diff
@@ -0,0 +1,16 @@
+using System.Threading;
+
+namespace Halibut.Transport.Protocol
+{
+    public class RequestMessageWithCancellationToken
```
nathanwoctopusdeploy, line 5: This contains the request message and a cancellation token that the dequeuing thread can observe to allow us to cancel across the queue.

source/Halibut/Transport/Streams/NetworkTimeoutStream.cs
```diff
@@ -30,7 +30,7 @@ public override async ValueTask DisposeAsync()
 
         public override async Task FlushAsync(CancellationToken cancellationToken)
         {
-            ThrowIfAlreadyTimedOut();
+            ThrowIfAlreadyCancelledOrTimedOut();
```
nathanwoctopusdeploy, line 33: The method and properties were a slight lie as it can be timed out but it can also be cancelled

source/Halibut/ServiceModel/HalibutProxyWithAsync.cs
```diff
@@ -72,9 +69,7 @@ async Task<(MethodInfo, object)> MakeRpcCall(MethodInfo asyncMethod, object[] ar
 
             var request = CreateRequest(asyncMethod, serviceMethod, args);
 
-            using var requestCancellationTokens = RequestCancellationTokens(halibutProxyRequestOptions);
-
-            var response = await messageRouter(request, serviceMethod, requestCancellationTokens);
+            var response = await messageRouter(request, serviceMethod, halibutProxyRequestOptions?.RequestCancellationToken ?? CancellationToken.None);
```
sburmanoctopus, line 72: Just wondering, what situation would `halibutProxyRequestOptions` be null in? I thought put this on all async methods?

sburmanoctopus: (I'm seeing if we can simplify it and ensure `halibutProxyRequestOptions` will always be there 😁 )

source/Halibut/ServiceModel/HalibutProxyRequestOptions.cs
```diff
@@ -1,34 +1,17 @@
...
 
-        public HalibutProxyRequestOptions(CancellationToken? connectingCancellationToken, 
-            CancellationToken? inProgressRequestCancellationToken)
+        public HalibutProxyRequestOptions(CancellationToken cancellationToken)
```
sburmanoctopus, line 12: I'll probably form a stronger idea as I go through the PR, but just looking at this class in isolation, do we need it anymore now that it has just a `CancellationToken`? I.e. can we simply replace all usages of `HalibutProxyRequestOptions` with a `CancellationToken`?

source/Halibut/Transport/Protocol/MessageExchangeProtocol.cs
```diff
@@ -253,18 +264,18 @@ async Task<bool> ProcessReceiverInternalAsync(IPendingRequestQueue pendingReques
             return true;
         }
         
-        async Task<ResponseMessage> SendAndReceiveRequest(RequestMessage nextRequest, CancellationToken cancellationToken)
+        async Task<ResponseMessage> SendAndReceiveRequest(RequestMessageWithCancellationToken nextRequest, CancellationToken cancellationToken)
```
sburmanoctopus, line 267: Nit: Instead of passing `RequestMessageWithCancellationToken` as the first parameter and using `nextRequest.RequestMessage` in this method, maybe pass in the message itself.

We don't use the cancellation token in this method, so it scopes it down.

But more importantly, I think it makes it more clear that the cancellation token we give this method is the one to use. Otherwise, someone might ask "why aren't we using the cancellation token from `RequestMessageWithCancellationToken`?"

source/Halibut/Transport/Protocol/ResponseMessage.cs
```diff
@@ -32,12 +33,14 @@ public static ResponseMessage FromException(RequestMessage request, Exception ex
 
         internal static ServerError ServerErrorFromException(Exception ex)
         {
-            string ErrorType = null;
```
sburmanoctopus, line 35: Lol, `ErrorType` 🤦

source/Halibut/Transport/TcpClientExtensions.cs
```diff
@@ -43,6 +43,7 @@ void DisposeClient()
             {
                 try
                 {
+                    client.CloseImmediately();
```
sburmanoctopus, line 46: Just wondering, what was the motivation for this?

source/Halibut.Tests/WhenCancellingARequestForAPollingTentacle.cs
```diff
@@ -39,7 +37,8 @@ await using (var clientAndService = await clientAndServiceTestCase.CreateTestCas
                 {
                     var doSomeActionService = clientAndService.CreateAsyncClient<IDoSomeActionService, IAsyncClientDoSomeActionServiceWithOptions>();
 
-                    await AssertAsync.Throws<OperationCanceledException>(() => doSomeActionService.ActionAsync(halibutProxyRequestOptions));
+                    (await AssertAsync.Throws<Exception>(() => doSomeActionService.ActionAsync(halibutProxyRequestOptions)))
```
sburmanoctopus, line 40: Just wondering, could we just check `Throws<ConnectingRequestCancelledException>` instead of `.And.Should().Match(x => x is ConnectingRequestCancelledException);`?

source/Halibut/ServiceModel/HalibutProxyWithAsync.cs
nathanwoctopusdeploy: Good question, short answer I'm not sure without more investigation, I assumed it would never be null originally, and it blew up a lot of tests so reverted to the original logic of defaulting if null.

source/Halibut/ServiceModel/HalibutProxyRequestOptions.cs
nathanwoctopusdeploy: It's an extension point and I do plan to add something to it during sharpening so would like it to stay around.

source/Halibut.Tests/WhenCancellingARequestForAPollingTentacle.cs
sburmanoctopus: Whatever the outcome, there are other instances of this type of check below as well

source/Halibut/Transport/TcpClientExtensions.cs
nathanwoctopusdeploy: I was having a whole lot of trouble making things cancel to the socket and thought something might be wrong in async Halibut so started trying to ensure everything was closed and disposed.  Turned out to be a pretty stupid mistake in my test but I did add this, which seemed right and decided to leave it. 

I'm happy to revert from this PR, as you said it's unrelated, and we can add in another PR to remove multiple concerns being changed.

source/Halibut.Tests/WhenCancellingARequestForAPollingTentacle.cs
nathanwoctopusdeploy: Yes, good catch. These can be simplified now. It's evolved a bit, and started out having to test a type and property of the exception but that's all changed.

nathanwoctopusdeploy: Done

source/Halibut/Transport/TcpClientExtensions.cs
nathanwoctopusdeploy: Done

source/Halibut/Transport/Protocol/MessageExchangeProtocol.cs
nathanwoctopusdeploy: Done

source/Halibut/ServiceModel/HalibutProxyWithAsync.cs
nathanwoctopusdeploy: A quick look shows that we do not enforce that a HalibutProxyRequestOptions has to be provided when making an RPC call and 100's of our tests don't provide it.

It probably makes sense to charge this and enforce the parameter is supplied so it's used in the best way e.g. cancellation tokens passed etc. I'll card it.

source/Halibut.Tests/WhenCancellingARequestForAPollingTentacle.cs
```diff
@@ -206,50 +187,61 @@ public async Task<ResponseMessage> QueueAndWaitAsync(RequestMessage request, Req
...
-
-            public CancelWhenRequestDequeuedPendingRequestQueueFactory(IPendingRequestQueueFactory inner, CancellationTokenSource cancellationTokenSource): this(inner, new []{ cancellationTokenSource })
-            {
+                this.onResponseApplied = onResponseApplied;
             }
 
             public IPendingRequestQueue CreateQueue(Uri endpoint)
             {
-                return new Decorator(inner.CreateQueue(endpoint), cancellationTokenSources);
+                return new Decorator(inner.CreateQueue(endpoint), cancellationTokenSource, shouldCancel, onResponseApplied);
```
sburmanoctopus, line 203: Nit: maybe `shouldCancelOnDequeue`?

sburmanoctopus (approved):
LGTM

source/Halibut.Tests/WhenCancellingARequestForAPollingTentacle.cs
nathanwoctopusdeploy: Done

sburmanoctopus: Awesome! 

Although, I noticed the renamed stopped happening from line 209 downwards.

Just a nit 🙂

PR #526, opened 2023-11-06
Record all log lines, and don't let logging alter the test run

LukeButters:
# Background

An issue exists in which recording trace logs appears to:
- Not record all log lines, since logging can be done in parallel and this results in an exception when two threads write to the same file.
- That exception would bubble out of the sink, although maybe something catches it before it throws an exception in halibut.
- Appeared to alter the timings of test causing some web-socket tests to fail. (Which is probably a real bug?)

# Results

Logging is now done to a queue, which is async written to disk. This means:
- We don't throw exceptions in the sink.
- Writing logs is thread safe.
- Each log line written does not `open, write to, and close a file` which should stop changing the timings of tests and improve performance/decrease disk IO.

The queue is used over a lock to avoid logging introducing a hidden lock which could cause code to become dependent on the lock.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Halibut.Tests.csproj
+++ b/source/Halibut.Tests/Halibut.Tests.csproj
@@ -36,6 +36,7 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.EnvironmentVariables" Version="7.0.0" />
     <PackageReference Include="Assent" Version="1.8.2" />
     <PackageReference Include="FluentAssertions" Version="6.8.0" />
+    <PackageReference Include="Microsoft.VisualStudio.Threading" Version="17.7.30" />
     <PackageReference Include="NSubstitute" Version="4.4.0" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.3.2" />
     <PackageReference Include="NUnit" Version="3.13.3" />
--- a/source/Halibut.Tests/TraceLogFileLogger.cs
+++ b/source/Halibut.Tests/TraceLogFileLogger.cs
@@ -2,7 +2,6 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Text;
-using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.VisualStudio.Threading;
 
@@ -10,18 +9,18 @@ namespace Halibut.Tests
 {
     public class TraceLogFileLogger : IDisposable
     {
-        readonly AsyncQueue<string> queue = new AsyncQueue<string>();
+        readonly AsyncQueue<string> queue = new();
         readonly string tempFilePath = Path.GetTempFileName();
         string testHash;
 
-        readonly TaskCompletionSource<bool> noMoreDelaysToWritingLogs;
-        readonly Task noMoreDelaysTask;
+        readonly TaskCompletionSource<bool> prepareForLogCollectionTaskCompletionSource;
+        readonly Task prepareForLogCollection;
         readonly Task writeDataToDiskTask;
 
         public TraceLogFileLogger()
         {
-            noMoreDelaysToWritingLogs = new TaskCompletionSource<bool>();
-            noMoreDelaysTask = noMoreDelaysToWritingLogs.Task;
+            prepareForLogCollectionTaskCompletionSource = new TaskCompletionSource<bool>();
+            prepareForLogCollection = prepareForLogCollectionTaskCompletionSource.Task;
             writeDataToDiskTask = WriteDataToFile();
         }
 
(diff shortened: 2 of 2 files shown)


source/Halibut.Tests/TraceLogFileLogger.cs
```diff
@@ -1,25 +1,82 @@
...
             this.testHash = testHash;
         }
         
         public void WriteLine(string logMessage)
         {
-            File.AppendAllLines(tempFilePath, new[] { logMessage });
+            queue.Enqueue(logMessage);
+        } 
+        
+        async Task WriteDataToFile()
```
LukeButters, line 38: This task is never terminated, my guess it wont be an issue since we don't have enough tests for this to become an issue.

source/Halibut.Tests/TraceLogFileLogger.cs
```diff
@@ -1,25 +1,82 @@
...
+
         public void SetTestHash(string testHash)
         {
             this.testHash = testHash;
         }
         
         public void WriteLine(string logMessage)
         {
-            File.AppendAllLines(tempFilePath, new[] { logMessage });
+            queue.Enqueue(logMessage);
```
LukeButters, line 35: The queue here probably does have some locking under the hood. I suspect the time that lock is held for is so small that it wont impact tests, compared to a lock held for the duration of a file append done on every trace or higher log event raised.

source/Halibut.Tests/TraceLogFileLogger.cs
```diff
@@ -1,25 +1,82 @@
...
+                }
+
+                var list = new List<string>();
+                list.Add(await queue.DequeueAsync());
+                while (queue.TryDequeue(out var log))
+                {
+                    list.Add(log);
+                }
+
+                using (var fileAppender = new StreamWriter(tempFilePath, true, Encoding.UTF8, 8192))
```
LukeButters, line 55: default buffer size is 1KB, i chose 8KB to try and minimise disk writes which tents to be very slow on windows VMs.

source/Halibut.Tests/Halibut.Tests.csproj
```diff
@@ -36,6 +36,7 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.EnvironmentVariables" Version="7.0.0" />
     <PackageReference Include="Assent" Version="1.8.2" />
     <PackageReference Include="FluentAssertions" Version="6.8.0" />
+    <PackageReference Include="Microsoft.VisualStudio.Threading" Version="17.7.30" />
```
sburmanoctopus, line 39: Is this package needed? What does it bring to the table?

sburmanoctopus: Ah, it's for `AsyncQueue`

source/Halibut.Tests/TraceLogFileLogger.cs
```diff
@@ -1,25 +1,82 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.Threading;
 
 namespace Halibut.Tests
 {
     public class TraceLogFileLogger : IDisposable
     {
-        string tempFilePath = Path.GetTempFileName();
+        readonly AsyncQueue<string> queue = new AsyncQueue<string>();
```
sburmanoctopus, line 13: Why not a `ConcurrentQueue`? We tend to use those more, so I'm wondering why we went this special type of queue that uses `VisualStudio` namespaces instead.

source/Halibut.Tests/TraceLogFileLogger.cs
```diff
@@ -1,25 +1,82 @@
...
+            while (true)
+            {
+                // Don't hammer the disk, let some log message queue up before writing them.
+                if (!noMoreDelaysTask.IsCompleted)
+                {
+                    await Task.WhenAny(Task.Delay(5), noMoreDelaysTask);
+                }
+
+                var list = new List<string>();
+                list.Add(await queue.DequeueAsync());
```
sburmanoctopus, line 49: Why do we dequeue, then do a loop of TryDequeue?

I would have thought the loop would have been enough, right?

source/Halibut.Tests/TraceLogFileLogger.cs
LukeButters: No async methods :(

source/Halibut.Tests/TraceLogFileLogger.cs
LukeButters: Because we want one call the first to block until we have something, then we want to read everything we have. To then write all of that in one go.

source/Halibut.Tests/TraceLogFileLogger.cs
```diff
@@ -1,25 +1,82 @@
...
+
+                    await fileAppender.FlushAsync();
+                }
+            }
         }
+        
+        void AllowSomeTimeForTheQueueToEmpty()
+        {
+            noMoreDelaysToWritingLogs.SetResult(false);
+            for (int i = 0; i < 100; i++)
```
sburmanoctopus, line 70: Why do we only wait a little, instead of waiting for all log entries to be flushed?

sburmanoctopus: Is it because the same loop is used for all tests, so we don't actually ever shut it down?

sburmanoctopus: But if it's the same task for all tests, then we never reset `SetResult`. Maybe that's not the answer then

LukeButters: Yeah good question, I was wondering what should we do if all the logs never get written because IO is too slow or something keeps spamming out to the logs.

Maybe waiting 100ms is not long enough, and instead we should wait longer? Or we could just wait forever and hope we can eventually have everything written....

LukeButters: Each tests gets its own `TraceLogFileLogger`

source/Halibut.Tests/TraceLogFileLogger.cs
sburmanoctopus: Ah, I see. Might be worth adding a comment about that 🙂

source/Halibut.Tests/TraceLogFileLogger.cs
borland: Mmmm... This is only a test DLL not the actual product so it's not so bad, but I'm still fairly 👎  on overly-clever ConcurrentCollection type stuff.

source/Halibut.Tests/TraceLogFileLogger.cs
borland: Is the idea here that there are multiple threads adding log entries to the queue but only one thread dequeueing and writing?

source/Halibut.Tests/TraceLogFileLogger.cs
sburmanoctopus: My gut feel is that we should terminate the writing loop thread/task, wait for that to finish, then write whatever is in the queue. And simply wait for all that to happen.

LukeButters: The thing is, things can write to the logs after this method is called. Actually they do right after we call this method we log another thing :D

LukeButters: OK updated it to wait for the write task to finish.

acodrington (approved):
Some nice improvements here 🙇🏻 👍🏻 :shipit:

sburmanoctopus (approved):
LGTM

LukeButters:
@borland points out it might make more sense to use a serilog file sink. That probably would make more sense, since I suspect we could just write directly to the artifact's folder and it looks like it supports buffering so maybe everything is done out of the critical I am not sure. Talking to @acodrington he found that the serilog file sinks didn't do exactly what we needed.

PR #527, opened 2023-11-06
Reduce StandardIterationCount on linux

LukeButters:
# Background

Linux is slower than windows and this does impact the time it takes to run all of the tests, which especially frustrating for local development.

This reduces the iterations for listening from 1000 to 250 on linux, which is probably still well enough.

Some tests go from `1m38s` to `28s`.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


sburmanoctopus (approved):
If you're happy that we are doing enough iterations on Linux, then I'm happy too 😁

PR #544, opened 2023-11-16
Improve how Stream Read and Write Timeouts are changed

nathanwoctopusdeploy:
# Background

Improve the way the read and write stream timeouts are changed for a particular operation.

# Results

This PR does not change any timeouts, just changes the API for how those timeouts are changed.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Halibut/Transport/Protocol/MessageExchangeStream.cs
```diff
@@ -63,27 +64,29 @@ async Task SendIdentityMessageAsync(string identityLine, CancellationToken cance
...
+                MessageExchangeStreamTimeout.ControlMessageExchangeShortTimeout,
+                async () => await SendControlMessageAsync(Next, cancellationToken));
         }
 
         public async Task SendProceedAsync(CancellationToken cancellationToken)
         {
-            await SendControlMessageAsync(Proceed, cancellationToken);
+            await WithTimeout(
+                MessageExchangeStreamTimeout.ControlMessageExchangeShortTimeout,
+                async () => await SendControlMessageAsync(Proceed, cancellationToken));
```
sburmanoctopus, line 76: As we discussed, this is different to what was there before, but we believed that this message does not need the full timeout set, and would make sense to use the lower timeout.

source/Halibut/Transport/Protocol/MessageExchangeStreamTimeout.cs
```diff
@@ -0,0 +1,10 @@
+using System;
+
+namespace Halibut.Transport.Protocol
+{
+    public enum MessageExchangeStreamTimeout
```
sburmanoctopus, line 5: 💭 Just a thought, but I wonder if an alternative to this enum is to acknowledge that the send/receive timeouts are a concept, and to make a class instead (which has a send timeout property and a receive timeout property), and replace the 4 `HalibutTimeoutsAndLimits` properties (TcpClientSendTimeout, TcpClientReceiveTimeout, TcpClientHeartbeatSendTimeout, TcpClientHeartbeatReceiveTimeout) with just 2 properties using this new class.

With this remodelling, we would not need this enum, and could simply pass that object into the extension methods below.

Anyway, not vital for this PR, just thought I'd float the idea 😁

sburmanoctopus (approved):
LGTM, nice one

source/Halibut/Transport/Protocol/MessageExchangeStream.cs
nathanwoctopusdeploy: Ahh, good spot. I did not mean for this to make it into this PR so will revert!!

nathanwoctopusdeploy: For a future PR

source/Halibut/Transport/Protocol/MessageExchangeStreamTimeout.cs
nathanwoctopusdeploy: Yeh that would be a lot cleaner 👍 I'll rework

nathanwoctopusdeploy: I've reworked the change, thoughts @sburmanoctopus

sburmanoctopus: Nice one! LGTM 😁

PR #548, opened 2023-11-22
Thread-safety and nullability for some tests

borland:
# Background

While looking into future .NET 8 support, I encountered some small issues in the tests. I've extracted these out of the .NET 8 work as they aren't connected to .NET 8 versioning

# Results

- Fix certificate tests which could fail due to a race condition between asserting a result, and the result being captured
- `Halibut.Tests.csproj` referrred to TargetFramework `net60` in one place, it should have been `net6.0`
- Thread-safety in the SyncIoRecording classes in Halibut.Tests
- Fix various null-safety warnings in `Octopus.TestPortForwarder`; this now builds cleanly

# How to review this PR

Quality :heavy_check_mark:

These changes are all in the tests. By that definition, if the tests pass we should be good to merge.
I ran them all locally on my windows PC and they are all passing

# Pre-requisites


source/Octopus.TestPortForwarder/PortForwarder.cs
```diff
@@ -48,13 +49,15 @@ public class PortForwarder : IDisposable
...
-            ListeningPort = ((IPEndPoint)listeningSocket.LocalEndPoint).Port;
+            var ipEndPoint = listeningSocket.LocalEndPoint as IPEndPoint ?? throw new InvalidOperationException("listeningSocket.LocalEndPoint was not an IPEndPoint");
+            
+            ListeningPort = ipEndPoint.Port;
             PublicEndpoint = new UriBuilder(scheme, "localhost", ListeningPort).Uri;
 
             Task.Factory.StartNew(() => WorkerTask(cancellationTokenSource.Token).ConfigureAwait(false), TaskCreationOptions.LongRunning);
         }
 
+        [MemberNotNull(nameof(listeningSocket))]
```
borland, line 60: This tells the code in the constructor that `Start()` is guaranteed to initialize `listeningSocket` so it knows it won't be null there

source/Halibut.Tests/BadCertificatesTests.cs
```diff
@@ -29,6 +29,9 @@ public async Task SucceedsWhenPollingServicePresentsWrongCertificate_ButServiceI
             var clientTrustProvider = new DefaultTrustProvider();
             var unauthorizedThumbprint = "";
             var firstCall = true;
+            
+            var unauthorizedClientHasConnected = new TaskCompletionSource<bool>();
+            CancellationToken.Register(() => unauthorizedClientHasConnected.TrySetCanceled()); // backup to fail the test in case it never connects
```
LukeButters, line 34: nit: could be unregistered

LukeButters (approved):

source/Halibut.Tests/BadCertificatesTests.cs
borland: There's no value in unregistering things in the same object graph.

Put another way:

On the "cancelled" code path:
When the CancellationToken fires, the test will tear down, the class/variable/etc will go out of scope and the garbage collector will eat the whole lot in one go.

On the "normal" code path:
TearDown in BaseTest will cancel and dispose the CancellationTokenSource, which will clean up all this stuff; the garbage collector will eat the whole lot there too.

PR #549, opened 2023-11-22
Add Polling Service next request timeout

LukeButters:
# Background

[[SC-59505]]

Adds timeout:
```
        /// <summary>
        ///     Amount of time the polling service will wait for a request from an RPC call.
        ///     Specifically the amount of time to wait for the first byte of the request to arrive.
        ///
        ///     This value must never be less than the PollingQueueWaitTimeout, of the client, since this
        ///     is the timeout of the long poll the polling service makes to the client to get the next request.
        ///
        ///     Currently set to 10 minutes as that is what the timeout used to be. 
        /// </summary>
        public TimeSpan TcpClientReceiveRequestTimeoutForPolling { get; set; } = TimeSpan.FromMinutes(10);
```

Which will allow as to reduce the time the polling service will wait for the beginning of a `Response` message to be received since we know the polling queue will send `null` messages every `30s` if it has no Requests to make. This makes it possible to have the service more readily  detect a stalled connection.

This value only makes sense to be set in the polling service.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Diagnostics/HalibutTimeoutsAndLimits.cs
+++ b/source/Halibut/Diagnostics/HalibutTimeoutsAndLimits.cs
@@ -51,8 +51,21 @@ namespace Halibut.Diagnostics
 
         /// <summary>
         ///     Amount of time to wait for a response from an RPC call.
+        /// 
+        ///     Specifically the amount of time to wait for the first byte of the request to arrive.
         /// </summary>
         public TimeSpan TcpClientReceiveResponseTimeout { get; set; } = TimeSpan.FromMinutes(10);
+        
+        /// <summary>
+        ///     Amount of time the polling service will wait for a request from an RPC call.
+        ///     Specifically the amount of time to wait for the first byte of the request to arrive.
+        ///
+        ///     This value must never be less than the PollingQueueWaitTimeout, of the client, since this
+        ///     is the timeout of the long poll the polling service makes to the client to get the next request.
+        ///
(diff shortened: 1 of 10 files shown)


nathanwoctopusdeploy (approved):
LGTM

source/Halibut/Diagnostics/HalibutTimeoutsAndLimits.cs
```diff
@@ -51,8 +51,21 @@ public class HalibutTimeoutsAndLimits
...
+        /// 
+        ///     Specifically the amount of time to wait for the first byte of the request to arrive.
         /// </summary>
         public TimeSpan TcpClientReceiveResponseTimeout { get; set; } = TimeSpan.FromMinutes(10);
+        
+        /// <summary>
+        ///     Amount of time the polling service will wait for a request from an RPC call.
+        ///     Specifically the amount of time to wait for the first byte of the request to arrive.
+        ///
+        ///     This value must never be less than the PollingQueueWaitTimeout, of the client, since this
```
sburmanoctopus, line 63: If this is a hard rule, then maybe we should validate this in the 'set' of here and `PollingQueueWaitTimeout`?

LukeButters: I guess we could, keep in mind that the setting on the service must be larger than the setting on the client.

PR #551, opened 2023-11-22
Adding a test to ensure data streams respect TcpClientReceiveResponseTransmissionAfterInitialReadTimeout

sburmanoctopus:
[sc-42848]

# Background

We wanted to make sure data streams respected the new `TcpClientReceiveResponseTransmissionAfterInitialReadTimeout`. Turns out they did. So let's write a test to prove it.

The data streams work because `MessageExchangeStream.ReceiveResponseAsync` calls `ReceiveAsync` using the new timeouts, and `ReceiveAsync` calls `ReadStreamsAsync`. So the timeouts are in place already.

```
public async Task<ResponseMessage> ReceiveResponseAsync(CancellationToken cancellationToken)
{
    // Wait for data to become available using existing timeouts, then once we have data streaming in, use the smaller timeout (so we do not wait as long if an error happens here).
    await stream.WithReadTimeout(
        halibutTimeoutsAndLimits.TcpClientReceiveResponseTimeout,
        async () => await stream.WaitForDataToBeAvailableAsync(cancellationToken));

    return await stream.WithReadTimeout(
        halibutTimeoutsAndLimits.TcpClientReceiveResponseTransmissionAfterInitialReadTimeout,
        async () => await ReceiveAsync<ResponseMessage>(cancellationToken));
}

async Task<T> ReceiveAsync<T>(CancellationToken cancellationToken)
{
    var (result, dataStreams) = await serializer.ReadMessageAsync<T>(stream, cancellationToken);
    await ReadStreamsAsync(dataStreams, cancellationToken);
    log.Write(EventType.Diagnostic, "Received: {0}", result);
    return result;
}
```

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Timeouts/ReceiveResponseTimeoutTests.cs
+++ b/source/Halibut.Tests/Timeouts/ReceiveResponseTimeoutTests.cs
@@ -9,11 +9,12 @@ using Halibut.Tests.Support.TestAttributes;
 using Halibut.Tests.Support.TestCases;
 using Halibut.Tests.TestServices;
 using Halibut.Tests.TestServices.Async;
+using Halibut.Tests.Util;
 using Halibut.TestUtils.Contracts;
 using NUnit.Framework;
 using Octopus.TestPortForwarder;
 
-namespace Halibut.Tests
+namespace Halibut.Tests.Timeouts
 {
     public class ReceiveResponseTimeoutTests : BaseTest
     {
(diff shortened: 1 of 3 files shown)


source/Halibut.Tests/Timeouts/ReceiveResponseTimeoutTests.cs
```diff
@@ -92,5 +93,37 @@ await using (var clientAndService = await clientAndServiceTestCase.CreateTestCas
...
         }
+        
+        [Test]
+        [LatestClientAndLatestServiceTestCases(testNetworkConditions: false, testWebSocket: false)]
+        public async Task WhenRpcExecutionIsWithinReceiveResponseTimeout_ButDataStreamDataIsDelayed_ThenTimeoutShouldOccur(ClientAndServiceTestCase clientAndServiceTestCase)
+        {
+            var halibutTimeoutsAndLimits = new HalibutTimeoutsAndLimits
+            {
+                // Ensure execution time does not trigger timeouts
+                TcpClientReceiveResponseTimeout = TimeSpan.FromHours(1),
```
LukeButters, line 104: You didn't want to use that method that increases all TCP timeouts, so that you can be sure it is not some other timeout at play?

source/Halibut.Tests/Timeouts/ReceiveResponseTimeoutTests.cs
```diff
@@ -92,5 +93,37 @@ await using (var clientAndService = await clientAndServiceTestCase.CreateTestCas
...
+                TcpClientReceiveResponseTransmissionAfterInitialReadTimeout = TimeSpan.FromMilliseconds(100)
```
LukeButters, line 105: Maybe bump this timeout, since it is low enough that on TC it could be the reading of the JSON response that trips it.

LukeButters (approved):

source/Halibut.Tests/Timeouts/ReceiveResponseTimeoutTests.cs
sburmanoctopus: Yeah, good idea. Done!

sburmanoctopus: I did it for the other tests as well.

source/Halibut.Tests/Timeouts/ReceiveResponseTimeoutTests.cs
sburmanoctopus: I bumped it to 1 second. I wasn't sure what number you had in mind.

sburmanoctopus:
@LukeButters , I also ninja'd in a namespace change, as the version in the project name and namespace did not align 😁

PR #556, opened 2023-11-27
Capture the inner exception type for a halibut client exception

nathanwoctopusdeploy:
# Background

This PR capture the inner exception type for a halibut client exception.

Usage of this will be in process, rather than from Tentacle to TentacleClient so there is no issue with exceptions from Tentalce not setting this property. However, it does add a property to Halibut Client Exception that may or may not be set depending on the source of the exception. If the source of the error is In-process it will be set, if the source is out of process, it won't be set.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/CancellationViaClientProxyFixture.cs
+++ b/source/Halibut.Tests/CancellationViaClientProxyFixture.cs
@@ -12,6 +12,7 @@ using Halibut.Tests.Support.TestCases;
 using Halibut.Tests.TestServices.Async;
 using Halibut.TestUtils.Contracts;
 using Halibut.Transport.Protocol;
+using Halibut.Util;
 using NUnit.Framework;
 
 namespace Halibut.Tests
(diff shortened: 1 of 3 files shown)


source/Halibut.Tests/CancellationViaClientProxyFixture.cs
```diff
@@ -41,7 +42,7 @@ await using (var clientAndService = await clientAndServiceTestCase.CreateTestCas
                 tokenSourceToCancel.CancelAfter(TimeSpan.FromMilliseconds(100));
 
                 (await AssertAsync.Throws<Exception>(() => echo.IncrementAsync(halibutRequestOption)))
-                    .And.Should().Match(x => x is ConnectingRequestCancelledException || (x is HalibutClientException && x.As<HalibutClientException>().Message.Contains("The Request was cancelled while Connecting")));
```
LukeButters, line 44: In theory these messages are important for backwards comparability, are we missing tests?

PR #557, opened 2023-11-27
Always make available trace logs

LukeButters:
# Background

We always recorded trace logs to file, this PR changes the tests so that those trace logs are always available.

This means:
* On teamcity if a test times out and never reaches a tear down, we can still see trace logs.
* Locally we can view trace logs.

This test will also print the location of the trace log for easy debugging.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/BaseTest.cs
+++ b/source/Halibut.Tests/BaseTest.cs
@@ -15,15 +15,16 @@ namespace Halibut.Tests
         public CancellationToken CancellationToken { get; private set; }
         public ILogger Logger { get; private set; } = null!;
 
-
         [SetUp]
         public void SetUp()
         {
-            traceLogFileLogger = new TraceLogFileLogger();
+            traceLogFileLogger = new TraceLogFileLogger(SerilogLoggerBuilder.CurrentTestHash());
             Logger = new SerilogLoggerBuilder()
                 .SetTraceLogFileLogger(traceLogFileLogger)
                 .Build()
                 .ForContext(GetType());
+            
+            Logger.Information("Trace log file {LogFile}", traceLogFileLogger.logFilePath);
 
             Logger.Information("Test started");
             cancellationTokenSource = new CancellationTokenSource();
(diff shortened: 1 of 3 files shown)


source/Halibut.Tests/TraceLogFileLogger.cs
```diff
@@ -85,30 +87,14 @@ public bool CopyLogFileToArtifacts()
...
-            try
-            {
-                File.Delete(tempFilePath);
-            }
-            catch
-            {
-                // Best effort clean-up, but we don't want to
-                // fail the test because we couldn't delete this file
-            }
+            cancellationTokenSource.Cancel();
```
LukeButters, line 95: No need to delete the log file.

but we do await the writing of logs, since we really want to know if writing logs doesn't work.

source/Halibut.Tests/TraceLogFileLogger.cs
```diff
@@ -64,16 +64,18 @@ async Task WriteDataToFile()
...
-            writeDataToDiskTask.GetAwaiter().GetResult();
-            cancellationTokenSource.Dispose();
+            var traceLogsDirectory = LogFileDirectory();
+            var fileName = $"{testHash}.tracelog";
+            var logFilePath = Path.Combine(traceLogsDirectory.ToString(), fileName);
+            return logFilePath;
         }
 
-        public bool CopyLogFileToArtifacts()
+        public static DirectoryInfo LogFileDirectory()
```
acodrington, line 77: Maybe this doesn't need to be `public` anymore

source/Halibut.Tests/BaseTest.cs
```diff
@@ -15,15 +15,16 @@ public class BaseTest
...
         public void SetUp()
         {
-            traceLogFileLogger = new TraceLogFileLogger();
+            traceLogFileLogger = new TraceLogFileLogger(SerilogLoggerBuilder.CurrentTestHash());
             Logger = new SerilogLoggerBuilder()
                 .SetTraceLogFileLogger(traceLogFileLogger)
                 .Build()
                 .ForContext(GetType());
+            
+            Logger.Information("Trace log file {LogFile}", traceLogFileLogger.logFilePath);
```
acodrington, line 27: Thanks for this bit, I've got some WIP code to associate the TeamCity artifact with the test run so the trace log file can be accessed in-line in the build log, but there's still some TeamCity black magic I need to figure out to get it to work.

This will help in the meantime!

acodrington (approved):
Nice one, thanks! 👍🏻

source/Halibut.Tests/BaseTest.cs
LukeButters: Ah nice!

I also like this message for local development, since when the test is run the first line is "here is more info if you need it"

acodrington: Yep, it was on my to-do list for that exact reason, but got distracted by the TeamCity black magic and it didn't make it to a PR.

PR #558, opened 2023-11-27
Fix a bug in cooperative cancellation for Polling Services causing a request to time out to quickly

nathanwoctopusdeploy:
# Background

This PR fixes a bug in cooperative cancellation for Polling Requests which cause the transferring request to be cancelled after the  PollingRequestQueueTimeout rather than waiting for the PollingRequestMaximumMessageProcessingTimeout 

[sc-58429]

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Support/LatestClientAndLatestServiceBuilderExtensionMethods.cs
+++ b/source/Halibut.Tests/Support/LatestClientAndLatestServiceBuilderExtensionMethods.cs
@@ -59,13 +59,6 @@ namespace Halibut.Tests.Support
         {
             return builder.WithPollingReconnectRetryPolicy(() => new RetryPolicy(1, TimeSpan.Zero, TimeSpan.Zero));
         }
-
-        public static LatestClientAndLatestServiceBuilder WhenTestingAsyncClient(this LatestClientAndLatestServiceBuilder builder, ClientAndServiceTestCase clientAndServiceTestCase, Action<LatestClientAndLatestServiceBuilder> action)
-        {
-
-            action(builder);
-            return builder;
-        }
         
         public static LatestClientAndLatestServiceBuilder WithConnectionObserverOnTcpServer(this LatestClientAndLatestServiceBuilder builder, IConnectionsObserver connectionsObserver)
         {
--- a/source/Halibut/ServiceModel/PendingRequestQueueAsync.cs
+++ b/source/Halibut/ServiceModel/PendingRequestQueueAsync.cs
@@ -199,7 +199,11 @@ namespace Halibut.ServiceModel
 
                 try
                 {
-                    responseSet = await WaitForResponseToBeSet(request.Destination.PollingRequestQueueTimeout, cancellationToken);
+                    responseSet = await WaitForResponseToBeSet(
+                        request.Destination.PollingRequestQueueTimeout, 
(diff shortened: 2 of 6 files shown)


source/Halibut.Tests/Support/LatestClientAndLatestServiceBuilderExtensionMethods.cs
```diff
@@ -59,13 +59,6 @@ public static LatestClientAndLatestServiceBuilder WithInstantReconnectPollingRet
         {
             return builder.WithPollingReconnectRetryPolicy(() => new RetryPolicy(1, TimeSpan.Zero, TimeSpan.Zero));
         }
-
-        public static LatestClientAndLatestServiceBuilder WhenTestingAsyncClient(this LatestClientAndLatestServiceBuilder builder, ClientAndServiceTestCase clientAndServiceTestCase, Action<LatestClientAndLatestServiceBuilder> action)
```
nathanwoctopusdeploy, line 63: Removed as everything is Async

source/Halibut/Transport/TcpClientManager.cs
```diff
@@ -62,22 +62,25 @@ public void RemoveClient(TcpClient client)
         {
             lock (syncLock)
             {
-                foreach(var thumbprintClientsPair in activeClients)
+                if (activeClients != null)
```
nathanwoctopusdeploy, line 65: This was throwing during debug of some tests

source/Halibut/ServiceModel/PendingRequestQueueAsync.cs
```diff
@@ -283,16 +291,24 @@ async Task<bool> WaitForResponseToBeSet(TimeSpan timeout, CancellationToken canc
...
                 {
-                    // Cancel the Request to force cancellation to the socket if it is currently being transferred 
-                    CancelRequest();
-
-                    if (timeoutCancellationTokenSource.IsCancellationRequested)
-                    {
-                        return false;
-                    }
-
                     using (await transferLock.LockAsync(CancellationToken.None))
```
nathanwoctopusdeploy, line 294: Ensure there is no race condition allowing a just cancelled request to be dequeued

source/Halibut/Transport/Protocol/StreamTimeoutExtensionMethods.cs
```diff
@@ -26,8 +26,17 @@ public static async Task WithTimeout(this Stream stream, SendReceiveTimeout time
...
-                stream.ReadTimeout = currentReadTimeout;
-                stream.WriteTimeout = currentWriteTimeout;
+                try
+                {
+                    stream.ReadTimeout = currentReadTimeout;
+                    stream.WriteTimeout = currentWriteTimeout;
+                }
+                catch (Exception e)
+                {
+#pragma warning disable CA2200
```
nathanwoctopusdeploy, line 36: WIP

source/Halibut/Transport/TcpClientManager.cs
sburmanoctopus: Although, why is it null? It looks like it's only null because it was disposed. 

Is it an issue that our tests are causing us to access an object that has been disposed? 🤔 

If we're cool with that, an alternative is to set `activeClients` to a new empty object if we didn't want to do a null check here. 

But it still feels a little sus that we're accessing a disposed object.

sburmanoctopus (approved):
LGTM

PR #562, opened 2023-11-29
Code analysis fixes

sburmanoctopus:
# Background

For years, no one has dealt with the warnings that have been accumulating as .NET code analysis have matured.

# Results

## Before

There were hundreds of warnings that were being ignored.

## After

These have all been addressed, and each project has had "Treat warnings as errors" turned on.

There was a lot to get through, so in many cases, the simplest solution was often taken (e.g. using "!" or "?" etc.)

But this is **more about helping future developers**, and **less about fixing all the issues that already exist**.

Some places it was easier to suppress the warning than solve it, as the difference between .NET4.8 and .NET6.0 made it harder to solve.

With models, I tried to reflect whether a property was legitimately nullable or not.

### AssertAsync
We had to add an overload that was not async (i.e. sync), which made this class name not make sense anymore. So we renamed to `AssertException`.

Also, whenever we used the async version, and gave it a task from a `Task.Run`, it would warn about this. So we made the overload that just takes the task itself, and suppress the warning in that overload.

### .editorconfig
This was added so we could disable VSTHRD200. This was a nightmare in the test project, as it wanted all our async tests to end with the word 'Async'. But this isn't a rule we are respecting in Halibut itself, so it was disabled.

# How to review this PR
There is a lot here.

To help break it up, each commit attempts to achieve something specific towards the final goal.

Please review each commit (or the whole thing if that works better for you).

As you review, please try to see if I have converted something incorrectly.

Otherwise, Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.TestProxy/Halibut.TestProxy.csproj
+++ b/source/Halibut.TestProxy/Halibut.TestProxy.csproj
@@ -4,6 +4,7 @@
     <Nullable>enable</Nullable>
     <LangVersion>9.0</LangVersion>
     <RootNamespace>Halibut.TestProxy</RootNamespace>
+	  <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
 
   <PropertyGroup Condition="!$([MSBuild]::IsOSUnixLike())">
--- a/source/Halibut.TestProxy/HttpProxyService.cs
+++ b/source/Halibut.TestProxy/HttpProxyService.cs
@@ -104,7 +104,7 @@ namespace Halibut.TestProxy
                         using var socketCancellationTokenSource = new CancellationTokenSource();
                         using (cancellationToken.Register(() => Try.CatchingError(() => socketCancellationTokenSource.Cancel(), _ => { })))
                         {
-                            var cancelTask = socketCancellationTokenSource.Token.AsTask<TcpClient?>();
+                            var cancelTask = socketCancellationTokenSource.Token.AsTask<TcpClient>();
                             var actionTask = Task.Run(async () =>
                             {
                                 using var _ = cancellationToken.Register(() => Try.CatchingError(() =>
@@ -188,7 +188,7 @@ namespace Halibut.TestProxy
                 try
                 {
                     logger.LogInformation("Creating Proxy Connection Forwarder");
-                    await CreateProxyConnectionAndForward(client, connectionRequest, cancellationToken);
+                    CreateProxyConnectionAndForward(client, connectionRequest, cancellationToken);
                 }
                 catch (Exception ex)
                 {
@@ -219,10 +219,10 @@ namespace Halibut.TestProxy
             }
         }
 
-        async Task CreateProxyConnectionAndForward(TcpClient client, HttpProxyConnectionRequest connectionRequest, CancellationToken cancellationToken)
+        void CreateProxyConnectionAndForward(TcpClient client, HttpProxyConnectionRequest connectionRequest, CancellationToken cancellationToken)
         {
             var proxyConnection = proxyConnectionService.GetProxyConnection(connectionRequest.Endpoint);
-            await proxyConnection.Connect(client, cancellationToken);
+            proxyConnection.Connect(client, cancellationToken);
         }
 
         async Task SendConnectResponse(ProxyEndpoint endpoint, string httpVersion, int responseCode, string message, PipeWriter writer, CancellationToken cancellationToken)
--- a/source/Halibut.TestProxy/ProxyConnection.cs
+++ b/source/Halibut.TestProxy/ProxyConnection.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Net;
 using System.Net.Sockets;
 using System.Threading;
@@ -11,7 +10,7 @@ namespace Halibut.TestProxy
 {
     interface IProxyConnection : IDisposable
     {
-        Task Connect(TcpClient source, CancellationToken cancellationToken);
+        void Connect(TcpClient source, CancellationToken cancellationToken);
     }
 
     class ProxyConnection : IProxyConnection
@@ -26,7 +25,7 @@ namespace Halibut.TestProxy
             this.logger = logger;
         }
 
-        public async Task Connect(TcpClient source, CancellationToken cancellationToken)
+        public void Connect(TcpClient source, CancellationToken cancellationToken)
         {
             if (source.Client.RemoteEndPoint is not IPEndPoint sourceRemoteEndpoint)
             {
--- a/source/Halibut.TestUtils.CompatBinary.Base/BackwardsCompatProgramBase.cs
+++ b/source/Halibut.TestUtils.CompatBinary.Base/BackwardsCompatProgramBase.cs
@@ -42,7 +42,7 @@ namespace Halibut.TestUtils.SampleProgram.Base
             var serviceConnectionType = SettingsHelper.GetServiceConnectionType();
             var octopusThumbprint = SettingsHelper.GetClientThumbprint();
 
-            string addressToPoll = null;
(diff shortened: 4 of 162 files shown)


source/Halibut.TestProxy/Halibut.TestProxy.csproj
```diff
@@ -4,6 +4,7 @@
     <Nullable>enable</Nullable>
     <LangVersion>9.0</LangVersion>
     <RootNamespace>Halibut.TestProxy</RootNamespace>
+	  <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
LukeButters, line 7: Nice

source/Halibut.TestProxy/HttpProxyService.cs
```diff
@@ -219,10 +219,10 @@ async Task HandleProxyRequest(TcpClient client, CancellationToken cancellationTo
             }
         }
 
-        async Task CreateProxyConnectionAndForward(TcpClient client, HttpProxyConnectionRequest connectionRequest, CancellationToken cancellationToken)
+        void CreateProxyConnectionAndForward(TcpClient client, HttpProxyConnectionRequest connectionRequest, CancellationToken cancellationToken)
         {
             var proxyConnection = proxyConnectionService.GetProxyConnection(connectionRequest.Endpoint);
-            await proxyConnection.Connect(client, cancellationToken);
+            proxyConnection.Connect(client, cancellationToken);
```
LukeButters, line 225: We don't want this to be async?

source/Halibut.TestProxy/ProxyConnection.cs
```diff
@@ -26,7 +25,7 @@ public ProxyConnection(ProxyEndpoint destinationEndpoint, ILogger<ProxyConnectio
             this.logger = logger;
         }
 
-        public async Task Connect(TcpClient source, CancellationToken cancellationToken)
+        public void Connect(TcpClient source, CancellationToken cancellationToken)
         {
```
LukeButters, line 29: nit: would adding `await Task.CompletedTask;` have made the warning go away?

source/Halibut.TestProxy/HttpProxyService.cs
sburmanoctopus: Maybe. But the reason I removed it completely was because:
- It didn't actually do anything that was async
- Therefore it's misleading, as you expect it to run on another thread, but in reality, .NET is clever enough to run the code in its original thread until is reaches something it has to await

So I thought it more clear to call it what it was (not async)

source/Halibut.TestProxy/ProxyConnection.cs
sburmanoctopus: It would have. But I didn't for the reasons outlined in [this comment](https://github.com/OctopusDeploy/Halibut/pull/562#discussion_r1409993158)

source/Halibut.TestUtils.CompatBinary.Base/SettingsHelper.cs
```diff
@@ -39,7 +39,7 @@ int AsInt(string s)
 
         public static string GetSetting(string name)
         {
-            return Environment.GetEnvironmentVariable(name);
+            return Environment.GetEnvironmentVariable(name) ?? string.Empty;
```
LukeButters, line 42: nit: null might be more meaningful when the setting is not set rather than empty.

source/Halibut.Tests/Support/AssertAsync.cs
```diff
@@ -1,15 +0,0 @@
-﻿using System;
-using System.Threading.Tasks;
-using FluentAssertions;
-using FluentAssertions.Specialized;
-
-namespace Halibut.Tests.Support
-{
-    public static class AssertAsync
-    {
-        public static async Task<ExceptionAssertions<T>> Throws<T>(this Func<Task> task, string because = "") where T : Exception
```
LukeButters, line 10: Does this mean we can no longer do:
```
await someTask.Throws<Exception>();
```

source/Halibut.Tests/Support/BackwardsCompatibility/HalibutTestBinaryPath.cs
```diff
@@ -10,7 +10,7 @@ public string BinPath(string version)
         {
             var onDiskVersion = version.Replace(".", "_");
             var assemblyDir = new DirectoryInfo(Path.GetDirectoryName(typeof(HalibutTestBinaryRunner).Assembly.Location)!);
-            var upAt = assemblyDir.Parent.Parent.Parent.Parent;
+            var upAt = assemblyDir.Parent!.Parent!.Parent!.Parent!;
```
LukeButters, line 13: bang bang bang

source/Halibut.Tests/Support/BackwardsCompatibility/PreviousClientVersionAndLatestServiceBuilder.cs
```diff
@@ -30,17 +30,14 @@ public class PreviousClientVersionAndLatestServiceBuilder: IClientAndServiceBuil
     {
         readonly ServiceConnectionType serviceConnectionType;
         
-        ServiceFactoryBuilder serviceFactoryBuilder = new();
```
LukeButters, line 33: I guess this was all never used.

source/Halibut.TestUtils.CompatBinary.Base/BackwardsCompatProgramBase.cs
```diff
@@ -42,7 +42,7 @@ static async Task RunExternalService(CancellationToken cancellationToken)
             var serviceConnectionType = SettingsHelper.GetServiceConnectionType();
             var octopusThumbprint = SettingsHelper.GetClientThumbprint();
 
-            string addressToPoll = null;
```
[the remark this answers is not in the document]
sburmanoctopus: Yeah, agreed. It's not perfect, but does the job. `addressToPoll` will be assigned if it's used. We can always circle back on this later if we want.

source/Halibut.TestUtils.CompatBinary.Base/SettingsHelper.cs
sburmanoctopus: The motivation was more that everywhere that uses `GetSetting` expects there to be a value. I want to keep that assumption true.

An alternative, could be to throw a `NullReferenceException` (or something else) instead? That way failures are loud and proud?

source/Halibut.Tests/Util/DataStreamExtensionMethods.cs
```diff
@@ -10,17 +10,18 @@ public static class DataStreamExtensionMethods
...
             stream.Receiver().ReadAsync(async (s, ct) =>
                     {
-                        await Task.CompletedTask;
                         using (var reader = new StreamReader(s))
                         {
                             result = await reader.ReadToEndAsync();
                         }
                     },
                     CancellationToken.None)
                 .GetAwaiter().GetResult();
```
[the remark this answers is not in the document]
sburmanoctopus: Agreed. I _almost_ did it this time, but saw how big the splash damage would be and ran away 😁 

I'll put it on the list for later.

LukeButters (approved):
LGTM

Ideally though it would make sense to have a PR with just the changes to the `Halibut` repo which shows the changes in Halibut work with the existing tests. Followed by another PR which change the tests

source/Halibut.TestUtils.CompatBinary.Base/SettingsHelper.cs
LukeButters: leave as is

source/Halibut.TestUtils.CompatBinary.Base/BackwardsCompatProgramBase.cs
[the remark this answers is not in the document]
LukeButters: agreed

PR #563, opened 2023-11-30
Stop throwing an exception in the finally of WithTimeouts

nathanwoctopusdeploy:
# Background

Stop throwing an exception in the finally of WithTimeouts when the Stream is in an error state.

[sc-65700]

This has been run through a Tentacle build here https://github.com/OctopusDeploy/OctopusTentacle/pull/711

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Transport/Protocol/StreamTimeoutExtensionMethods.cs
+++ b/source/Halibut/Transport/Protocol/StreamTimeoutExtensionMethods.cs
@@ -18,6 +18,7 @@ namespace Halibut.Transport.Protocol
 
             var currentReadTimeout = stream.ReadTimeout;
             var currentWriteTimeout = stream.WriteTimeout;
+            var timeoutsReverted = false;
 
             try
             {
@@ -26,8 +27,20 @@ namespace Halibut.Transport.Protocol
             }
             finally
             {
-                stream.ReadTimeout = currentReadTimeout;
-                stream.WriteTimeout = currentWriteTimeout;
+                try
+                {
+                    stream.ReadTimeout = currentReadTimeout;
+                    stream.WriteTimeout = currentWriteTimeout;
+                    timeoutsReverted = true;
+                }
+                catch
+                {
+                }
+            }
+
+            if (!timeoutsReverted)
+            {
+                throw new InvalidOperationException("Could not revert the Timeouts. This should not happen.");
             }
         }
 
@@ -40,17 +53,33 @@ namespace Halibut.Transport.Protocol
 
             var currentReadTimeout = stream.ReadTimeout;
             var currentWriteTimeout = stream.WriteTimeout;
+            var timeoutsReverted = false;
+            T result;
 
             try
             {
                 stream.SetReadAndWriteTimeouts(timeout);
-                return await func();
+                result = await func();
             }
             finally
             {
-                stream.ReadTimeout = currentReadTimeout;
-                stream.WriteTimeout = currentWriteTimeout;
+                try
+                {
+                    stream.ReadTimeout = currentReadTimeout;
+                    stream.WriteTimeout = currentWriteTimeout;
+                    timeoutsReverted = true;
+                }
+                catch
+                {
+                }
             }
+
+            if (!timeoutsReverted)
+            {
+                throw new InvalidOperationException("Could not revert the Timeouts. This should not happen.");
+            }
+
+            return result;
         }
 
         public static async Task WithReadTimeout(this Stream stream, TimeSpan timeout, Func<Task> func)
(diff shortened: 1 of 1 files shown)


source/Halibut/Transport/Protocol/StreamTimeoutExtensionMethods.cs
```diff
@@ -26,8 +27,20 @@ public static async Task WithTimeout(this Stream stream, SendReceiveTimeout? tim
             }
             finally
             {
-                stream.ReadTimeout = currentReadTimeout;
-                stream.WriteTimeout = currentWriteTimeout;
+                try
+                {
+                    stream.ReadTimeout = currentReadTimeout;
```
sburmanoctopus, line 32: 🤔 Just wondering, do other native .NET streams throw an exception when adjusting the Read/Write timeouts?

If not, then I wonder if an alternative would be to instead change the `NetworkTimeoutStream` to not throw an exception when changing `ReadTimeout` or `WriteTimeout`.

The current solution will be more safe, as it guarantees any read/write timeout alteration won't get in the way. But I couldn't help but wonder if we were guarding against something that shouldn't happen (and we get read/write timeout properties in `NetworkTimeoutStream` wrong instead).

source/Halibut/Transport/Protocol/StreamTimeoutExtensionMethods.cs
```diff
@@ -26,8 +27,20 @@ public static async Task WithTimeout(this Stream stream, SendReceiveTimeout? tim
...
+                    stream.WriteTimeout = currentWriteTimeout;
+                    timeoutsReverted = true;
+                }
+                catch
+                {
+                }
+            }
+
+            if (!timeoutsReverted)
```
sburmanoctopus, line 41: Is `timeoutsReverted` what we want? 

If `await func()` succeeds, but Read/Write timeout throw an exception, then `timeoutsReverted` will be false, and we will throw an `InvalidOperationException` here. But I thought the point of this PR was to stop throwing an exception in the finally. I.e., aren't we still throwing an exception, just a different one?

I'm likely not understanding something...

source/Halibut/Transport/Protocol/StreamTimeoutExtensionMethods.cs
nathanwoctopusdeploy: > If not, then I wonder if an alternative would be to instead change the NetworkTimeoutStream to not throw an exception when changing ReadTimeout or WriteTimeout.

It's up to the underlying stream and if it's a network stream, possibly the OS? I think they do throw if the stream is not usable, and you throw an unrelated, side effect error, hence why we wrapped it and rethrow the real error.

source/Halibut/Transport/Protocol/StreamTimeoutExtensionMethods.cs
nathanwoctopusdeploy: > This should not happen.

This was a fail safe to make sure it is working as I expected e.g. we throw so it doesn't matter that the timeouts were not reverted. 

I didn't want to swallow errors setting the timeouts when it was a real issue, e.g. the actual call has not thrown, but for some reason the set timeouts throws. That should never happen.

nathanwoctopusdeploy: > If await func() succeeds, but Read/Write timeout throw an exception
Should this happen? Is it valid? Will it leave us in a weird state if it did happen

source/Halibut/Transport/Protocol/StreamTimeoutExtensionMethods.cs
nathanwoctopusdeploy: If `NetworkTimeoutStream` doesn't throw, then do we risk getting into a situation where we swallow real errors and end up not correctly changing timeouts, but we think we have.

nathanwoctopusdeploy:
> I'd like to understand the question on `timeoutsReverted` and why we want to throw an exception first before approving.

There may be a better approach to this. Happy to pair and revisit. The general idea was don't set the timeouts when we can't as we are in error but don't hide errors and allow this to be an area that could cause bugs.

source/Halibut/Transport/Protocol/StreamTimeoutExtensionMethods.cs
sburmanoctopus: Ah, I see. The subtilty is that if we set the read/write timeout, and it throws a timeout exception, then this only really happens because it would have already thrown an exception with `await func()`.

Yep, that makes sense 🙂

PR #565, opened 2023-11-30
Separate Latest client and Latest service

sburmanoctopus:
# Background

Our `LatestClientAndLatestServiceBuilder` tangles the concept of client and service setup into one big class, making it hard to determine the boundaries. 

# Results

## Before
It would allows the developer to setup a test case without a service, or without a client. This resulted in the `ClientAndService` object having nullable `Client` and `Service` properties. Contractually, this makes it uncertain if these properties have values.

It also meant it was harder to use just the client or service by itself.

### No More NoService
The `NoService` property is now replaced with a builder for specifically getting a client only.

Despite `NoService` being on the `IClientAndServiceBuilder`, all test cases except 1 (details in code comment) only ever used it on the latest client and latest service.

So we only implemented a client only version for the latest client.

### Creating Clients When No Service Exists
When we create a "client only" client, we often want to actually attempt an RPC. Therefore we have to `CreateAsyncClient`. But this expects an endpoint (as we are trying to reach a service). But if we have not created a service, then what is the URI we should be attempting to reach?

Therefore, we added the `CreateClientWithoutService` method, which creates a real but unattached endpoint.

### Port Forwarding in IClientAndService
We removed the `PortForwarder` property from the `IClientAndService`, as there was already a partial implementation to get this via the builder.

Like with `Client` and `Service`, means it no longer is 'nullable'.

In hindsight, this enhancement should be a separate PR. But at the time it was thought to be needed to finish this feature. It wasn't until near the end where it was clear that it could have been separate sorry 😞. Due to the time constraint of Sharpening, we kept it as one PR.

### Port Forwarding - Client or Service?
The port that needs forwarding very much depends on whether we are listening or polling. If we are listening, then we need to forward the port on the service side. If we are polling, then we need to forward the port on the client side.

As a result, if we are using a port forwarder for a client and service, the builder that builds it will change depending on the type.

### TCPListenerWhichKillsNewConnections
The original `LatestClientAndLatestServiceBuilder` used to create a `TCPListenerWhichKillsNewConnections` when there was no service.

We are assuming that this was done to ensure we chose a port that was free, and would not accept connections.

So to keep this behaviour, it was reworked into the `LatestClient` when using `CreateClientWithoutService`, as attempting to connect to a service that does not exist is the way this would get exercised.

# How to review this PR
The meat of the change is within `LatestClientBuilder`, `LatestServiceBuilder`, and the old `LatestClientAndLatestServiceBuilder`. So I would recommend starting there.

Please ensure that the split is correct, and that I didn't overlook anything (or get it wrong) when I split them apart.

Quality :heavy_check_mark:

# Pre-requisites


source/Halibut.Tests/FailureModesFixture.cs
```diff
@@ -28,21 +28,17 @@ static DelegateServiceFactory GetDelegateServiceFactory()
 
         [Test]
         [LatestClientAndLatestServiceTestCases(testNetworkConditions: false, testListening: false)]
-        [LatestClientAndPreviousServiceVersionsTestCases(testNetworkConditions: false, testListening: false)]
```
sburmanoctopus, line 31: As we were making the latest client only version, we found this test case, as it was going to be an issue.

But looking closer, `LatestClientAndPreviousServiceVersionsTestCases` is still just the latest client (when you have 'no service'). 

So it looked like it was an unrequired double up on effectively the same test case. 

That is why it was removed.

source/Halibut.Tests/Support/LatestClientAndLatestServiceBuilder.cs
```diff
@@ -102,127 +80,71 @@ public static LatestClientAndLatestServiceBuilder ForServiceConnectionType(Servi
...
-            return WithPortForwarding(port => PortForwarderUtil.ForwardingToLocalPort(port).Build());
+            return this.WithPortForwarding(out portForwarder, portForwarderFactory);
         }
 
-        IClientAndServiceBuilder IClientAndServiceBuilder.WithPortForwarding(Func<int, PortForwarder> portForwarderFactory)
+        public LatestClientAndLatestServiceBuilder WithPortForwarding(out Reference<PortForwarder> portForwarder, Func<int, PortForwarder> portForwarderFactory)
         {
-            return this.WithPortForwarding(portForwarderFactory);
-        }
+            // Based on ServiceConnectionType, the port forwarder will be created in either clientBuilder or serviceBuilder.
```
sburmanoctopus, line 142: I don't like this, so please let me know if there is a better way.

The issue is that we do not know which builder will end up making the port forwarder. So we have to register references for both, and then see later which one was the one that made it 😑

source/Halibut.Tests/Support/LatestClientAndLatestServiceBuilder.cs
```diff
@@ -356,301 +277,82 @@ async Task<IClientAndService> IClientAndServiceBuilder.Build(CancellationToken c
...
-                return pendingRequestQueueFactory(octopusLogFactory);
+                clientBuilder.WithProxyDetails(httpProxyDetails);
+                serviceBuilder.WithProxyDetails(httpProxyDetails);
             }
 
-            var pendingRequestQueueFactoryBuilder = new PendingRequestQueueFactoryBuilder(octopusLogFactory);
-
-            if (this.pendingRequestQueueFactoryBuilder != null)
+            var client = await clientBuilder.Build(cancellationToken);
+            if (client.ListeningUri is not null)
```
sburmanoctopus, line 292: If we are testing "Polling", then the client is listening, and the service needs to know which endpoint to "Poll"

source/Halibut.Tests/Support/LatestClientAndLatestServiceBuilder.cs
```diff
@@ -356,301 +277,82 @@ async Task<IClientAndService> IClientAndServiceBuilder.Build(CancellationToken c
...
             {
-                return new TestContextLogCreator("Client", halibutLogLevel).ToCachingLogFactory();
-            }
-
-
-
-            return new AggregateLogWriterLogCreator(
-                    new TestContextLogCreator("Client", halibutLogLevel),
-                s =>
+                // Only one of these will have created a port forwarder. Use that one.
```
sburmanoctopus, line 301: Again, this is yuck. So am keen to think of better ways

source/Halibut.Tests/FailureModesFixture.cs
LukeButters: lol yep

LukeButters (approved):

PR #569, opened 2023-11-30
Tests to ensure the type of exceptions thrown during RPC calls for connecting, timeouts and cancellations errors do not change

nathanwoctopusdeploy:
# Background

Tests to cover the type of exception that occurs during an RPC call for connecting, timeout and cancellation scenarios.

The behaviour and the type of exception thrown are important to the caller.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/ExceptionContractFixture.cs
+++ b/source/Halibut.Tests/ExceptionContractFixture.cs
@@ -1,0 +1,290 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Halibut.Tests.Support;
+using Halibut.Tests.Support.PendingRequestQueueFactories;
+using Halibut.Tests.Support.TestAttributes;
+using Halibut.Tests.Support.TestCases;
+using Halibut.Tests.TestServices;
+using Halibut.Tests.TestServices.Async;
+using Halibut.TestUtils.Contracts;
+using NUnit.Framework;
+
+namespace Halibut.Tests
+{
+    /// <summary>
+    /// The type of exception that is throw by Halibut is important for the caller to be able to determine what went wrong and how to handle it.
+    /// These tests ensure the contract is maintained in Halibut and that the exception does not change and have un-intended consequences for the caller.
+    /// </summary>
+    public class ExceptionContractFixture : BaseTest
+    {
+        [Test]
+        [LatestClientAndLatestServiceTestCases(testNetworkConditions: false, testListening: false)]
+        public async Task WhenThePollingRequestQueueTimeoutIsReached_AHalibutClientExceptionShouldBeThrown(ClientAndServiceTestCase clientAndServiceTestCase)
+        {
+            var halibutTimeoutsAndLimits = new HalibutTimeoutsAndLimitsForTestsBuilder().Build();
+            halibutTimeoutsAndLimits.PollingRequestQueueTimeout = TimeSpan.FromSeconds(5);
+
+            await using var clientOnly = await clientAndServiceTestCase.CreateClientOnlyTestCaseBuilder()
+                .AsLatestClientBuilder()
+                .WithHalibutTimeoutsAndLimits(halibutTimeoutsAndLimits)
+                .Build(CancellationToken);
+
+            var client = clientOnly.CreateClientWithoutService<IEchoService, IAsyncClientEchoServiceWithOptions>();
+
+            (await AssertException.Throws<HalibutClientException>(async () => await client.SayHelloAsync("Hello", new(CancellationToken, CancellationToken.None))))
+                .And.Message.Should().Contain("A request was sent to a polling endpoint, but the polling endpoint did not collect the request within the allowed time (00:00:05), so the request timed out.");
+        }
+
+        [Test]
+        [LatestClientAndLatestServiceTestCases(testNetworkConditions: false, testListening: false)]
+        public async Task WhenThePollingRequestMaximumMessageProcessingTimeoutIsReached_AHalibutClientExceptionShouldBeThrown(ClientAndServiceTestCase clientAndServiceTestCase)
+        {
+            var halibutTimeoutsAndLimits = new HalibutTimeoutsAndLimitsForTestsBuilder().Build();
+            halibutTimeoutsAndLimits.PollingRequestQueueTimeout = TimeSpan.FromSeconds(5);
+            halibutTimeoutsAndLimits.PollingRequestMaximumMessageProcessingTimeout = TimeSpan.FromSeconds(6);
+
+            var waitSemaphore = new SemaphoreSlim(0, 1);
+
+            await using var clientAndService = await clientAndServiceTestCase.CreateTestCaseBuilder()
+                .AsLatestClientAndLatestServiceBuilder()
+                .WithDoSomeActionService(() => waitSemaphore.Wait(CancellationToken))
+                .WithHalibutTimeoutsAndLimits(halibutTimeoutsAndLimits)
+                .Build(CancellationToken);
+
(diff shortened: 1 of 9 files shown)


source/Halibut.Tests/ExceptionContractFixture.cs
```diff
@@ -0,0 +1,228 @@
...
+
+            var cancellationTokenSource = new CancellationTokenSource();
+            
+            var cancellationTask = Task.Run(async () =>
+            {
+                await executingSemaphore.WaitAsync(CancellationToken);
+                cancellationTokenSource.Cancel();
+            });
+
+            // CONTRACT BUG - This will be interpreted as a Transient Error in Server and cause the incorrect behaviour on cancellation
```
nathanwoctopusdeploy, line 219: This was why Cooperative cancellation had unexpected side effects in Server 😭 

@sburmanoctopus it was a pre-existing bug from adding cancellation to the socket!!

source/Halibut.Tests/Support/PendingRequestQueueFactories/CancelWhenRequestDequeuedPendingRequestQueueFactory.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Threading.Tasks;
+using Halibut.ServiceModel;
+using Halibut.Transport.Protocol;
+
+namespace Halibut.Tests.Support.PendingRequestQueueFactories
+{
+    /// <summary>
+    /// CancelWhenRequestDequeuedPendingRequestQueueFactory cancels the cancellation token source when a request is queued
+    /// </summary>
+    public class CancelWhenRequestDequeuedPendingRequestQueueFactory : IPendingRequestQueueFactory
```
nathanwoctopusdeploy, line 12: Copy paste move

source/Halibut.Tests/Support/PendingRequestQueueFactories/CancelWhenRequestQueuedPendingRequestQueueFactory.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.Threading.Tasks;
+using Halibut.ServiceModel;
+using Halibut.Transport.Protocol;
+
+namespace Halibut.Tests.Support.PendingRequestQueueFactories
+{
+    /// <summary>
+    /// CancelWhenRequestQueuedPendingRequestQueueFactory cancels the cancellation token source when a request is queued
+    /// </summary>
+    public class CancelWhenRequestQueuedPendingRequestQueueFactory : IPendingRequestQueueFactory
```
nathanwoctopusdeploy, line 12: Copy paste move

source/Halibut.Tests/Support/LatestClientAndLatestServiceBuilder.cs
```diff
@@ -481,12 +484,16 @@ public async Task<ClientAndService> Build(CancellationToken cancellationToken)
                     service.Trust(serviceTrustsThumbprint);
                     listenPort = service.Listen();
                 }
-                else
+                else if(createTcpKillingDummyServiceWhenListeningService)
```
nathanwoctopusdeploy, line 487: By defaulting to a TCP killing service it was having unintended side effects for a test I was trying to adde.g. I really don't want anything to be there.

nathanwoctopusdeploy: I've left the default behavour so the builder provides a tcp killing service, but it might make sense for this to always be an opt-in. Future PR though!!

source/Halibut/Transport/SecureListeningClient.cs
```diff
@@ -128,7 +128,7 @@ public async Task ExecuteTransactionAsync(ExchangeActionAsync protocolHandler, R
                     // against all connections in the pool being bad
                     if (i == 1)
                     {
-                        await connectionManager.ClearPooledConnectionsAsync(ServiceEndpoint, log, requestCancellationTokens.InProgressRequestCancellationToken);
+                        await connectionManager.ClearPooledConnectionsAsync(ServiceEndpoint, log, requestCancellationTokens.LinkedCancellationToken);
```
sburmanoctopus, line 131: What was the motivation for this change?

source/Halibut.Tests/ExceptionContractFixture.cs
```diff
@@ -0,0 +1,290 @@
...
+        [Test]
+        [LatestClientAndLatestServiceTestCases(testNetworkConditions: false, testPolling: false, testWebSocket: false)]
+        public async Task WhenTheListeningRequestIsCancelledImmediatelyWhileTryingToConnect_AHalibutClientExceptionShouldBeThrown(ClientAndServiceTestCase clientAndServiceTestCase)
+        {
+            var halibutTimeoutsAndLimits = new HalibutTimeoutsAndLimitsForTestsBuilder().Build();
+            halibutTimeoutsAndLimits.TcpClientConnectTimeout = TimeSpan.FromSeconds(5);
+            halibutTimeoutsAndLimits.ConnectionErrorRetryTimeout = TimeSpan.FromSeconds(600);
+            halibutTimeoutsAndLimits.RetryCountLimit = int.MaxValue;
+
+            var connectionsObserver = new TestConnectionsObserver();
```
sburmanoctopus, line 175: What's the observer used for?

source/Halibut.Tests/ExceptionContractFixture.cs
```diff
@@ -0,0 +1,290 @@
...
+            
+            portForwarder.Value.EnterPauseNewAndExistingConnectionsMode();
+            var client = clientAndService.CreateAsyncClient<IEchoService, IAsyncClientEchoServiceWithOptions>();
+
+            var cancellationTokenSource = new CancellationTokenSource(TimeSpan.FromSeconds(5));
+
+#if NETFRAMEWORK
+            // net48 does not support cancellation of the request as the DeflateStream ends up using Begin and End methods which don't get passed the cancellation token
+            // This results in a different exception message / a different code path executing. This highlights an inconsistency in the response from 
+            // Halibut due to the way the code is written, so leaving this here as it should really be consistent.
```
sburmanoctopus, line 246: Is this something we need to address?

sburmanoctopus (approved):
LGTM

source/Halibut/Transport/SecureListeningClient.cs
nathanwoctopusdeploy: Good question, I think it was a hangover from combining the tokens in my head and this just looked wrong. I'll revert though as it's not the intent to do this in this PR!!

source/Halibut.Tests/ExceptionContractFixture.cs
nathanwoctopusdeploy: Occupying a small amount of memory 😉  Good spot 🙇

source/Halibut.Tests/ExceptionContractFixture.cs
nathanwoctopusdeploy: It's a known issue and currently, only TentacleClient will cancel requests and that is always and can only be net6.0+ so this is "OK". Support for net48 should hopefully be on the way out as well so we probably won't ever need to fix this. If we do we can.

source/Halibut.Tests/ExceptionContractFixture.cs
nathanwoctopusdeploy: Removed

source/Halibut/Transport/SecureListeningClient.cs
nathanwoctopusdeploy: Removed

PR #570, opened 2023-12-01
Remove proxy from client and service

sburmanoctopus:
# Background

If you want to get access to the proxy in your test (if you added one), then you have to use a nullable property on the "client and service".

This contract is limiting, as it is not clear when this nullable property will actually contain a value or not.

# Results

## Before

The proxy property will only include a value if you have called `WithProxy`

## After

We make `WithProxy` optionally take an 'out' parameter so that the caller will know for sure that they have the correct reference to the proxy.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


LukeButters (approved):

source/Halibut.Tests/PollingServiceTimeoutsFixture.cs
```diff
@@ -22,14 +22,12 @@ public async Task WhenThePollingRequestQueueTimeoutIsReached_TheRequestShouldTim
             var halibutTimeoutsAndLimits = new HalibutTimeoutsAndLimitsForTestsBuilder().Build();
             halibutTimeoutsAndLimits.PollingRequestQueueTimeout = TimeSpan.FromSeconds(5);
 
-            await using (var clientAndService = await clientAndServiceTestCase.CreateTestCaseBuilder()
-                       .As<LatestClientAndLatestServiceBuilder>()
-                       .NoService()
-                       .WithStandardServices()
-                       .WithHalibutTimeoutsAndLimits(halibutTimeoutsAndLimits)
-                       .Build(CancellationToken))
+            await using (var clientAndService = await clientAndServiceTestCase.CreateClientOnlyTestCaseBuilder()
```
sburmanoctopus, line 25: This was a fix to a poison merge from the previous PR.

source/Halibut.Tests/PollingServiceTimeoutsFixture.cs
```diff
@@ -22,14 +22,12 @@ public async Task WhenThePollingRequestQueueTimeoutIsReached_TheRequestShouldTim
             var halibutTimeoutsAndLimits = new HalibutTimeoutsAndLimitsForTestsBuilder().Build();
             halibutTimeoutsAndLimits.PollingRequestQueueTimeout = TimeSpan.FromSeconds(5);
 
-            await using (var clientAndService = await clientAndServiceTestCase.CreateTestCaseBuilder()
-                       .As<LatestClientAndLatestServiceBuilder>()
-                       .NoService()
-                       .WithStandardServices()
-                       .WithHalibutTimeoutsAndLimits(halibutTimeoutsAndLimits)
-                       .Build(CancellationToken))
+            await using (var clientOnly = await clientAndServiceTestCase.CreateClientOnlyTestCaseBuilder()
```
sburmanoctopus, line 25: This was the fix to a poison merge from the previous PR.

PR #576, opened 2023-12-05
Co-operatively cancel all RPC Calls

nathanwoctopusdeploy:
This is a combination of the reverted PRs https://github.com/OctopusDeploy/Halibut/pull/525 and https://github.com/OctopusDeploy/Halibut/pull/558 combined and rebased

# Background

[sc-58429]

Before Async Halibut was introduced, Halibut only supported cancellation of an RPC call when the request was in the connecting state, queued for a Polling Request and the TCP Connection being established for a Listening Request. Once the RPC was in flight, dequeued for a Polling Request, and the connection established for a Listening Request, cancellation was not supported.

As the cancellation of an in-flight request was not supported, a caller would only have the option to walk away from an RPC call and leave threads and IO running. 

When Async Halibut was introduced, cooperative cancellation was introduced down to the socket for RPC calls to Listening Service and to the queue for Polling Services. Cancellation could be performed if the RPC call was connecting and in-flight. So as not to force clients to change there existing behaviour, two cancellation tokens were introduced, one to cancel when connecting and one to cancel when in-flight.

# Results

This PR extends the previous cancellation support so that an in-flight Polling Request can be cancelled to the Socket. Terminating any in-flight Reads or Writes on the TCP Connection.

The two cancellation tokens have been combined into one to simplify the cancellation logic. The ability to cancel just when connecting is no longer required in TentacleClient. Instead, the desire is to co-operatively cancel an RPC call regardless of it's current state.

Tentacle Client does need to know if the cancelled RPC call was connecting or in flight to make the most efficient decisions about what it has to do next, so cancellation will throw specific errors based on the state of the RPC call when it was cancelled.

net48 does not support full cancellation due to the usage of a deflate stream, which uses APM Begin and End methods and loses the cancellation token. As TentacleClient only supports net6.0, this is not an issue.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Halibut.Tests/ExceptionContractFixture.cs
```diff
@@ -215,8 +215,8 @@ public async Task WhenTheListeningRequestIsCancelledWhileConnecting_AHalibutClie
 
             var cancellationTokenSource = new CancellationTokenSource(TimeSpan.FromSeconds(5));
 
-            (await AssertException.Throws<HalibutClientException>(async () => await client.SayHelloAsync("Hello", new(cancellationTokenSource.Token, CancellationToken.None))))
-                .And.Message.Should().Contain("An error occurred when sending a request to 'https://20.5.79.31:10933/', after the request began: The operation was canceled.");
+            (await AssertException.Throws<HalibutClientException>(async () => await client.SayHelloAsync("Hello", new(cancellationTokenSource.Token))))
+                .And.Message.Should().Contain("An error occurred when sending a request to 'https://20.5.79.31:10933/', after the request began: The Request was cancelled while Connecting.");
```
nathanwoctopusdeploy, line 219: Exception contract change - The error message changes

source/Halibut.Tests/ExceptionContractFixture.cs
```diff
@@ -240,15 +240,8 @@ public async Task WhenTheListeningRequestIsCancelledWhileConnecting_AndTheConnec
...
-            // Halibut due to the way the code is written, so leaving this here as it should really be consistent.
-            await AssertException.Throws<OperationCanceledException>(async () => await client.SayHelloAsync("Hello", new(cancellationTokenSource.Token, CancellationToken.None)));
-#else
-            var exception = (await AssertException.Throws<HalibutClientException>(async () => await client.SayHelloAsync("Hello", new(cancellationTokenSource.Token, CancellationToken.None)))).And;
-            exception.Message.Should().Be($"An error occurred when sending a request to '{clientAndService.ServiceUri}', after the request began: The ReadAsync operation was cancelled.");
-#endif
+            var exception = (await AssertException.Throws<HalibutClientException>(async () => await client.SayHelloAsync("Hello", new(cancellationTokenSource.Token)))).And;
+            exception.Message.Should().Be($"An error occurred when sending a request to '{clientAndService.ServiceUri}', after the request began: The Request was cancelled while Connecting.");
```
nathanwoctopusdeploy, line 244: Exception contract change - The error message changes

source/Halibut.Tests/ExceptionContractFixture.cs
```diff
@@ -279,8 +272,8 @@ public async Task WhenTheListeningRequestIsCancelledWhileInProgress_AnOperationC
                 cancellationTokenSource.Cancel();
             });
 
-            (await AssertException.Throws<HalibutClientException>(async () => await doSomeActionClient.ActionAsync(new(cancellationTokenSource.Token, cancellationTokenSource.Token))))
-                .And.Message.Should().Contain($"An error occurred when sending a request to '{clientAndService.ServiceUri}', after the request began: The ReadAsync operation was cancelled.");
+            (await AssertException.Throws<HalibutClientException>(async () => await doSomeActionClient.ActionAsync(new(cancellationTokenSource.Token))))
+                .And.Message.Should().Contain($"An error occurred when sending a request to '{clientAndService.ServiceUri}', after the request began: The Request was cancelled while Transferring.");
```
nathanwoctopusdeploy, line 276: Exception contract change - The error message changes

source/Halibut.Tests/CancellationViaClientProxyFixture.cs
```diff
@@ -156,9 +95,12 @@ await using (var clientAndService = await clientAndServiceTestCase.CreateTestCas
                 // Give time for the cancellation to do something
                 await Task.Delay(TimeSpan.FromSeconds(2), CancellationToken);
 
-                (await AssertException.Throws<Exception>(inFlightRequest))
-                    .And
-                    .Should().Match<Exception>(x => x is OperationCanceledException || (x.GetType() == typeof(HalibutClientException) && x.Message.Contains("The ReadAsync operation was cancelled")));
+#pragma warning disable VSTHRD003
```
sburmanoctopus, line 98: I think we can get away with not disabling this warning.

If we use the override that just takes the task, that should fix it? E.g.:

```
(await AssertionExtensions.Should(inFlightRequest).ThrowAsync<Exception>())
```

source/Halibut.Tests/ExceptionContractFixture.cs
```diff
@@ -240,15 +240,8 @@ public async Task WhenTheListeningRequestIsCancelledWhileConnecting_AndTheConnec
 
             var cancellationTokenSource = new CancellationTokenSource(TimeSpan.FromSeconds(5));
 
-#if NETFRAMEWORK
```
sburmanoctopus, line 243: Nice that we could get rid of this 😄

source/Halibut.Tests/PollingServiceTimeoutsFixture.cs
```diff
@@ -100,9 +100,8 @@ await using (var clientAndService = await clientAndServiceTestCase.CreateTestCas
 
                 Wait.UntilActionSucceeds(() =>
                 {
-                    // Leaving these asserts as when cooperative cancellation is supported it should cause them to fail at which point they can be fixed to assert cancellation to the socket works as expected.
-                    connectionsObserver.ConnectionClosedCount.Should().Be(0, "Cancelling the PendingRequest does not cause the TCP Connection to be cancelled to stop the in-flight request");
-                    connectionsObserver.ConnectionAcceptedCount.Should().Be(1, "The Service won't have reconnected after the request was cancelled");
+                    connectionsObserver.ConnectionClosedCount.Should().Be(1, "Cancelling the PendingRequest does not cause the TCP Connection to be cancelled to stop the in-flight request");
```
sburmanoctopus, line 103: Does the comment here (and the line below) still apply?

source/Halibut.Tests/ServiceModel/PendingRequestQueueFixture.cs
```diff
@@ -379,47 +382,22 @@ public async Task QueueAndWait_CancellingAPendingRequestAfterItIsDequeued_Should
...
             cancellationTokenSource.Cancel();
             await Task.Delay(1000, CancellationToken);
+            dequeued!.CancellationToken.IsCancellationRequested.Should().BeTrue("Should have cancelled the request");
 
-            await sut.ApplyResponse(expectedResponse, request.Destination);
-
-            var response = await queueAndWaitTask;
-
+            
+#pragma warning disable VSTHRD003
```
sburmanoctopus, line 388: Like before, you can use the overload that just takes a task to remove the need for adding the pragma.

source/Halibut.Tests/WhenCancellingARequestForAPollingTentacle.cs
```diff
@@ -47,99 +46,201 @@ public class AndTheRequestHasBeenDequeuedButNoResponseReceived : BaseTest
...
-                await using (var clientAndService = await clientAndServiceTestCase.CreateTestCaseBuilder()
-                           .AsLatestClientAndLatestServiceBuilder()
-                           .WithHalibutLoggingLevel(LogLevel.Trace)
-                           .WithDoSomeActionService(() =>
-                           {
-                               calls.Add(DateTime.UtcNow);
+        /// <summary>
+        /// CancelWhenRequestQueuedPendingRequestQueueFactory cancels the cancellation token source when a request is queued
+        /// </summary>
+        class CancelWhenRequestQueuedPendingRequestQueueFactory : IPendingRequestQueueFactory
```
sburmanoctopus, line 127: It looks like this is putting back a class we moved into its own file before.

nathanwoctopusdeploy: Ahh good spot, it wasn't a pleasant rebase!!

source/Halibut.Tests/CancellationViaClientProxyFixture.cs
nathanwoctopusdeploy: Ah nice, I just hammered in the ignores 😆

source/Halibut.Tests/PollingServiceTimeoutsFixture.cs
nathanwoctopusdeploy: Ahh good spot, I haven't updated the because 🙇

sburmanoctopus (approved):
LGTM. Nice to see this make its way back again 😁

source/Halibut.Tests/Support/PendingRequestQueueFactories/CancelWhenRequestDequeuedPendingRequestQueueFactory.cs
```diff
@@ -0,0 +1,72 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Halibut.ServiceModel;
+using Halibut.Transport.Protocol;
+
+namespace Halibut.Tests.Support.PendingRequestQueueFactories
+{
+    /// <summary>
+    /// CancelWhenRequestDequeuedPendingRequestQueueFactory cancels the cancellation token source when a request is queued
```
LukeButters, line 10: when request is `de`queued?

LukeButters (approved):

PR #577, opened 2023-12-05
Stop Timeout While Waiting for Response

sburmanoctopus:
[sc-59506]

# Background

It was noticed that `PollingRequestMaximumMessageProcessingTimeout` was redundant. 

`PollingRequestMaximumMessageProcessingTimeout` is the amount of time we will wait for a response to be given from a request that has already been dequeued. 

The **down side** of keeping `PollingRequestMaximumMessageProcessingTimeout` is that it will also timeout an operation that is successful, but taking a long time (e.g. downloading a large package to a Tentacle, or over a slow network, complex script/action for Tentacle to run etc.).

We think it was added years ago as a way to ensure we do not wait forever if communications with the Tentacle dies.

But we believe this won't be an issue because:
- We always make sure a response is set, regardless of success or failure. Applying a response stops us waiting
- We have timeouts and keep alives applied to the TCP connection. So if we loose communications with the Tentacle, these mechanisms will be what fail and stop us from waiting forever.
- Web sockets recently was enhanced to time out. So there should be no other types of polling tentacle that wait forever.

# Results

Related to OctopusDeploy/Issues#8229

## Before

No matter why it took a long time to complete a task, we would always timeout after `PollingRequestMaximumMessageProcessingTimeout`

## After

We now simply wait until a response is applied, and rely on the communication timeouts and response applying to break the waiting.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Builders/PendingRequestQueueBuilder.cs
+++ b/source/Halibut.Tests/Builders/PendingRequestQueueBuilder.cs
@@ -9,6 +9,7 @@ namespace Halibut.Tests.Builders
         ILog? log;
         string? endpoint;
         TimeSpan? pollingQueueWaitTimeout;
+        bool? relyOnConnectionTimeoutsInsteadOfPollingRequestMaximumMessageProcessingTimeout;
 
         public PendingRequestQueueBuilder WithEndpoint(string endpoint)
         {
@@ -27,14 +28,23 @@ namespace Halibut.Tests.Builders
             this.pollingQueueWaitTimeout = pollingQueueWaitTimeout;
             return this;
         }
-        
+
+        public PendingRequestQueueBuilder WithRelyOnConnectionTimeoutsInsteadOfPollingRequestMaximumMessageProcessingTimeout(bool relyOnConnectionTimeoutsInsteadOfPollingRequestMaximumMessageProcessingTimeout)
+        {
+            this.relyOnConnectionTimeoutsInsteadOfPollingRequestMaximumMessageProcessingTimeout = relyOnConnectionTimeoutsInsteadOfPollingRequestMaximumMessageProcessingTimeout;
+            return this;
+        }
+
         public IPendingRequestQueue Build()
         {
             var endpoint = this.endpoint ?? "poll://endpoint001";
-            var pollingQueueWaitTimeout = this.pollingQueueWaitTimeout ?? new HalibutTimeoutsAndLimitsForTestsBuilder().Build().PollingQueueWaitTimeout;
+            var halibutTimeoutsAndLimits = new HalibutTimeoutsAndLimitsForTestsBuilder().Build();
             var log = this.log ?? new InMemoryConnectionLog(endpoint);
 
-            return new PendingRequestQueueAsync(log, pollingQueueWaitTimeout);
+            halibutTimeoutsAndLimits.PollingQueueWaitTimeout = pollingQueueWaitTimeout ?? halibutTimeoutsAndLimits.PollingQueueWaitTimeout;
+            halibutTimeoutsAndLimits.RelyOnConnectionTimeoutsInsteadOfPollingRequestMaximumMessageProcessingTimeout = relyOnConnectionTimeoutsInsteadOfPollingRequestMaximumMessageProcessingTimeout ?? halibutTimeoutsAndLimits.RelyOnConnectionTimeoutsInsteadOfPollingRequestMaximumMessageProcessingTimeout;
+
+            return new PendingRequestQueueAsync(log, halibutTimeoutsAndLimits);
         }
     }
 }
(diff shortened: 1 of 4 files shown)


source/Halibut.Tests/Builders/PendingRequestQueueBuilder.cs
```diff
@@ -9,6 +9,7 @@ public class PendingRequestQueueBuilder
         ILog? log;
         string? endpoint;
         TimeSpan? pollingQueueWaitTimeout;
+        bool? relyOnConnectionTimeoutsInsteadOfPollingRequestMaximumMessageProcessingTimeout;
```
LukeButters, line 12: Could we have this as non nullable, and just default it to false?

sburmanoctopus: I personally would prefer to use the default from `HalibutTimeoutsAndLimits` by default, and only use this if the test has actually chosen to change it.

LukeButters: Oh i see the complexity lower down, probably is better here :+1:

source/Halibut.Tests/ServiceModel/PendingRequestQueueFixture.cs
```diff
@@ -132,6 +131,44 @@ public async Task QueueAndWait_WhenRequestIsDequeued_ButPollingRequestQueueTimeo
...
         
+        [Test]
+        public async Task QueueAndWait_WhenRequestIsDequeued_WithRelyOnConnectionTimeoutsInsteadOfPollingRequestMaximumMessageProcessingTimeout_WillWaitForeverUntilResponseIsSet()
+        {
+            // Arrange
+            const string endpoint = "poll://endpoint001";
+
+            var sut = new PendingRequestQueueBuilder()
+                .WithEndpoint(endpoint)
+                .WithPollingQueueWaitTimeout(TimeSpan.Zero) // Remove delay, otherwise we wait the full 20 seconds for DequeueAsync at the end of the test
```
LukeButters, line 142: Maybe set this to `1s` otherwise we might be flooding the network with null requests and we might not be giving enough time to actually collect from the queue.

sburmanoctopus: We do this throughout this test file.

We should do them as a set. Do it as part of this PR?

LukeButters (approved):

source/Halibut.Tests/ServiceModel/PendingRequestQueueFixture.cs
LukeButters: Oh i didn't realise this is how all other tests were configured, I think it makes sense to just do what is working :D

source/Halibut/Diagnostics/HalibutTimeoutsAndLimits.cs
```diff
@@ -18,6 +18,12 @@ public class HalibutTimeoutsAndLimits
         /// </summary>
         public TimeSpan PollingRequestMaximumMessageProcessingTimeout { get; set; } = TimeSpan.FromMinutes(10);
 
+        /// <summary>
+        ///     We believe that PollingRequestMaximumMessageProcessingTimeout is redundant.
+        ///     This setting allows us to feature toggle turning off PollingRequestMaximumMessageProcessingTimeout
```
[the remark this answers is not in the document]
sburmanoctopus: Sounds good. Done!

source/Halibut/ServiceModel/PendingRequestQueueAsync.cs
```diff
@@ -16,22 +16,22 @@ public class PendingRequestQueueAsync : IPendingRequestQueue
         readonly SemaphoreSlim queueLock = new(1, 1);
         readonly AsyncManualResetEvent itemAddedToQueue = new(false);
         readonly ILog log;
-        readonly TimeSpan pollingQueueWaitTimeout;
+        readonly HalibutTimeoutsAndLimits halibutTimeoutsAndLimits;
```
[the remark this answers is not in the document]
sburmanoctopus: Lol, I was undecided on whether to do that or not.

I went with what I did so that if we need more settings from `HalibutTimeoutsAndLimits` in future, it would be much easier to add.

But I'm happy with passing the extra bool in specifically, as I was 50/50 on it 😁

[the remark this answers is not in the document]
sburmanoctopus: @LukeButters , [this commit](https://github.com/OctopusDeploy/Halibut/pull/577/commits/3cafbe1a3d165d6536eef56f8d6229c665e67638) addresses this.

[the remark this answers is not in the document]
LukeButters: lgtm

PR #582, opened 2023-12-15
Passing operation cancelled exceptions up instead of wrapping them in HalibutClientExceptions

sburmanoctopus:
[sc-66743]

# Background

We want to make it easier to detect an `OperationCanceledException` from consumers of Halibut (e.g. Tentacle Client).

# Results

## Before

All exceptions throw within the `SecureListeningClient` were wrapped in a `HalibutClientException`. Meaning consumers would have to 'unpack' the exception to see if it was an `OperationCanceledException`

## After

We now throw an exception of the respective `OperationCanceledException` type instead.

Contractually, this should be fine for the consumer, as Polling tentacles will already throw `OperationCanceledExceptions`.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Transport/SecureListeningClient.cs
+++ b/source/Halibut/Transport/SecureListeningClient.cs
@@ -183,7 +183,9 @@ namespace Halibut.Transport
         void HandleError(Exception? lastError, bool retryAllowed, bool hasConnected)
         {
             if (lastError == null)
+            {
                 return;
+            }
 
             lastError = lastError.UnpackFromContainers();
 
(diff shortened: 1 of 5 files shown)


source/Halibut/Transport/SecureListeningClient.cs
```diff
@@ -183,23 +183,35 @@ public async Task ExecuteTransactionAsync(ExchangeActionAsync protocolHandler, C
...
             var error = new StringBuilder();
             error.Append("An error occurred when sending a request to '").Append(ServiceEndpoint.BaseUri).Append("', ");
             error.Append(retryAllowed ? "before the request could begin: " : "after the request began: ");
             error.Append(lastError.Message);
-
-            var inner = lastError as SocketException;
-            if (inner != null)
+            
+            
+            switch (lastError)
```
nathanwoctopusdeploy, line 198: Should we do the same for the websocket client? https://github.com/OctopusDeploy/Halibut/blob/cdf0c4606b4903121fca866e203a34978a25a952/source/Halibut/Transport/SecureWebSocketClient.cs#L150

nathanwoctopusdeploy (approved):
LGTM

PR #589, opened 2023-12-19
Remove a race condition from QueueAndWait_Can_Queue_Dequeue_WhenRequestsAreBeingCancelled

LukeButters:
# Background

[SC-67432]

Sometimes (2 in 1000 runs) this test fails with:
```
Expected completedTask.Count(x => x != *******) to be less than 400, but found 400.
```

This might be because:
* cancelling the cancellation token on the tuple is relatively slow, while de-queueing from the queue is relatively fast.
* the count of cancelled requests was incremented before the `CancellationTokenSource` was cancelled.
* the number of completed tasks could legitimately be 400 if exactly 100 requests were cancelled. (total request count is 500)

this means it might be possible to be in a state in which only:
* 99
* or 100

tasks are cancelled!

Thus 400 tasks (or technically even 401 tasks) could have completed successfully. Explaining the `to be less than 400, but found 400.` error.

The fix is to:
* check that completed is smaller than or **equal to** total tasks - cancelled tasks.
* the count of cancelled request to only be incremented if the request is actually cancelled.

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/ServiceModel/PendingRequestQueueFixture.cs
+++ b/source/Halibut.Tests/ServiceModel/PendingRequestQueueFixture.cs
@@ -400,7 +400,7 @@ namespace Halibut.Tests.ServiceModel
             // This test is pretty non-deterministic. It attempts to create some cancellation chaos
             // and we just want to ensure some things did cancel and some things did complete and nothing
             // threw an exception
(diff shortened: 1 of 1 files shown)


sburmanoctopus (approved):
LGTM

PR #595, opened 2024-02-19
Remove pollingRequestMaximumMessageProcessingTimeout

LukeButters:
# Background

[SC-67117]

Removes the `pollingRequestMaximumMessageProcessingTimeout` since it provided no value, we now entirely depend on TCP connection timeouts to determine if the request is dead similar to listening tentacles.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Support/PendingRequestQueueFactories/CancelWhenRequestDequeuedPendingRequestQueueFactory.cs
+++ b/source/Halibut.Tests/Support/PendingRequestQueueFactories/CancelWhenRequestDequeuedPendingRequestQueueFactory.cs
@@ -67,6 +67,11 @@ namespace Halibut.Tests.Support.PendingRequestQueueFactories
 
             public async Task<ResponseMessage> QueueAndWaitAsync(RequestMessage request, CancellationToken requestCancellationToken)
                 => await inner.QueueAndWaitAsync(request, requestCancellationToken);
+
+            public ValueTask DisposeAsync()
+            {
+                return inner.DisposeAsync();
+            }
         }
     }
 }
--- a/source/Halibut.Tests/Support/PendingRequestQueueFactories/CancelWhenRequestQueuedPendingRequestQueueFactory.cs
+++ b/source/Halibut.Tests/Support/PendingRequestQueueFactories/CancelWhenRequestQueuedPendingRequestQueueFactory.cs
@@ -61,6 +61,11 @@ namespace Halibut.Tests.Support.PendingRequestQueueFactories
                 await task;
                 return result;
             }
+
+            public ValueTask DisposeAsync()
+            {
+                return this.inner.DisposeAsync();
+            }
         }
     }
 }
--- a/source/Halibut.sln.DotSettings
+++ b/source/Halibut.sln.DotSettings
@@ -246,6 +246,7 @@ II.2.12 &lt;HandlesEvent /&gt;&#xD;
 	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/ALIGN_MULTILINE_PARAMETER/@EntryValue">False</s:Boolean>
 	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/EXPLICIT_PRIVATE_MODIFIER/@EntryValue">False</s:Boolean>
 	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/EXPLICIT_INTERNAL_MODIFIER/@EntryValue">False</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EFeature_002EServices_002ECodeCleanup_002EFileHeader_002EFileHeaderSettingsMigrate/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpAttributeForSingleLineMethodUpgrade/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpFileLayoutPatternsUpgrade/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpKeepExistingMigration/@EntryIndexedValue">True</s:Boolean>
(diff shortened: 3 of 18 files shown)


source/Halibut.Tests/Support/PendingRequestQueueFactories/CancelWhenRequestDequeuedPendingRequestQueueFactory.cs
```diff
@@ -67,6 +67,11 @@ public async Task ApplyResponse(ResponseMessage response, ServiceEndPoint destin
 
             public async Task<ResponseMessage> QueueAndWaitAsync(RequestMessage request, CancellationToken requestCancellationToken)
                 => await inner.QueueAndWaitAsync(request, requestCancellationToken);
+
+            public ValueTask DisposeAsync()
```
sburmanoctopus, line 71: Ooh, nice one. 

Just wondering, how did this manifest as a problem?

source/Halibut.Tests/Support/PendingRequestQueueFactories/CancelWhenRequestQueuedPendingRequestQueueFactory.cs
```diff
@@ -61,6 +61,11 @@ public async Task<ResponseMessage> QueueAndWaitAsync(RequestMessage request, Can
                 await task;
                 return result;
             }
+
```
sburmanoctopus, line 64: Same here (just wondering 😁 )

source/Halibut.sln.DotSettings
```diff
@@ -246,6 +246,7 @@ II.2.12 &lt;HandlesEvent /&gt;&#xD;
 	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/ALIGN_MULTILINE_PARAMETER/@EntryValue">False</s:Boolean>
 	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/EXPLICIT_PRIVATE_MODIFIER/@EntryValue">False</s:Boolean>
 	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/EXPLICIT_INTERNAL_MODIFIER/@EntryValue">False</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EFeature_002EServices_002ECodeCleanup_002EFileHeader_002EFileHeaderSettingsMigrate/@EntryIndexedValue">True</s:Boolean>
```
sburmanoctopus, line 249: What was this added for?

source/Halibut/HalibutRuntime.cs
```diff
@@ -338,6 +338,16 @@ public void Dispose()
 
         public async ValueTask DisposeAsync()
         {
+            foreach (var pendingRequestQueue in this.queues.Values)
```
sburmanoctopus, line 341: Oh what! We weren't disposing of these... Far out.

Just wondering (again), how did you notice this?

source/Halibut/HalibutRuntime.cs
```diff
@@ -338,6 +338,16 @@ public void Dispose()
...
+            {
+                try
```
sburmanoctopus, line 343: Don't we have an extension method somewhere that tries to dispose? Or is that Tentacle? 🤔 

Not a big deal, but it would be nice to use it if we have it (for consistency)

source/Halibut/ServiceModel/PendingRequestQueueAsync.cs
```diff
@@ -11,34 +11,35 @@
...
-        public PendingRequestQueueAsync(ILog log, TimeSpan pollingQueueWaitTimeout, bool relyOnConnectionTimeoutsInsteadOfPollingRequestMaximumMessageProcessingTimeout)
+        public PendingRequestQueueAsync(ILog log, TimeSpan pollingQueueWaitTimeout)
         {
             this.log = log;
             this.pollingQueueWaitTimeout = pollingQueueWaitTimeout;
-            this.relyOnConnectionTimeoutsInsteadOfPollingRequestMaximumMessageProcessingTimeout = relyOnConnectionTimeoutsInsteadOfPollingRequestMaximumMessageProcessingTimeout;
         }
 
-        public async Task<ResponseMessage> QueueAndWaitAsync(RequestMessage request, CancellationToken cancellationToken)
+        public async Task<ResponseMessage> QueueAndWaitAsync(RequestMessage request, CancellationToken ct)
```
sburmanoctopus, line 37: Nit: I think it's nicer as `cancellationToken` instead of `ct`.

Abbreviations are ok for lambda etc, but I think spelling it out for methods (especially public ones) is clearer.

sburmanoctopus: And adjust the variable below instead.

source/Halibut/ServiceModel/PendingRequestQueueAsync.cs
```diff
@@ -378,5 +378,12 @@ public void Dispose()
                 transferLock?.Dispose();
             }
         }
+
```
sburmanoctopus, line 381: Nit: I think I'd move this method higher. It feels a little like it's part of `PendingRequest`, where it's not.

source/Halibut/ServiceModel/PendingRequestQueueInternalException.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+
+namespace Halibut.ServiceModel
+{
+    public class PendingRequestQueueInternalException : Exception
```
sburmanoctopus, line 5: Do we not have something similar already?

Also, are we planning on catching this type in Tentacle Client/Octopus Server?

source/Halibut/HalibutRuntime.cs
sburmanoctopus: Disregard. I see lower that disposing is a new thing.

source/Halibut.Tests/Support/PendingRequestQueueFactories/CancelWhenRequestDequeuedPendingRequestQueueFactory.cs
LukeButters: One of the tests would enqueue work, deque that work then never respond. I saw tests never finishing dumped process and could see the stack trace.

I think before with the timeout it would eventually undo.

Without the timeout removed here that would never happen. Note that when used for real we ALWAYS respond to the in memory queue.

source/Halibut.sln.DotSettings
LukeButters: What is this? Time to remove.

LukeButters: Removed

source/Halibut/HalibutRuntime.cs
LukeButters: lol yes

source/Halibut/HalibutRuntime.cs
LukeButters: I could not see one.

source/Halibut/ServiceModel/PendingRequestQueueInternalException.cs
LukeButters: I didn't see something like this already existing.

I don't plan on catching it yet, but I do want a specific exception type to capture what is happening. Rather than always doing string matching.

source/Halibut/ServiceModel/PendingRequestQueueAsync.cs
LukeButters: I actually just re-used the same `cancellationToken` variable. Since it never makes sense to use the given one after we make the linked token. Doing this removes the possibility of using the wrong token.

PR #596, opened 2024-03-21
Add option for serializing the message envelope as generic

Discolai:
# Background

Moving a service response type to another namespace/assembly breaks backwards compatibility of s service. This is because `Newtonsoft.Json` is configured with `TypeNameHandling.Auto` and the response is serialized as `MessageEnvelope<object>` instead of `MessageEnvelope<T>`. 

My team is currently handling this by forwarding to the correct type manually in a custom `SerializationBinder`, but this is prone to errors and introduces manual steps to our releases.

You might also consider making this the default behavior if this is true: `Once ALL sources and targets are deserializing to MessageEnvelope<T>, (ReadBsonMessage) then this can be changed to T`

# Results

Users of the Halibut runtime can optionally specify that the message envelope will be serialized as a generic. Thus removing the `$type` from the response 

# Pre-requisites
- [ x] I have read [How we use GitHub Issues](https://github.com/OctopusDeploy/Issues/blob/main/docs/CONTRIBUTING.internal.md) for help deciding when and where it's appropriate to make an issue.
- [ x] I have considered informing or consulting the right people, according to the [ownership map](https://whimsical.com/ownership-map-NzbiD4HJyvhC9jNJNfS6TG).
- [ x] I have considered appropriate testing for my change.

### DIFF
--- a/source/Halibut/Transport/Protocol/MessageSerializer.cs
+++ b/source/Halibut/Transport/Protocol/MessageSerializer.cs
@@ -20,6 +20,7 @@ namespace Halibut.Transport.Protocol
         readonly long readIntoMemoryLimitBytes;
         readonly long writeIntoMemoryLimitBytes;
         readonly DeflateStreamInputBufferReflector deflateReflector;
+        readonly bool useGenericMessageEnvelope;
         
         internal MessageSerializer(
             ITypeRegistry typeRegistry,
@@ -27,7 +28,8 @@ namespace Halibut.Transport.Protocol
             IMessageSerializerObserver observer,
             long readIntoMemoryLimitBytes,
             long writeIntoMemoryLimitBytes,
-            ILogFactory logFactory)
+            ILogFactory logFactory,
+            bool useGenericMessageEnvelope)
         {
             this.typeRegistry = typeRegistry;
             this.createStreamCapturingSerializer = createStreamCapturingSerializer;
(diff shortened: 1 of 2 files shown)


Discolai:
@LukeButters I am not sure if you are the person to tag, so feel free to bring in another reviewer.

This addition will improve our release experience significantly, so I would appreciate it if you could consider merging it

LukeButters:
Could you add a test? It should be possible since you could put the old type in https://github.com/OctopusDeploy/Halibut/tree/main/source/Halibut.TestUtils.CompatBinary.Base
and the new types within the normal test project https://github.com/OctopusDeploy/Halibut/tree/main/source/Halibut.Tests

Ideally the tests will show this fixes the issue.

Discolai:
> Could you add a test? It should be possible since you could put the old type in https://github.com/OctopusDeploy/Halibut/tree/main/source/Halibut.TestUtils.CompatBinary.Base and the new types within the normal test project https://github.com/OctopusDeploy/Halibut/tree/main/source/Halibut.Tests
> 
> Ideally the tests will show this fixes the issue.

I am sorry, but my initial testing was more of a proof of conecpt. And I didn't realize that `ResponseMessage.Result` was an `object` type. So this change does not fix the issue afterall. One would need to make `ResponseMessage.Result` generic as well.

And that seems like a big task which could easily break general backwards compatibility. So I will close this pr

LukeButters:
Thanks for your PR @Discolai. While we didn’t get to a merge this time, we welcome PRs and suggestions. For significant changes, we recommend that you open an issue to discuss proposed changes to get alignment on the direction. :smile_cat: 

If it helps [Tentacle](https://github.com/OctopusDeploy/OctopusTentacle), which uses Halibut, had to deal with a change in namespace and assembly for some types. If I remember correctly this is done by converting to the old namespace/assembly on the way out and converts back to the new ones on the way in see [here](https://github.com/OctopusDeploy/OctopusTentacle/blob/main/source/Octopus.Tentacle.Contracts/Legacy/MessageSerializerBuilderExtensionMethods.cs), [here](https://github.com/OctopusDeploy/OctopusTentacle/blob/main/source/Octopus.Tentacle.Contracts/Legacy/NamespaceMappingSerializationBinderDecorator.cs) and [here](https://github.com/OctopusDeploy/OctopusTentacle/blob/main/source/Octopus.Tentacle.Contracts/Legacy/AssemblyMappingSerializationBinderDecorator.cs).

PR #602, opened 2024-05-02
Add simple message to polling timeout error

susanpann:
# Background
The errors around polling tentacles not connecting are vague and don't help the customer investigate issues that may arise. This PR adds a simple message to re-direct them to the tentacle logs to investigate, as there is no current way of writing the halibut exception from tentacle to the server/task logs.

# Results
Relates to https://github.com/OctopusDeploy/Issues/issues/8761

# Pre-requisites


tothegills (changes requested):
Halibut doesn't know that Tentacle exists. We would need to provide the suggestion somewhere that is Tentacle-aware, probably in Octopus server.

PR #604, opened 2024-05-02
Add test showing exception when nullable parameter is called with nul…

### LINKED ISSUE #603: Passing null for a Nullable<T> parameter fails with AmbiguousMatchException
### Team

- [X] I've assigned a team label to this issue

### What happened?

When passing `null` to an asynchronous service method accepting a `Nullable<T>` parameter, Halibut will not consider it a match and fails with the following exception:

```
Halibut.Exceptions.AmbiguousMethodMatchHalibutClientException : Could not decide which candidate to call out of the following methods:
 - System.Threading.Tasks.Task`1[System.Int32] IncrementAsync(System.Threading.CancellationToken)
 - System.Threading.Tasks.Task`1[System.Int32] IncrementAsync(System.Nullable`1[System.Int32], System.Threading.CancellationToken)
The request arguments were:
<null>
```

### Reproduction

Define an async service method like this:

```csharp
    public interface ICountingService
    {
        public int Increment(int? number);
    }

    public interface IAsyncCountingService
    {
        public Task<int> IncrementAsync(int? number, CancellationToken cancellationToken);
    }
```

And invoke it like this:

```csharp
[Test]
public async Task AsyncInvokeWithNullableParamsOnAsyncService()
{
    var serviceFactory = new ServiceFactoryBuilder()
        .WithConventionVerificationDisabled()
        .WithService<ICountingService, IAsyncCountingService>(() => new AsyncCountingService())
        .Build();

    var sut = new ServiceInvoker(serviceFactory);
    var request = new RequestMessage()
    {
        ServiceName = nameof(ICountingService),
        MethodName = nameof(ICountingService.Increment),
        Params = new object[] { null! },
    };

    var response = await sut.InvokeAsync(request);
    response.Result.Should().Be(1);
}
```

### Error and Stacktrace

```shell
Halibut.Exceptions.AmbiguousMethodMatchHalibutClientException : Could not decide which candidate to call out of the following methods:
 - System.Threading.Tasks.Task`1[System.Int32] IncrementAsync(System.Threading.CancellationToken)
 - System.Threading.Tasks.Task`1[System.Int32] IncrementAsync(System.Nullable`1[System.Int32], System.Threading.CancellationToken)
The request arguments were:
<null>
```

### More Information

I have written a test and patched the bug, will submit a PR ASAP.

### Workaround

_No response_

khellang:
# Background

# Results

Fixes #603

## Before

Passing `null` to a service method with a `Nullable<T>` parameter failed. See #603

## After

Invoking the method no longer fails.

# Pre-requisites


khellang:
Hello good folks! 👋 Any hint on when we can get this looked at? We've worked around it by splitting the RPC interface with different overloads with and without the nullable argument, but would really like a proper fix 😅

LukeButters (approved):

PR #606, opened 2024-05-03
Treat thumbprints as case-insensitive

### LINKED ISSUE #605: Certificate thumbprints should not be case-sensitive
### Team

- [X] I've assigned a team label to this issue

### What happened?

Providing a thumbprint with different casing on the client and server-side gives the following exception:

> The server at https://localhost:8080/ presented an unexpected security certificate. We expected the server to present a certificate with the thumbprint 'a7b12a038945afdee39661bcf59cf1600c811122'. Instead, it presented a certificate with a thumbprint of 'A7B12A038945AFDEE39661BCF59CF1600C811122' and subject '<subject>'.

As you can see, the thumbprint is correct, but differs in casing. It expected lowercase, but got uppercase. Certificate thumbprints are a hex-string representation of SHA-based hashes and shouldn't be treated as equal whether they are upper- or lower case.

### Reproduction

Provide a thumbprint with different casing on the client and server-side.

### Workaround

Aligning the thumbprints with the same case fixes the problem.

khellang:
# Background

See #605

# Results

Fixes #605

# Pre-requisites

### DIFF
--- a/source/Halibut/ServiceEndPoint.cs
+++ b/source/Halibut/ServiceEndPoint.cs
@@ -86,7 +86,9 @@ namespace Halibut
         {
             if (ReferenceEquals(null, other)) return false;
             if (ReferenceEquals(this, other)) return true;
-            return string.Equals(RemoteThumbprint, other.RemoteThumbprint) && string.Equals(baseUriString, other.baseUriString) && Equals(Proxy, other.Proxy);
+            return StringComparer.OrdinalIgnoreCase.Equals(RemoteThumbprint, other.RemoteThumbprint)
+                && string.Equals(baseUriString, other.baseUriString)
+                && Equals(Proxy, other.Proxy);
         }
 
         public override bool Equals(object? obj)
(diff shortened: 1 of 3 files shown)


sburmanoctopus:
Thank you for you submission.

We were wondering about the reasons behind doing this change. Could you please elaborate on the motivation?

khellang:
> Could you please elaborate on the motivation?

We encountered the error described in #605, where the server had the thumbprint in uppercase and the client had the thumbprint in lowercase (or vice versa).

sburmanoctopus:
Hi @khellang,

Thank you for taking the time to prepare this contribution, we appreciate and value your time, unfortunately this time we won't be able to accept this contribution.

Halibut is a very important part of Octopus Deploy, and technically this is a breaking change. While we agree this is a positive change, however the fact that consumers of the library should be able to account for it's shortcoming easily means we aren't able to justify the cost of making this breaking change right now.

khellang:
Hi @sburmanoctopus ! 👋

Can you explains why this is a breaking change? I'm falling to see how this could break anything 😅

sburmanoctopus:
Hi @khellang,

To make this change we would also want to update the trust accepting side of halibut to be case insensitive, that way we don't have some checks as case sensitive and other checks as case insensitive. Users of Halibut can supply their own `ITrustProvider`, which won't know we have switched from case sensitive to case in-sensitive. In a sense that is a breaking change, although in theory it should never be a problem.

Since this is in the security area of Halibut, we would prefer to not make any change unless absolutely necessary. As we want to avoid inadvertently introducing a security vulnerability.

In theory a workaround to the issue your facing is to always upper case the thumbprint to the ServiceEndpoint`.

Since this issue has a workaround, and would require changes in the security areas of Halibut and with our general hesitation to add unnecessary changes to Halibut (we have Halibut services in very remote locations we don't want to break) let's see if holding off on this change is feasible.

PR #607, opened 2024-05-07
Add bidirectional tests

zentron:
Simple test confirming once a connection is established the service messaging layer is bi-directional

### DIFF
--- a/source/Halibut.Tests/BiDirectionalTests.cs
+++ b/source/Halibut.Tests/BiDirectionalTests.cs
@@ -1,0 +1,110 @@
+﻿using System;
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Halibut.ServiceModel;
+using Halibut.Tests.Support;
+using Halibut.Tests.TestServices;
+using Halibut.Tests.TestServices.Async;
+using Halibut.TestUtils.Contracts;
+using Microsoft.VisualStudio.Threading;
+using NUnit.Framework;
+
+namespace Halibut.Tests
+{
+    public class BiDirectionalTests : BaseTest
+    {
+        /// <summary>
+        /// Test is redundant but kept around since we really want the security part to work. 
+        /// </summary>
+        [Test]
+        public async Task ClientCanRespondToServerRequests()
+        {
+            await SetupBiDirectionalClients(async (octopusEchoClient, tentacleEchoClient) =>
+            {
+                var octopusResponse = await octopusEchoClient.SayHelloAsync("Hello");
+                var tentacleResponse = await tentacleEchoClient.SayHelloAsync("World");
+
+                octopusResponse.Should().Be("Hello...");
+                tentacleResponse.Should().Be("World...");
+            });
+        }
+
+        [Test]
+        public async Task ConcurrentTasksDoNotBlock()
+        {
+            await SetupBiDirectionalClients(async (octopusEchoClient, tentacleEchoClient) =>
+            {
+                var sw = Stopwatch.StartNew();
+                var t1 = Task.Run(async () => await RunLongRunningTask(octopusEchoClient)).WithTimeout(TimeSpan.FromSeconds(12));
+                var t2 = Task.Run(async () => await RunLongRunningTask(tentacleEchoClient)).WithTimeout(TimeSpan.FromSeconds(12));
+
+                await Task.WhenAll(t1, t2);
+                sw.Stop();
+
+                sw.ElapsedMilliseconds.Should().BeLessThan(await t1 + await t2);
+            });
+            
+            async Task<long> RunLongRunningTask(IAsyncClientEchoService tentacleEchoClient)
+            {
+                var longRunningDuration = Stopwatch.StartNew();
+                await tentacleEchoClient.LongRunningOperationAsync();
+                longRunningDuration.Stop();
+                return longRunningDuration.ElapsedMilliseconds;
+            }
+        }
+
+        [Test]
+        public async Task ConcurrentClientAndServerRequestsCorrectlyInterleaved()
+        {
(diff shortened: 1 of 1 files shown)


source/Halibut.Tests/BiDirectionalTests.cs
```diff
@@ -0,0 +1,110 @@
...
+using Microsoft.VisualStudio.Threading;
+using NUnit.Framework;
+
+namespace Halibut.Tests
+{
+    public class BiDirectionalTests : BaseTest
+    {
+        /// <summary>
+        /// Test is redundant but kept around since we really want the security part to work. 
+        /// </summary>
```
LukeButters, line 20: I think these lines can be dropped

zentron: :) Thanks. already gone locally.

source/Halibut.Tests/BiDirectionalTests.cs
```diff
@@ -0,0 +1,110 @@
...
+            await SetupBiDirectionalClients(async (octopusEchoClient, tentacleEchoClient) =>
+            {
+                var sw = Stopwatch.StartNew();
+                var t1 = Task.Run(async () => await RunLongRunningTask(octopusEchoClient)).WithTimeout(TimeSpan.FromSeconds(12));
+                var t2 = Task.Run(async () => await RunLongRunningTask(tentacleEchoClient)).WithTimeout(TimeSpan.FromSeconds(12));
+
+                await Task.WhenAll(t1, t2);
+                sw.Stop();
+
+                sw.ElapsedMilliseconds.Should().BeLessThan(await t1 + await t2);
```
LukeButters, line 47: It may be worth replacing the timing to test concurrency by using a "do some action service" and having that action service wait until both requests are in flight before returning.

The tests are run in parallel so the load tends to cause tests with timings to occasionally fail. Sometimes we make those tests run as non parallel but that just means it takes longer to run the tests :(

source/Halibut.Tests/BiDirectionalTests.cs
```diff
@@ -0,0 +1,110 @@
...
+                return longRunningDuration.ElapsedMilliseconds;
+            }
+        }
+
+        [Test]
+        public async Task ConcurrentClientAndServerRequestsCorrectlyInterleaved()
+        {
+            await SetupBiDirectionalClients(async (octopusEchoClient, tentacleEchoClient) =>
+            {
+                var t1 = Task.Run(async () => await RunEchoTask(octopusEchoClient)).WithTimeout(TimeSpan.FromSeconds(5));
```
LukeButters, line 66: What is the intent of the timeout? The base class should be providing a timeout on the test if the concern is this will run forever.

source/Halibut.Tests/BiDirectionalTests.cs
```diff
@@ -0,0 +1,110 @@
...
+                 var j = 0;
+                 while (j++ < 10)
+                 {
+                     var tentacleResponse = await tentacleEchoClient.SayHelloAsync($"Hello{j}");
+                     tentacleResponse.Should().Be($"Hello{j}...");
+                 }
+            }
+        }
+        
+        async Task SetupBiDirectionalClients(Func<IAsyncClientEchoService, IAsyncClientEchoService, Task> thing)
```
LukeButters: Nit: it would be great if this was made to be a builder like the builders used in other tests. However I don't suspect we will have lots more of these types of tests.

zentron: Unless we start using this format elsewhere I'd probably err towards YAGNI otherwise its hard to gauge what the reusable parts are.

source/Halibut.Tests/BiDirectionalTests.cs
zentron: 👍 I've update the timeout limits for use in the test,

source/Halibut.Tests/BiDirectionalTests.cs
zentron: have converted the quick POC hack towards something that is hopefully more reliable

source/Halibut.Tests/BiDirectionalTests.cs
```diff
@@ -0,0 +1,182 @@
...
+            Task BlockWaitingForSharedStateIncrementAsync(CancellationToken cancellationToken);
+            Task<string> SayHelloAsync(string name, CancellationToken cancellationToken);
+        }
+
+        public class MyEchoService : IAsyncMyEchoService
+        {
+            public static int SharedState = 0;
+            public async Task BlockWaitingForSharedStateIncrementAsync(CancellationToken cancellationToken)
+            {
+                SharedState++;
```
LukeButters, line 166: nit: I always wondered if ints need to be interlocked incremented to be safe, so I end up doing to be safe but maybe we don't need to?

source/Halibut.Tests/BiDirectionalTests.cs
```diff
@@ -0,0 +1,182 @@
...
+            });
+        }
+
+        [Test]
+        public async Task ConcurrentOpposingCommunicationsDoNotBlock()
+        {
+            await SetupBiDirectionalPollingClients(async (clientServiceA, clientServiceB) =>
+            {
+                MyEchoService.SharedState = 0;
+                var taskA = clientServiceA.BlockWaitingForSharedStateIncrementAsync().WithCancellation(CancellationToken);
```
LukeButters, line 40: This should prevent random race conditions when we run the tests in parallel thanks!

LukeButters (approved):

source/Halibut.Tests/BiDirectionalTests.cs
zentron: Yeah i think this is probably a good case.
Since this effectively amounts to
`SharedState = SharedState + 1` there is room for threading issues. :)

PR #609, opened 2024-05-14
Including `ConnectionInitializationFailedExceptions` as connection errors

sburmanoctopus:
# Background

# Results

Fixes https://github.com/OctopusDeploy/Issues/issues/... _(optional public issue)_

Fixes https://github.com/OctopusDeploy/ResearchAndDevelopment/issues/... _(optional private issue)_

See [How we use GitHub Issues](https://github.com/OctopusDeploy/Issues/blob/main/docs/CONTRIBUTING.internal.md) (including [this flowchart](https://whimsical.com/r-d-incoming-work-workflow-aug-21-NsDnGQXcwBLwU66a88Zhue) 

## Before

Before

## After

After

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Transport/ListeningServiceErrorConnectionStateFixture.cs
+++ b/source/Halibut.Tests/Transport/ListeningServiceErrorConnectionStateFixture.cs
@@ -4,6 +4,7 @@ using FluentAssertions;
 using Halibut.Tests.Support;
 using Halibut.Tests.Support.TestAttributes;
 using Halibut.Tests.Support.TestCases;
+using Halibut.Tests.TestServices;
 using Halibut.Tests.TestServices.Async;
 using Halibut.TestUtils.Contracts;
 using Halibut.Transport;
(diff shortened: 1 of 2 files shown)


source/Halibut.Tests/Transport/ListeningServiceErrorConnectionStateFixture.cs
```diff
@@ -49,6 +50,28 @@ await using (var clientAndService = await clientAndServiceTestCase.CreateTestCas
                                  portForwarderRef.Value.EnterKillNewAndExistingConnectionsMode();
                              })
                              .Build(CancellationToken))
+            {
```
sburmanoctopus, line 53: It's not obvious from the diff, but this test had to be adjusted.

It was configured to fail during `WithDoSomeActionService`. But rather than create and call an `IDoSomeActionService`, it ended up creating an `echoService`. 

As a result, it would fail with the right exception, but it was going down a different code path and not actually hitting the path this test was intended to hit.

LukeButters (approved):

PR #610, opened 2024-05-15
Improve linux performance by allowing users of Halibut to use TCP NoDelay (so as to be able to opt out of Nagle’s algorithm)

LukeButters:
# Background

Background: https://brooker.co.za/blog/2024/05/09/nagle.html

TLDR; `TCP_NODELAY=true`  results in much faster for RPC performance in linux. TCP_NODELAY disables a TCP setting that makes sense when humans type characters into telnet (buffer the chars before sending because humans are slow) but less sense in halibut since when we send a thing we actually would like it to go over the wire now.

Turns out that setting NoDelay to true significantly reduces the time to make RPC calls on linux. In particular the test 
`OctopusCanSendAndReceiveComplexObjects_WithMultipleDataStreams` runtime drops from `23-25s` to `1-5s` (depending on linux environment, how the test is run).

Teamcity runs show a difference of `22.9s` to `0.5s`

# Results

`HalibutTimeoutsAndLimits` supports `TcpNoDelay` which by default is `false` the old value, but when clients opt in they on linux will see performance improvements.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Octopus.TestPortForwarder/PortForwarder.cs
+++ b/source/Octopus.TestPortForwarder/PortForwarder.cs
@@ -13,8 +13,6 @@ namespace Octopus.TestPortForwarder
 {
     public class PortForwarder : IDisposable
     {
-
-        public static bool NoDelay = true;
         readonly Uri originServer;
         Socket? listeningSocket;
         readonly CancellationTokenSource cancellationTokenSource = new();
@@ -69,7 +67,7 @@ namespace Octopus.TestPortForwarder
             }
 
             listeningSocket ??= new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
-            listeningSocket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.NoDelay, NoDelay);
+            listeningSocket.NoDelay = true;
 
             listeningSocket!.Bind(new IPEndPoint(IPAddress.Loopback, ListeningPort));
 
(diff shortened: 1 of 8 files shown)


source/Octopus.TestPortForwarder/PortForwarder.cs
```diff
@@ -67,6 +67,7 @@ private void Start()
             }
 
             listeningSocket ??= new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
+            listeningSocket.NoDelay = true;
```
LukeButters, line 70: In our test portforwarder we never intended it to by default add buffering.

nathanwoctopusdeploy: When merged, it would be worth bumping the version of the Proxy in Tentacle as well

LukeButters: Good idea

nathanwoctopusdeploy (approved):
LGTM, very nice latency reduction for linux

nathanwoctopusdeploy:
Has this added a bit of flake to integration tests, or was the PR just unlucky?

LukeButters:
it may have since it changes timings! @nathanwoctopusdeploy 
here is the error message:
```
Expected sw.Elapsed to be greater than or equal to 13s, but found 877.8µs
```
Are we expecting something to be slow that is now no longer slow?

pawelpabich (approved):
The change looks ok on the surface. 

It would be good to test this branch version in a real environment and at scale to reduce the chances of our customers hitting some unknow unknows. E.g.

* Dynamic Workers branch build as suggested by @nathanwoctopusdeploy 
* Perf test in Cloud using Armanda with 10k+ tentacle simulators?

PR #612, opened 2024-05-17
Don't try to guess when Halibut is sending control messages in tests.

LukeButters:
# Background

On this [PR](https://github.com/OctopusDeploy/Halibut/pull/610) we are finding that `PauseOnPollingTentacleSendingNextControlMessage_ShouldNotHangForEver` sometimes fails because the port forwarder is observing the NEXT control message is arriving to the port forwarder close enough to the Response message they are considered a single "thing". This confuses the test which relies on finding small packets to know when to pause.

This improves on that rather than make assumption and have deep dependencies on what the code is doing, we (in test only) allow the test to observe when control messages are being sent. This allows the test to pause the port forwarder at exactly the right time without guessing.

This introduces a IControlMessageObserver which allows observation of what is happening with control messages. The halibut builder to set this is marked internal which should prevent surprise uses of this (although if anyone did do this they would probably know what their in for :D )

Using this along with existing byte counting streams we can fix the test by:
* Delaying sending of the "NEXT" control message
* until we know the client has received all bytes. 

An alternative would be to man in the middle ourselves, but then we are just parsing the halibut protocol and depending on how it exactly works, also this becomes tricky since the port forwarder allows pausing on partial SSL frames but if we de-crypted the SSL we wouldn't be able to do that type of pausing.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Support/FuncControlMessageObserver.cs
+++ b/source/Halibut.Tests/Support/FuncControlMessageObserver.cs
@@ -1,0 +1,47 @@
+﻿// Copyright 2012-2013 Octopus Deploy Pty. Ltd.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//   http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using Halibut.Transport.Observability;
+
+namespace Halibut.Tests.Support
+{
+    public class FuncControlMessageObserver : IControlMessageObserver
+    {
+        public Action<string> BeforeSendingControlMessageAction = (_) => { };
+        public Action<string> FinishSendingControlMessageAction = (_) => { };
+        public Action WaitingForControlMessageAction = () => { };
+        public Action<string> ReceivedControlMessageAction = (_) => { };
+
+        void IControlMessageObserver.BeforeSendingControlMessage(string controlMessage)
+        {
+            BeforeSendingControlMessageAction(controlMessage);
+        }
+
+        void IControlMessageObserver.FinishSendingControlMessage(string controlMessage)
+        {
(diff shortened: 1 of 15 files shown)


source/Halibut/HalibutRuntimeBuilder.cs
```diff
@@ -106,6 +107,12 @@ internal HalibutRuntimeBuilder WithPollingReconnectRetryPolicy(Func<RetryPolicy>
             return this;
         }
 
+        internal HalibutRuntimeBuilder WithControlMessageObserver(IControlMessageObserver controlMessageObserver)
```
LukeButters, line 110: internal since ideally clients would not call this (if they ever did we would have no hesitation to break compatibility ie we wont support this observer outside of our testing needs)

source/Halibut.Tests/Timeouts/NextAndProceedControlMessageTimeoutsFixture.cs
```diff
@@ -20,53 +16,47 @@ namespace Halibut.Tests.Timeouts
...
-                // --> NEXT sent and pump paused here.
-                
-                var sw = Stopwatch.StartNew();
-                // Service must detect that it can't get a PROCEED back in time and so must decide to abandon and then try again within
-                // its retry policy.
+                shouldPausePumpOnNextNextControlMessage = true;
+                // When the port forwarder is paused the stopwatch is started,
+                // we time the length it takes for recovery on that paused connection
                 await echo.SayHelloAsync(Some.RandomAsciiStringOfLength(2000));
+                await echo.SayHelloAsync(Some.RandomAsciiStringOfLength(2000)); // Do a second call just in case we missed the NEXT control message
```
rhysparry, line 53: Why might we miss the NEXT control message?

rhysparry (approved):
LGTM

source/Halibut.Tests/Support/FuncControlMessageObserver.cs
```diff
@@ -0,0 +1,47 @@
+﻿// Copyright 2012-2013 Octopus Deploy Pty. Ltd.
```
nathanwoctopusdeploy, line 1: ??

source/Halibut.Tests/Timeouts/NextAndProceedControlMessageTimeoutsFixture.cs
LukeButters: because the polling endpoint comes back and says "NEXT" please before we line up a request for the polling tentacle.

PR #626, opened 2025-02-17
Remove custom dotnet-install.ps1

rhysparry:
# Background

When downloading the .NET SDK during the build it would fail if the system did not enable TLS 1.2 by default. This is generally the case for older Windows versions.

This PR aims to move the modification into the build bootstrapper. This will mean that upstream fixes to the dotnet installer will be available immediately.

## Why not TLS 1.3?

Prior to .NET Framework 4.8, TLS 1.3 was not available. Supporting TLS 1.3 would require special-casing these older versions.

# Results

- Removes the custom `dotnet-install.ps1` file
- Restores `dotnet-install.ps1` usage in `build.ps1`
- Adds customization to `build.ps1` to ensure TLS1.2 is enabled when the download script attempts to download the .NET SDK.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


build.ps1
```diff
@@ -61,9 +57,13 @@ else {
     # Install by channel or version
     $DotNetDirectory = "$TempDirectory\dotnet-win"
     if (!(Test-Path variable:DotNetVersion)) {
-        ExecSafe { & powershell $DotNetInstallFile -InstallDir $DotNetDirectory -Channel $DotNetChannel -NoPath }
+        # Octopus Modification to ensure TLS1.2 is enabled within the script
```
LukeButters, line 60: Neat so we just set the protocol then call the script. That does seem like a better solution.

LukeButters (approved):

PR #662, opened 2025-07-21
Allow users (of halibut) to define the polling re-connect policy

LukeButters:
# Background

Halibut previously made assumptions around how users would want polling services to re-connected. Which meant that for local development we would need to wait for up to 2 minutes for tentacle or simulated tentacle to re-connect.

This change allows clients of Halibut to change the reconnect policy. For example in local dev this would allow us to define a faster retry policy avoiding that 2min delay that currently applies.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


sburmanoctopus (approved):
LGTM

PR #674, opened 2025-08-25
Fix a bug where the watcher would not terminate if it could not subscribe to Redis.

LukeButters:
# Background

A bug existed where if the watcher could not subscribe to redis, it would hang until the passed in CT was cancelled. Instead it should have bailed out since if it can't sub to redis then it can't get heart beats so as far as we are concerned the other node may have disconnected (although really we disconnected).

Also adds a test to show that the request processing can exceed the heart beat timeout.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


david-staniec-octopus (approved):
Discussed approach and impact in sync, LGTM.

PR #676, opened 2025-09-04
Fix a bug in message stream wrapping, where the streams are not disposed in time.

LukeButters:
# Background

The bug is that streams did not get disposed (closed) before we read the resulting bytes which meant sometimes bytes of the message would be missing.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Queue/QueueMessageSerializer.cs
+++ b/source/Halibut/Queue/QueueMessageSerializer.cs
@@ -1,11 +1,8 @@
 
 using System;
 using System.Collections.Generic;
-using System.Globalization;
 using System.IO;
-using System.IO.Compression;
 using System.Text;
-using System.Threading.Tasks;
 using Halibut.Queue.MessageStreamWrapping;
 using Halibut.Transport.Protocol;
 using Halibut.Util;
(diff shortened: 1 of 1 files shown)


source/Halibut/Queue/QueueMessageSerializer.cs
```diff
@@ -37,38 +34,49 @@ public QueueMessageSerializer(Func<StreamCapturingJsonSerializer> createStreamCa
...
 
             return (ms.ToArray(), dataStreams);
         }
 
+        public static Stream WrapInMessageSerialisationStreams(MessageStreamWrappers messageStreamWrappers, Stream stream, DisposableCollection disposables)
+        {
+            foreach (var streamer in messageStreamWrappers.Wrappers)
+            {
+                var wrappedStream = streamer.WrapMessageSerialisationStream(stream);
+                if (!ReferenceEquals(wrappedStream, stream))
```
LukeButters, line 64: This is done since it makes it easier to feature toggle stream wrappers.

acodrington (approved):
This PR diff is really hard to read, but the changes are all good (once I deciphered them!)

PR #684, opened 2025-09-12
Don't call RegisterType twice for each RPC and with the redis queue it was 4 times!

LukeButters:
# Background

We would re-register the types each time the serializer was built, since that is built with this lambda each time we serialise or deserialize a message:

image

With these changes we just register the types once, rather than over and over again.

This as a hot spot was somehow found with a profiler.

With this change we instead call register once for the Request and Response types rather than each time we serialise or deserialise a message (including in and out of the queue). This should have no impact other than improving performance.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Transport/Protocol/RegisteredSerializationBinder.cs
+++ b/source/Halibut/Transport/Protocol/RegisteredSerializationBinder.cs
@@ -23,7 +23,6 @@ namespace Halibut.Transport.Protocol
         internal RegisteredSerializationBinder(ITypeRegistry typeRegistry)
         {
             this.typeRegistry = typeRegistry;
-            
         }
 
         public Type BindToType(string? assemblyName, string typeName)
--- a/source/Halibut/Transport/Protocol/RegisteredSerializationBinder.cs
+++ b/source/Halibut/Transport/Protocol/RegisteredSerializationBinder.cs
@@ -16,7 +16,10 @@ namespace Halibut.Transport.Protocol
 
         public static void AddProtocolTypesToTypeRegistry(ITypeRegistry typeRegistry)
         {
-            foreach (var protocolType in protocolTypes) typeRegistry.RegisterType(protocolType, protocolType.Name, true);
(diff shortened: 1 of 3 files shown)


source/Halibut/Transport/Protocol/RegisteredSerializationBinder.cs
```diff
@@ -6,17 +6,23 @@ namespace Halibut.Transport.Protocol
...
+        static readonly Type[] protocolTypes = { typeof(ResponseMessage), typeof(RequestMessage) };
         readonly ITypeRegistry typeRegistry;
 
         public RegisteredSerializationBinder() : this(new TypeRegistry())
         {
-        } // kept for backwards compatibility.
+            AddProtocolTupesToTypeRegistry(this.typeRegistry);
+        }
+
+        public static void AddProtocolTupesToTypeRegistry(ITypeRegistry typeRegistry)
```
rhysparry, line 17: nit: should be `AddProtocolTypesToTypeRegistry` (i.e. Types not Tupes)

source/Halibut/Transport/Protocol/RegisteredSerializationBinder.cs
```diff
@@ -6,17 +6,23 @@ namespace Halibut.Transport.Protocol
...
+        {
+            foreach (var protocolType in protocolTypes) typeRegistry.RegisterType(protocolType, protocolType.Name, true);
```
rhysparry, line 19: nit: if we're moving this anyway, let's format it with curly braces across multiple lines.

rhysparry (approved):
A couple of nits. The typo in the method name being actually somewhat important.

source/Halibut/Transport/Protocol/RegisteredSerializationBinder.cs
LukeButters: ok

PR #685, opened 2025-09-14
Dont create ConcurrentBag within CancelOnDisposeCancellationToken unless required

LukeButters:
# Background

The profiler kept showing that within the ConcurrentBag was a place that time is spent, this PR removes that.
Its likely it showed up because of how many `CancelOnDisposeCancellationToken`s are created for each request in the Redis PRQ.

This should not have a functional change on the queue.

I don't think the resource drop will be in anyway significant, but if the profiler can stop showing this as a hot spot that might make it easier to find what is taking time.

# How to review this PR

Did I use the test and set correctly?

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Util/CancelOnDisposeCancellationToken.cs
+++ b/source/Halibut/Util/CancelOnDisposeCancellationToken.cs
@@ -34,7 +34,7 @@ namespace Halibut.Util
         readonly CancellationTokenSource cancellationTokenSource;
         bool disposed;
 
-        readonly ConcurrentBag<Task> tasks = new();
+        ConcurrentBag<Task>? tasks = null;
         
         public CancelOnDisposeCancellationToken(params CancellationToken[] token)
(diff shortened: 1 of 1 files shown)


source/Halibut/Util/CancelOnDisposeCancellationToken.cs
```diff
@@ -86,14 +89,20 @@ public void CancelAfter(TimeSpan timeSpan)
...
         {
+            if (tasks == null)
+            {
+                var newBag = new ConcurrentBag<Task>();
+                Interlocked.CompareExchange(ref tasks, newBag, null);
+            }
             foreach (var task in tasksUsingToken)
             {
-                tasks.Add(task);
+                tasks!.Add(task);
```
rhysparry, line 105: nit: It would be great if we could avoid the `!` here, although I think an assertion in the `if` block might be the only realistic way.

PR #686, opened 2025-09-15
Reduce the number of calls made to Redis when the RPC is fast.

LukeButters:
# Background

To reduce CPU usage mainly of the Halibut Queue (than Redis), with this PR we avoid as many calls as possible to Redis for fast RPC calls. Where fast is in the order of seconds or less.

With this change, for fast RPCs we:
- Don't send heart beats.
- Don't subscribe to hearts.
- Don't subscribe to the cancellation channel.

Our testing shows a reduction of about 10% in CPU usage with this type of change.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Queue/Redis/RedisPendingRequestQueueFixture.cs
+++ b/source/Halibut.Tests/Queue/Redis/RedisPendingRequestQueueFixture.cs
@@ -10,6 +10,7 @@ using Halibut.Logging;
 using Halibut.Queue;
 using Halibut.Queue.MessageStreamWrapping;
 using Halibut.Queue.Redis;
+using Halibut.Queue.Redis.Cancellation;
 using Halibut.Queue.Redis.Exceptions;
 using Halibut.Queue.Redis.MessageStorage;
 using Halibut.Queue.Redis.NodeHeartBeat;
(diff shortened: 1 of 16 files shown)


source/Halibut.Tests/Queue/Redis/RedisPendingRequestQueueFixture.cs
```diff
@@ -120,6 +121,61 @@ public async Task FullSendAndReceiveShouldWork()
...
         }
         
+        [Test]
+        public async Task WhenTheRequestIsQuicklyProcessedTheNumberOfRequestToRedisShouldBeLimited()
+        {
+            // Arrange
+            var endpoint = new Uri("poll://" + Guid.NewGuid());
+            await using var redisFacade = RedisFacadeBuilder.CreateRedisFacade();
+            var baseRedisTransport = new HalibutRedisTransport(redisFacade);
+            var callCountingTransport = new CallCountingHalibutRedisTransport(baseRedisTransport);
```
rhysparry, line 131: Where is this class?

rhysparry (commented):
I think you might be missing a file.

PR #687, opened 2025-09-15
Concurrent collections in Dotnet are not as safe as they seem

LukeButters:
# Background

Concurrent dictionary appears to be surprising in that `Func` passed to it are not executed in what one would consider to be "thread safe", so resort to manual locking.

ConcurrentBag does interesting things with thread locals, which has lead to strange memory leaks:
- https://github.com/autofac/Autofac/pull/1257

so lets just not use it.

For background read: https://octopusdeploy.slack.com/archives/CMESRJ1C3/p1703022896511849

This change is not because we have been seeing issues, but is done preemptively to prevent issues. 

The only non test change is the one done to the Redis PRQ.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Util/CancelOnDisposeCancellationToken.cs
+++ b/source/Halibut/Util/CancelOnDisposeCancellationToken.cs
@@ -34,7 +34,7 @@ namespace Halibut.Util
         readonly CancellationTokenSource cancellationTokenSource;
         bool disposed;
 
-        ConcurrentBag<Task>? tasks = null;
+        readonly List<Task> tasks = new();
         
         public CancelOnDisposeCancellationToken(params CancellationToken[] token)
             : this(CancellationTokenSource.CreateLinkedTokenSource(token))
@@ -64,7 +64,7 @@ namespace Halibut.Util
             await Try.IgnoringError(async () => await CancelAsync());
 
             // Wait for any tasks that are using the token, before disposal
-            if (tasks != null)
+            if (tasks.Count > 0)
             {
                 await Task.WhenAll(tasks.Select(t => Try.IgnoringError(() => t)));
             }
@@ -95,14 +95,12 @@ namespace Halibut.Util
         /// <param name="tasksUsingToken"></param>
         public void AwaitTasksBeforeCTSDispose(params Task[] tasksUsingToken)
         {
-            if (tasks == null)
+            lock (tasks)
             {
-                var newBag = new ConcurrentBag<Task>();
-                Interlocked.CompareExchange(ref tasks, newBag, null);
-            }
-            foreach (var task in tasksUsingToken)
-            {
-                tasks!.Add(task);
+                foreach (var task in tasksUsingToken)
+                {
+                    tasks.Add(task);
+                }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


source/Halibut/Util/CancelOnDisposeCancellationToken.cs
```diff
@@ -34,7 +34,7 @@ public sealed class CancelOnDisposeCancellationToken : IAsyncDisposable
         readonly CancellationTokenSource cancellationTokenSource;
         bool disposed;
 
-        ConcurrentBag<Task>? tasks = null;
+        readonly List<Task> tasks = new();
```
LukeButters, line 37: Creating a list should be well fast enough since the ctor for `List()` appears to just use a static to initialise itself. 

I think ConcurrentBag is relatively expensive to init because of all the thread local stuff it wants to create.

source/Halibut/Util/CancelOnDisposeCancellationToken.cs
```diff
@@ -64,7 +64,7 @@ public async ValueTask DisposeAsync()
             await Try.IgnoringError(async () => await CancelAsync());
 
             // Wait for any tasks that are using the token, before disposal
-            if (tasks != null)
+            if (tasks.Count > 0)
```
LukeButters, line 67: By the time we call dispose we are assuming (and this holds true in the Redis PRQ) that we will not be making further changes to the set of tasks to wait on.

Intentionally doing a task.Count check to avoid doing a bunch if "task stuff" we often wont need to.

source/Halibut/Util/CancelOnDisposeCancellationToken.cs
```diff
@@ -95,14 +95,12 @@ public void CancelAfter(TimeSpan timeSpan)
...
-            if (tasks == null)
+            lock (tasks)
             {
-                var newBag = new ConcurrentBag<Task>();
-                Interlocked.CompareExchange(ref tasks, newBag, null);
-            }
-            foreach (var task in tasksUsingToken)
-            {
-                tasks!.Add(task);
+                foreach (var task in tasksUsingToken)
```
david-staniec-octopus, line 100: AddRange to avoid foreach?

source/Halibut/Util/CancelOnDisposeCancellationToken.cs
david-staniec-octopus: From memory, locking on `tasks` in `AwaitTasksBeforeCTSDispose` won't lock access to the list contents. Was this handled by the `Interlocked.CompareExchange` call previously?

source/Halibut/Util/CancelOnDisposeCancellationToken.cs
```diff
@@ -95,14 +95,12 @@ public void CancelAfter(TimeSpan timeSpan)
         /// <param name="tasksUsingToken"></param>
```
david-staniec-octopus, line 95: Comment for:
```
/// Thread safe: uses Interlocked.CompareExchange for initialization and ConcurrentBag for storage.
```

no longer applies. Let's update that.

source/Halibut/Util/CancelOnDisposeCancellationToken.cs
LukeButters: done in other PR

PR #695, opened 2025-09-24
Collected requests do not fail in a connecting state

LukeButters:
# Background

If polling tentacle has collected the request then we are NOT in the state of connecting. The tentacle has potentially received and processed the request, we don't know and:

-> we can't know.

image

# Results

It is not clear if the code paths changed can ever be reached! At least I could not see how to craft a test that reaches them. In any case if they can be reached then this fixes cases where TentacleClient is wrongly assuming the request has not been collected and so is assuming the request has not started.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut/Queue/Redis/RedisPendingRequest.cs
+++ b/source/Halibut/Queue/Redis/RedisPendingRequest.cs
@@ -145,8 +145,7 @@ namespace Halibut.Queue.Redis
                         log.Write(EventType.MessageExchange, "Request {0} was cancelled before a response was received", request);
                         SetResponseNoLock(ResponseMessage.FromException(
                                 request,
-                                new TimeoutException("A request was sent to a polling endpoint, the polling endpoint collected it but the request was cancelled before the polling endpoint responded."),
-                                ConnectionState.Connecting),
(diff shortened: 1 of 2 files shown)


source/Halibut/Queue/Redis/RedisPendingRequest.cs
```diff
@@ -145,8 +145,7 @@ public async Task WaitUntilComplete(Func<Task> checkIfPendingRequestWasCollected
                         log.Write(EventType.MessageExchange, "Request {0} was cancelled before a response was received", request);
                         SetResponseNoLock(ResponseMessage.FromException(
                                 request,
-                                new TimeoutException("A request was sent to a polling endpoint, the polling endpoint collected it but the request was cancelled before the polling endpoint responded."),
-                                ConnectionState.Connecting),
+                                new TimeoutException("A request was sent to a polling endpoint, the polling endpoint collected it but the request was cancelled before the polling endpoint responded.")),
                             requestWasCollected: false);
```
sburmanoctopus, line 149: Just wondering, we have established that the request has been collected (and are now waiting for a response). So why is the variable `requestWasCollected` false?

sburmanoctopus (approved):
LGTM

PR #705, opened 2026-03-26
Fix a bug where we would not flush messages before waiting for a response

LukeButters:
# Background

ref CLOUDPT-11160
ref EFT-3141

Fixes a bug where if the RequestMessage has no data stream, we would not flush it and so we could be in a situation where by both ends are waiting for data to be sent. Since the sender has not flushed the request but is waiting for a response, while the receiver is waiting for the request to be sent.

Shout out to @sahanocto and claude for finding the issue.

# Results

## Before

Test failed

## After

Test passes.
# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites

### DIFF
--- a/source/Halibut.Tests/Support/Streams/FlushBufferedStream.cs
+++ b/source/Halibut.Tests/Support/Streams/FlushBufferedStream.cs
@@ -1,0 +1,81 @@
+#nullable enable
+using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Halibut.Tests.Support.Streams
+{
+    /// <summary>
+    /// A stream that buffers all writes in memory and only forwards them to the underlying
+    /// stream when Flush or FlushAsync is called.
+    /// </summary>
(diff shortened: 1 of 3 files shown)


source/Halibut.Tests/Support/Streams/FlushBufferedStream.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Halibut.Tests.Support.Streams
+{
+    /// <summary>
+    /// A stream that buffers all writes in memory and only forwards them to the underlying
+    /// stream when Flush or FlushAsync is called.
+    /// </summary>
+    public class FlushBufferedStream : DelegateStreamBase
```
rhysparry: Is it worth including 'Unbounded' or 'Test' in the name here? I wouldn't want to see this class go wandering outside of tests.

rhysparry (approved):
✅ LGTM

PR #706, opened 2026-03-26
Always use system TLS defaults

### LINKED ISSUE #716: Always use OS defaults for TLS protocol negotiation
Halibut originally hardcoded specific TLS protocol versions to use during negotiation.

This was then replaced with a configurable set, including the ability to delegate configuration to the underlying operating system.

To reduce complexity and to ensure Operating System configuration is respected, we will remove internal configuration and delegate to the Operating System.

This will require that the host Operating System has the necessary TLS protocol versions enabled by default.

rhysparry:
# Background

While we were rolling out the change to adapt Halibut configuration to use system defaults we provided a mechanism for consumers of Halibut (i.e. Tentacle and Octopus) to override these defaults if necessary.

Now that we have confirmed that switching to system defaults has not caused issues, and has improved security (by allowing older TLS protocols to be disabled by the Operating System) we are removing the ability to configure these protocols directly.

# Results

- Removes the `ISslConfigurationProvider` and its provided implementations
- Exposes the `SslProtocols` value used internally via `SslConfiguration.SupportedProtocols`. This value has been set to `None`, which "Allows the operating system to choose the best protocol to use, and to block protocols that are not secure." (see [docs](https://learn.microsoft.com/en-us/dotnet/api/system.security.authentication.sslprotocols?view=net-8.0))
- Updates usages of the `ISslConfigurationProvider` to access this static property.
- Removes the ability to supply SSL configuration within the `HalibutRuntimeBuilder`
- Resolves EFT-157
- Resolves https://github.com/OctopusDeploy/Halibut/issues/716

## Before

- TLS Protocols supported by Halibut could be overridden by the application.

## After

- TLS Protocols supported by Halibut are configured at the Operating System level.

# How to review this PR

Quality :heavy_check_mark:

# Pre-requisites


source/Halibut.Tests.DotMemory/MemoryFixture.cs
```diff
@@ -142,16 +160,33 @@ static HalibutRuntime RunServer(X509Certificate2 serverCertificate, out int port
...
-
             return server;
         }
 
+        // pollingServer intentionally uses Certificates.TentacleListening rather than
+        // Certificates.Octopus (which server uses). This keeps the two certificates in distinct
+        // TLS roles within this process: Octopus is used only as a TLS server cert (by server),
+        // and TentacleListening is used only as a TLS client cert (here, and in RunListeningClient).
+        // Using the same cert in both roles would trigger an SChannel session-cache collision on
+        // Windows with SslProtocols.None (see declaration comment above).
```
LukeButters: Remove this comment it is explained above.

source/Halibut.Tests.DotMemory/MemoryFixture.cs
```diff
@@ -142,16 +160,33 @@ static HalibutRuntime RunServer(X509Certificate2 serverCertificate, out int port
                 .WithLogFactory(new TestContextLogFactory("client", LogLevel.Info))
                 .Build();
 
-            //set up listening  
+            // Trust the listening tentacle certificate for inbound connections.
+            // This runtime only accepts connections — it never makes outbound polling connections —
+            // keeping it in a pure TLS server role (see declaration comment above).
             server.Trust(Certificates.TentacleListeningPublicThumbprint);
```
LukeButters, line 163: Comment here probs doesn't help

source/Halibut.Tests/Support/BackwardsCompatibility/LatestClientAndPreviousServiceVersionBuilder.cs
```diff
@@ -44,17 +46,23 @@ public LatestClientAndPreviousServiceVersionBuilder WithCertificates(
...
+            var tmpDirectory = TestCertificates.NewTmpDirectoryIfNeeded();
+            var clientCert = TestCertificates.CertFor(CertAndThumbprint.Octopus, tmpDirectory);
+            return new LatestClientAndPreviousServiceVersionBuilder(ServiceConnectionType.Polling, CertAndThumbprint.TentaclePolling, clientCert) { tmpDirectory = tmpDirectory };
         }
 
         public static LatestClientAndPreviousServiceVersionBuilder WithPollingOverWebSocketService()
         {
-            return new LatestClientAndPreviousServiceVersionBuilder(ServiceConnectionType.PollingOverWebSocket, CertAndThumbprint.TentaclePolling);
+            var tmpDirectory = TestCertificates.NewTmpDirectoryIfNeeded();
+            var clientCert = TestCertificates.CertFor(CertAndThumbprint.Octopus, tmpDirectory);
```
LukeButters: Why do we need to do all of this now?

source/Halibut.Tests/Support/BackwardsCompatibility/SchannelProbeBinaryRunner.cs
```diff
@@ -0,0 +1,128 @@
...
+using System.Text.RegularExpressions;
+using System.Threading;
+using System.Threading.Tasks;
+using CliWrap;
+using Halibut.Logging;
+using Serilog;
+
+namespace Halibut.Tests.Support.BackwardsCompatibility
+{
+    public class SchannelProbeBinaryRunner
```
LukeButters: It is worth explaining what this is for.

source/Halibut.Tests/Support/TestCertificates.cs
```diff
@@ -0,0 +1,51 @@
...
+            return null;
+#endif
+        }
+
+        /// <summary>
+        /// On .NET Framework, generates a fresh unique self-signed certificate into <paramref name="tmpDirectory"/>.
+        /// On other frameworks, returns the supplied <paramref name="staticCert"/> so static certificates are
+        /// shared (enabling TLS session resumption).
+        /// </summary>
+        public static CertAndThumbprint CertFor(CertAndThumbprint staticCert, TmpDirectory? tmpDirectory)
```
LukeButters: Maybe maybe make CertAndThumbprint disposable and then if in net4.8 we create a new tmp dir. Build the cert in it and the disposable cleans up the temp dir. Then we don't need that TmpDir? thing to be exposed outside of this.

The return could also be a CertAndThumbprintHolder or GeneratedCertAndThumbprint if we want a new type that is disposable

or return a tuple <CertAndThumbprint, IDisposable>

source/Halibut.Tests/TestServices/Async/IAsyncClientSayHelloService.cs
```diff
@@ -0,0 +1,9 @@
+using System.Threading.Tasks;
+
+namespace Halibut.Tests.TestServices.Async
+{
+    public interface IAsyncClientSayHelloService
+    {
+        Task<string> SayHelloAsync(string name);
```
LukeButters: Im am surprised the echo service isn't enough, why make this new service?

source/Halibut.Tests/SchannelSessionCacheFixture.cs
```diff
@@ -0,0 +1,172 @@
...
+            // triggering an SChannel session-cache collision if process isolation were absent.
+            var sharedTentacleCert = CertificateGenerator.GenerateSelfSignedCertificate(tmpDirectory.FullPath);
+
+            var octopus = new HalibutRuntimeBuilder()
+                .WithServerCertificate(octopusCert.Certificate2)
+                .WithLogFactory(new TestContextLogCreator("Octopus", Logging.LogLevel.Trace).ToCachingLogFactory())
+                .WithHalibutTimeoutsAndLimits(new HalibutTimeoutsAndLimitsForTestsBuilder().Build())
+                .Build();
+
+            await using var _ = new AsyncDisposableAction(async () => await octopus.DisposeAsync());
```
LukeButters: This is needed?

source/Halibut.TestUtils.CompatBinary.SchannelProbe/Program.cs
```diff
@@ -0,0 +1,137 @@
...
+using System.IO;
+using System.Security.Cryptography.X509Certificates;
+using System.Threading;
+using System.Threading.Tasks;
+using Halibut;
+using Halibut.Diagnostics;
+using Halibut.ServiceModel;
+using Halibut.TestUtils.Contracts;
+
+namespace Halibut.TestUtils.SampleProgram.SchannelProbe
```
LukeButters: Why don't the existing compat bins that run halibut work for the test?

source/Halibut.Tests/SchannelSessionCacheFixture.cs
```diff
@@ -0,0 +1,172 @@
...
+    /// cache (Windows). Running the tentacle in a separate process avoids this because the
+    /// SChannel session cache is per-process.
+    ///
+    /// These tests verify that two separate processes — one listening tentacle and one polling
+    /// tentacle — both using the same certificate and both connecting via localhost, can
+    /// simultaneously communicate successfully with an in-process Octopus server.
+    /// </summary>
+    [TestFixture]
+    [NonParallelizable]
+    public class SchannelSessionCacheFixture : BaseTest
```
LukeButters: Are these tests valuable do they get tested as part of the normal backwards compat tests?

LukeButters (approved):
I think we can drop the SchannelSessionCacheFixture, since it might not be testing anything of value.

source/Halibut.Tests/Support/BackwardsCompatibility/SchannelProbeBinaryRunner.cs
rhysparry: It's been removed

source/Halibut.Tests/Support/TestCertificates.cs
rhysparry: I've refactored this

source/Halibut.Tests/TestServices/Async/IAsyncClientSayHelloService.cs
rhysparry: Removed

source/Halibut.Tests/SchannelSessionCacheFixture.cs
rhysparry: Removed

source/Halibut.TestUtils.CompatBinary.SchannelProbe/Program.cs
rhysparry: Removed

source/Halibut.Tests/SchannelSessionCacheFixture.cs
rhysparry: Removed

PR #717, opened 2026-08-17
Add ability to observe the number of connections halibut opens for polling connections

pawelpabich:
Related server PR https://github.com/OctopusDeploy/OctopusDeploy/pull/46535

### DIFF
--- a/source/Halibut/Transport/ActiveTcpConnectionsLimiter.cs
+++ b/source/Halibut/Transport/ActiveTcpConnectionsLimiter.cs
@@ -6,10 +6,19 @@ using Halibut.Exceptions;
 
 namespace Halibut.Transport
 {
+    public interface IActiveTcpConnectionLease : IDisposable
+    {
+        /// <summary>
+        /// The number of active TCP connections for the leased subscription, as at the point the lease was
+        /// created. After Dispose() is called, this reflects the count immediately after this connection
+        /// was released.
+        /// </summary>
+        int CurrentCount { get; }
+    }
+
     public interface IActiveTcpConnectionsLimiter
     {
-        IDisposable LeaseActiveTcpConnection(Uri subscriptionId);
-
+        IActiveTcpConnectionLease LeaseActiveTcpConnection(Uri subscriptionId);
     }
 
     public class ActiveTcpConnectionsLimiter : IActiveTcpConnectionsLimiter
@@ -23,33 +32,29 @@ namespace Halibut.Transport
             this.timeoutsAndLimits = timeoutsAndLimits;
         }
 
-        public IDisposable LeaseActiveTcpConnection(Uri subscriptionId)
+        public IActiveTcpConnectionLease LeaseActiveTcpConnection(Uri subscriptionId)
         {
-            //if there is no limit, then we return a NoOp lease (which doesn't limit anything)
+            //if there is no limit, then we still count the connection (callers rely on the resulting count),
+            //we just never reject it
             if (!timeoutsAndLimits.MaximumActiveTcpConnectionsPerPollingSubscription.HasValue)
             {
-                return CreateUnlimitedLease();
+                return CreateUnlimitedLease(subscriptionId);
             }
 
             return new LimitingAuthorizedTcpConnectionLease(subscriptionId, activeConnectionCountPerSubscriptionId, timeoutsAndLimits.MaximumActiveTcpConnectionsPerPollingSubscription.Value);
         }
 
-        IDisposable CreateUnlimitedLease()
+        IActiveTcpConnectionLease CreateUnlimitedLease(Uri subscriptionId)
         {
-            return new UnlimitedAuthorizedTcpConnectionLease();
+            return new LimitingAuthorizedTcpConnectionLease(subscriptionId, activeConnectionCountPerSubscriptionId, int.MaxValue);
         }
 
-        class UnlimitedAuthorizedTcpConnectionLease : IDisposable
-        {
-            public void Dispose()
-            {
-            }
-        }
-
-        class LimitingAuthorizedTcpConnectionLease : IDisposable
+        class LimitingAuthorizedTcpConnectionLease : IActiveTcpConnectionLease
         {
             readonly Uri subscriptionId;
             readonly Dictionary<Uri, StrongBox<int>> activeConnectionCountPerSubscriptionId;
+
+            public int CurrentCount { get; private set; }
 
             public LimitingAuthorizedTcpConnectionLease(Uri subscriptionId, Dictionary<Uri, StrongBox<int>> activeConnectionCountPerSubscriptionId, int maximumAcceptedTcpConnectionsPerThumbprint)
             {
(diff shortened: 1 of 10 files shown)


source/Halibut/Transport/Protocol/MessageExchangeProtocol.cs
```diff
@@ -106,32 +109,30 @@ public async Task ExchangeAsServerAsync(Func<RequestMessage, Task<ResponseMessag
...
-                {
-                    case RemoteIdentityType.Client:
-                        await ProcessClientRequestsAsync(incomingRequestProcessor, cancellationToken);
-                        break;
-                    case RemoteIdentityType.Subscriber:
+                case RemoteIdentityType.Client:
+                    await IdentifyAsServerAsync(identity, cancellationToken);
+                    await ProcessClientRequestsAsync(incomingRequestProcessor, cancellationToken);
+                    break;
+                case RemoteIdentityType.Subscriber:
```
pawelpabich, line 115: I've moved the connection limit enforcement closer to the code that it applies to. 
1. There is no need to perform work for other types of remote identities
2. We were executing `IdentifyAsServerAsync` even for the default switch case, which results in an exception. This shows how easy it is to make a mistake if code that is case-specific is applied to all cases.

source/Halibut/Transport/Observability/IConnectionsObserver.cs
```diff
@@ -16,8 +18,20 @@ public interface IConnectionsObserver
...
+        /// A polling subscriber's connections' count has changed
+        /// </summary>
+        /// <param name="subscriptionId">The polling subscriber's subscription id.</param>
+        /// <param name="previousCount">
+        /// The number of active TCP connections for this subscriptionId immediately before the change
+        /// </param>
+        /// <param name="currentCount">
+        /// The number of active TCP connections for this subscriptionId immediately after the change
+        /// </param>
+        public void ConnectionsCountChangedFor(Uri subscriptionId, int previousCount, int currentCount);
```
pawelpabich, line 35: I considrerd a pair of `ConnectionOpendFor` and `ConnectionClosedFor` methods but then we would keep yet another copy of <SubscriptionId, int> in the metric producer in the server.  This might matter when the Server needs to deal with 20k of tentacles.  Thoughts?

LukeButters: Make it simpler for the caller is probably the best approach.

source/Halibut/Transport/ActiveTcpConnectionsLimiter.cs
```diff
@@ -1,61 +1,58 @@
...
-
-        IDisposable CreateUnlimitedLease();
     }
 
     public class ActiveTcpConnectionsLimiter : IActiveTcpConnectionsLimiter
     {
         readonly HalibutTimeoutsAndLimits timeoutsAndLimits;
+        readonly IConnectionsObserver connectionsObserver;
 
         Dictionary<Uri, StrongBox<int>> activeConnectionCountPerSubscriptionId = new();
```
LukeButters, line 20: This looks like the information you want.

Maybe have a method that will give you back a copy of this dictionary upon request.

source/Halibut/Transport/ActiveTcpConnectionsLimiter.cs
```diff
@@ -65,17 +62,18 @@ public LimitingAuthorizedTcpConnectionLease(Uri subscriptionId, Dictionary<Uri,
...
-                        //decrement as this connection has been rejected
-                        count.Value--;
-
                         //throw an exception, bailing on the connection
                         throw new ActiveTcpConnectionsExceededException(this.subscriptionId, $"Exceeded the maximum number ({maximumAcceptedTcpConnectionsPerThumbprint}) of active TCP connections for subscription {subscriptionId}");
                     }
+
+                    count.Value++;
+
+                    connectionsObserver.ConnectionsCountChangedFor(subscriptionId, previousCount, count.Value);
```
LukeButters: I guess this means we need to be super careful about what `ConnectionsCountChangedFor` does since slow code here will impact ALL connecting tentacles  one after the other.

source/Halibut/Transport/ActiveTcpConnectionsLimiter.cs
pawelpabich: That would create a tighter coupling between these two.

source/Halibut/Transport/ActiveTcpConnectionsLimiter.cs
pawelpabich: Yes, the same applies to other methods of the observers. The only difference is that previously only a single connection would be blocked whereas here all new connections would be blocked.  Witout the lock, we would get inconsitent results.

source/Halibut/Transport/Observability/IConnectionsObserver.cs
LukeButters: Looking at how this is used, what we have now is fine.

source/Halibut/Transport/Protocol/MessageExchangeProtocol.cs
LukeButters: How was the issue found?

source/Halibut/Transport/ActiveTcpConnectionsLimiter.cs
LukeButters: I think what you are doing with buckets is faster over processing the entire dict each time we want metrics so this is fine.

As for coupling we already have some form, I think the form that works the best for the telemetry we want to collect makes sense. I guess that comes with a "I would be comfortable changing the contract for telemetry callbacks if we had"

source/Halibut/Transport/ActiveTcpConnectionsLimiter.cs
LukeButters: Actually with your implementation,  the call to `ConnectionsCountChangedFor` does not need to be within the lock provided that the values we would pass in are immutable at this point.

I think I would suggest preserving the values and calling `ConnectionsCountChangedFor` outside the lock, so as to prevent coupling where someone depends on being within the lock.

I am happy for the PR to go in with the lock around that call but would prefer it not be around the call. We can always move the call to be within the lock later but going the other way is a little more tricky.

LukeButters (approved):
The change: lgtm

I think though it would make sense to have some tests added before merging, could you add some before merging?

source/Halibut/Transport/Protocol/MessageExchangeProtocol.cs
pawelpabich: I found it by accident

source/Halibut/Transport/ActiveTcpConnectionsLimiter.cs
pawelpabich: Without lock we will be exposed to suprising short term values. E.g. 
(1, 2) might be processed before (0, 1) which will result in (-1, 1) -> (0, 1) values in the bucket.  (-1, 1) should not really last for long as the connections are rather long lived. I will move it.

pawelpabich:
Tests added.

PR #719, opened 2026-09-13
Don't log to error when a remote client connects and then closes the connection early on (e.g. during auth) since that could any client doing any number of things.

LukeButters:
## Summary

If a client connects but bails out before completing the auth dance we no longer log that as an error since it could be literally anything.

Fixes: LEV-1837

### AI summary.
- Clients (self-host and cloud) can connect and then disconnect, get reset, or otherwise bail out before or during the TLS handshake (e.g. a health check probe, a load balancer, a client giving up on a slow/failed connection attempt) — causing `SslStream.AuthenticateAsServerAsync` to throw `IOException: Received an unexpected EOF or 0 bytes from the transport stream`.
- This was previously falling through to the generic unhandled-error handler in `SecureListener.ExecuteRequest` and logged as an `Error`, spamming logs for what is a normal occurrence rather than a Halibut fault.
- `SecureListener` now wraps `AuthenticateAsServerAsync`, rethrows any `IOException` from it as a new `RemoteFailedToCompleteTlsHandshakeException` (indicating the remote failed to complete authentication), and logs that at `EventType.Diagnostic` (verbose) instead of `ErrorInInitialisation`.

## Test plan
- [x] Added `ClientDisconnectingBeforeCompletingTheTlsHandshakeIsLoggedQuietly` (`SecureListenerFixture`) — connects and disconnects before starting TLS, watched it fail against old behaviour (logged at `ErrorInInitialisation`), then verified it passes with the fix.
- [x] `SecureListenerFixture` + `BadCertificatesTests` suites pass (36 tests), confirming no regression to existing auth-failure logging (bad certs, untrusted thumbprints still log via `ClientDenied`/`AuthenticationException`).


source/Halibut.Tests/Diagnostics/FailuresWhenPollingTentaclesConnectAreLoggedFixture.cs
```diff
@@ -34,21 +34,29 @@ public async Task VeryEarlyOnFailuresAreRecordedAsInitialisationErrors(ClientAnd
...
-                logs.Should().Match(logs => logs.Any(l => l.Type == EventType.ErrorInInitialisation));
+                logs.Should().Match(logs => logs.Any(IsListeningIntialisationFailureLogMessage));
             }
         }
 
+        static bool IsListeningIntialisationFailureLogMessage(LogEvent l)
+        {
+            return l.Type == EventType.ErrorInInitialisation
+                   // Also check for cases where it looks like the client just walked away, maybe nmap scanned us?
+                   || (l.Type == EventType.Diagnostic && l.FormattedMessage.Contains("did not complete the TLS handshake"));
```
LukeButters, line 57: This is the change to the test

source/Halibut/Transport/SecureListener.cs
```diff
@@ -305,13 +305,24 @@ async Task ExecuteRequest(TcpClient client)
...
+                    {
+                        await ssl
+                            .AuthenticateAsServerAsync(
+                                serverCertificate,
+                                true,
+                                SslConfiguration.SupportedProtocols,
+                                false)
+                            .ConfigureAwait(false);
+                    }
+                    catch (IOException ex)
```
sburmanoctopus, line 318: The story is all about when the exception is _"Received an unexpected EOF or 0 bytes from the transport stream."_. 

Is there a way to detect that? Or is **any** `IOException` going to be for this reason?

I guess I'm just trying to make sure we don't swallow a legit error, when all we want to do is hide the spam.

sburmanoctopus (approved):
LGTM. I would like to consider if our try/catch is catching legit errors as well though. If it's not, then this PR is fine.

source/Halibut/Transport/SecureListener.cs
LukeButters: It will be IO from one reason or another which could be our network legit went down OR the client just disappeared without the common decency to let us know its done.

IOException will cover all of those cases (at least I think).

Otherwise we enter into whack-a-mole of exception messages.

LukeButters:
> LGTM. I would like to consider if our try/catch is catching legit errors as well though. If it's not, then this PR is fine.

In a sense we can't know :p, since we don't know what is on the other end or why it failed.

Since we can't know, their is not much we can do with this information. If we want it ALL back we can make that change in the future or look in verbose logs.

